Issue #18, opened 2019-10-14
RCE (radiative convective equilibrium)


cacraigucar:
RCE changes for trunk (were reviewed in svn for the release branch)

PR #19, opened 2019-10-14
RCE for CAM development (identical to changes on release branch)

cacraigucar:
Addresses #18

Not sure if anyone needs to review this or not, as these changes were reviewed when they went on the release branch (in the svn repo)


cacraigucar:
Since this has already been reviewed (in the release branch) - I am moving ahead with testing and will make the commit once they all pass.

doc/ChangeLog
```diff
@@ -1,4 +1,82 @@
...
+List all files eliminated:
+
+List all files added and what they do:
+A       components/cam/bld/namelist_files/use_cases/aquaplanet_rce_cam6.xml
+          - use case for CAM6 RCE runs
+
+List all existing files that have been modified, and describe the changes:
+M       components/cam/bld/build-namelist
+M       components/cam/bld/namelist_files/namelist_definition.xml
+          - Add use_rad_uniform_angle and rad_uniform_angle namelist variables
```
gold2718: I do not see any defaults for use_rad_uniform_angle and rad_uniform_angle so how are these new items being used? At the very least, this ChangeLog entry is misleading by implying that these new items show up in build_namelist.

brian-eaton: They are used in the RCE use case file.

gold2718: Thanks for pointing this out. I must have mistyped my search for these items in that file.

cacraigucar: And this is regression tested in prebeta.

The ChangeLog needs work (It was written quite a while ago).  I always update my ChangeLog at the end and forgot there was even anything in there yet.  It will be cleaned up prior to the merge.

README.md
```diff
@@ -1,7 +1,5 @@
 # CAM: The Community Atmosphere Model
 
-## NOTE: This is unsupported development code and is subject to the [CESM developer's agreement](http://www.cgd.ucar.edu/cseg/development-code.html).
```
nusbaume: I can't tell if this change is being merged into the cam_development branch?  If so then we will want to un-do this (as it will erase the current warning that shows up on the branch's github page).

cacraigucar: Good catch.  I would have seen this when I wrote the ChangeLog.  It was the status of the README.md when the branch was created and should not be going in.  I will delete it from the merge (hopefully this "delete file"  just deletes the merge request and not the file).

cacraigucar: A few minor tweaks, but it is pretty much the same.

nusbaume: Great, looks good to me!

Issue #17, opened 2019-10-11
COSP2 updates


brian-eaton:
. Fix bug for reversed vertical order in several outputs with cosp_ht
  vertical coordinate.
. Update to COSP v2.1.6.
. Add Cloudsat diagnostics

These changes need to go on both dev and release branches.

PR #22, opened 2019-10-15
COSP updates (#17) for cam_cesm2_1_rel

brian-eaton:
Tests are running.
I don't see the procedure for adding the branch tag in the workflow description.


brian-eaton:
Just asked Cheryl about how to add the tag and it's not clear that we've established this procedure yet.

cacraigucar:
For tagging, I would recommend after the merge, cloning ESCOMP/CAM, applying the tag to the appropriate branch and pushing the tag back directly.  What does everyone think about this approach?

doc/ChangeLog
```diff
@@ -1,5 +1,92 @@
...
+appropriate machine below.  All failed tests must be justified.
+
+cheyenne/intel:
+
+cheyenne/intel/aux_cam:
+
+hobart/nag:
+
+hobart/pgi:
+
```
cacraigucar: From a comment Chris made today - make sure you update your ChangeLog, commit and push it back to your fork BEFORE you merge.

nusbaume:
In general the tagging workflow Cheryl outlined sounds good to me.  

However, I personally get a little anxious anytime there is a direct push to the ESCOMP/CAM repos, as I could imagine that non-reviewed code could mistakenly be added to the repo that way.  An alternative could be to use [github releases](https://help.github.com/en/articles/creating-releases), which are basically the github term for tags, after the PR has been merged. 

Of course, I also appreciate being able to do things on the command line, so I am happy to be over-ruled if we want to avoid GUIs.

doc/ChangeLog
```diff
@@ -1,5 +1,92 @@
...
+Code reviewed by: eaton
+
+List all files eliminated: none
+
+List all files added and what they do: none
+
+List all existing files that have been modified, and describe the changes:
+
+Externals_CAM.cfg
+. update cosp2 external from v2.0.3cesm to v2.1.6
```
nusbaume: The Externals_CAM.cfg file has the tag updated to "v2.1.4cesm".  Is that the same as COSP v2.1.6 described here?  If not, should it be v2.1.4 instead?

src/physics/cam/cospsimulator_intr.F90
```diff
@@ -4,7 +4,7 @@ module cospsimulator_intr
   !         Name:         CFMIP Observational Simulator Package Version 2 (COSP2)
   !         What:         Simulate ISCCP/CloudSat/CALIPSO/MISR/MODIS cloud products from 
   !                       GCM inputs
-  !         Version:      v2.0 (August 2017)
+  !         Version:      v2.1.3 (August 2019)
```
nusbaume: Just like the ChangeLog, should this be v2.1.4 (or v2.1.6) instead?

gold2718:
I am confused by what the problem is. The branch is on your fork, why not tag it whenever you want?
```
git tag rel_cosp01_cam_cesm2_1_rel
git push origin rel_cosp01_cam_cesm2_1_rel
```
or
```
git tag -a rel_cosp01_cam_cesm2_1_rel -m 'This is an important tag'
git push origin rel_cosp01_cam_cesm2_1_rel
```
The branch and the tags stay in your fork. Is someone suggesting something else?

doc/ChangeLog
brian-eaton: This wasn't updated yet since I haven't finalized the ChangeLog with the testing results.  Thanks for the reminder.

gold2718:
@brian-eaton, thanks for the clarification. I have not found an alternative way to make a tag.
All: Here is a proposed workflow:
To create a tag (`<tag>`) for a branch (`<branch_name>`):

1. Clone https://github.com/ESCOMP/CAM or update an existing clone with `git fetch origin <branch_name>`.
2. Find the PR merge commit (`<merge_commit>`) for the tag using one of these techniques:
    - On GitHub, open the PR and click on the "Commits" tab. The merge commit is the last one.
    - In the ESCOMP/CAM clone, find the merge commit with `git log --oneline -1 - origin/<branch_name>` where the merge commit is the first field in the output.
5. Tag the `<merge_commit>` with `git tag -a <tag> <merge_commit> -m '<commit_message'>` where`<commit_message>` is the same as the message for the `<merge_commit>`

Thoughts?

nusbaume:
That looks like a reasonable plan to me, and is similar to Cheryl's idea.  However, you would still need to push that tag back to ESCOMP/CAM, right?  I guess you could use this command:

`git push <escomp remote> <tag>`

To make sure you are pushing only the new tag and nothing else (which is my only, overly paranoid concern).  Otherwise is there a better way to push new tags back to ESCOMP/CAM?

gold2718:
I added some stuff to the SE workflows wiki page (https://github.com/NCAR/CAM/wiki/CAM-SE-Workflows) but the workflow itself still needs a bunch of work. The tag instructions are at https://github.com/NCAR/CAM/wiki/CAM-SE-Workflows#make-a-cam-tag.

PR #32, opened 2019-11-06
cam_cesm2_1_rel_37: Bring in ww3 external from git

cacraigucar:
SInce it appears that there are no more tags coming before the release, I am bringing this change of ww3 coming from git instead of svn in by itself.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -47,8 +47,8 @@ local_path = components/rtm
 
 [ww3]
 tag = ww3_cesm2_0_rel_01
-protocol = svn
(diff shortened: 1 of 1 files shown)


cacraigucar:
Only testing which I propose doing with this commit  (which has been completed successfully) is to run manage_externals and compare the resulting code tree with cam_cesm2_1_rel_36.  The only difference is that the git version does not contain the directory components/ww3/src/cpl_esmf as it is an empty directory in svn.

nusbaume:
Thanks for taking care of this issue Cheryl!

Issue #34, opened 2019-11-21
Changes needed to finish 2deg DECK


fvitt:
Mike Mills has described the desired changes in:
https://docs.google.com/document/d/1TXfpAkVAR3hyHZpprc2LFZMg-RAk7RWVkHa5CHnSLis/edit

nusbaume:
@fvitt Does this issue need to remain open?  It looks like a pull request for it has been merged (and the branch associated with it is now frozen), but I wanted to check with you first  before closing it.

fvitt:
Please close

PR #35, opened 2019-11-21
cam_cesm2_1_rel_38: Updates to compsets at 2-deg resolution

fvitt:
modified:   cime_config/config_component.xml
        modified:   cime_config/config_compsets.xml

See issue #34

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -293,7 +293,7 @@
       <value compset="_CAM60%(WCTS|WCCM)%4xCO2"> flbc_file='$DIN_LOC_ROOT/atm/waccm/lb/LBC_CMIP6abrupt4xCO2_cyclicalYear1_0p5degLat_c180929.nc'  </value>
       <value compset="_CAM60%(WCTS|WCCM)%4xCO2"> nlte_limit_co2=.true.  </value>
       <value compset="_CAM60%WCTS%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e21.BW1850.f09_g17.CMIP6-piControl.001.cam.i.0070-01-01.abrupt4xCO2_c181003.nc'  </value>
-      <value compset="_CAM60%WCCM%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e20.BWma1850.f19_g17.release_cesm2_1_0.020.cam.i.0289-01-01.abrupt4xCO2_c190426.nc'</value>
+      <value compset="_CAM60%WCCM%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e21.BWma1850.f19_g17.CMIP6-piControl-WACCM-MA-2deg.001.cam.i.0301-01-01.abrupt4xCO2_c191118.nc'</value>
       <value compset="_CAM60%1PCT"> flbc_type='SERIAL'</value>
       <value compset="_CAM60%1PCT"> flbc_file='$DIN_LOC_ROOT/atm/waccm/lb/LBC_CMIP6_1pctCO2_y1-165_GlobAnnAvg_0p5degLat_c180929.nc'</value>
       <value compset="_CAM60%1PCT"> flbc_list='CO2','CH4','N2O','CFC11eq','CFC12' </value>
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -408,6 +408,8 @@
 	<value  compset="_CAM60%WCCM*%SDYN">1980-01-01</value>
 	<value  compset="_CAM60%WCTS*%SDYN">1980-01-01</value>
  	<value  compset="_CAM40%WX.*%SDYN">2000-01-01</value>
+	<value  compset="HIST_CAM60_CLM50.*_CICE%PRES_DOCN%DOM_MOSART_CISM2%NOEVOLVE_SWAV"  grid="a%1.9x2.5">1950-01-01</value>
+	<value  compset="HIST_CAM60%WCCM_CLM50.*_CICE%PRES_DOCN%DOM_MOSART_CISM2%NOEVOLVE_SWAV" grid="a%1.9x2.5">1950-01-01</value>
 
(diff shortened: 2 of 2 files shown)


cime_config/config_compsets.xml
```diff
@@ -609,6 +615,13 @@
      </values>
     </entry>
 
+    <entry id="CLM_NAMELIST_OPTS">
+      <values match="first">
+        <value grid="a%1.9x2.5_l%1.9x2.5_oi%1.9x2.5_r%r05_g%gland4_w%null_m%gx1v7"    compset="HIST_CAM60_CLM50.*_CICE%PRES_DOCN%DOM_MOSART_CISM2%NOEVOLVE_SWAV">use_init_interp=.true.</value>
+        <value grid="a%1.9x2.5_l%1.9x2.5_oi%1.9x2.5_r%r05_g%gland4_w%null_m%gx1v7"    compset="HIST_CAM60%WCCM_CLM50.*_CICE%PRES_DOCN%DOM_MOSART_CISM2%NOEVOLVE_SWAV">use_init_interp = .true.</value>
+      </values>
+    </entry>
+
```
cacraigucar: I thought the use_init_interp=.true. settings were no longer to be used by CLM runs.  Is this a case where it really should be (i..e have you confirmed that they are needed with Erik or Bill?)

cime_config/config_component.xml
```diff
@@ -293,7 +293,7 @@
       <value compset="_CAM60%(WCTS|WCCM)%4xCO2"> flbc_file='$DIN_LOC_ROOT/atm/waccm/lb/LBC_CMIP6abrupt4xCO2_cyclicalYear1_0p5degLat_c180929.nc'  </value>
       <value compset="_CAM60%(WCTS|WCCM)%4xCO2"> nlte_limit_co2=.true.  </value>
       <value compset="_CAM60%WCTS%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e21.BW1850.f09_g17.CMIP6-piControl.001.cam.i.0070-01-01.abrupt4xCO2_c181003.nc'  </value>
-      <value compset="_CAM60%WCCM%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e20.BWma1850.f19_g17.release_cesm2_1_0.020.cam.i.0289-01-01.abrupt4xCO2_c190426.nc'</value>
+      <value compset="_CAM60%WCCM%4xCO2"> ncdata='$DIN_LOC_ROOT/atm/waccm/ic/b.e21.BWma1850.f19_g17.CMIP6-piControl-WACCM-MA-2deg.001.cam.i.0301-01-01.abrupt4xCO2_c191118.nc'</value>
```
cacraigucar: According to Mike Mills, the 4xCO2 runs have not been made yet, so updating this file (and changing answers) should not  be an issue.

cime_config/config_compsets.xml
fvitt: I have confirmed with release-clm5.0.29 "use_init_interp = .true." is needed for these 2-degree set ups.  Otherwise the jobs will fail.  I will let @billsacks or @ekluzek weigh in on this.

billsacks: I don't know the details of these configurations and the refcases they're using. @ekluzek do you? Setting use_init_interp seems okay if it's really needed, but it could be a sign that you're pointing to a refcase that is incompatible in some way with the model configuration, and it would be good to understand *why* use_init_interp is needed here.

ekluzek: I'm not seeing why these are needed. They are coming from a historical simulation that's also at 2-degree, so I wouldn't think you would need it. If it was from a control simulation I'd expect it to be needed with init_interp_method='use_finidat_areas'. But, since this is a historical simulation I wouldn't expect that, unless the years are off, which I assume they aren't. I'm guessing there's something else that's different in the simulation that you are starting from that requires this to be set.

fvitt: I believe the ref cases were done with release-clm5.0.25.   Is this why use_init_interp is needed?  I don't know.   Here are some case failures if you would like to investigate:

/glade/scratch/fvitt/SMS_D_Ln9.f19_f19_mg17.FHIST.cheyenne_intel.cam-outfrq9s.cesm2_1_test
/glade/scratch/fvitt/SMS_D_Ln9.f19_f19_mg17.FWmaHIST.cheyenne_intel.cam-outfrq9s.cesm2_1_test
/glade/scratch/fvitt/SMS_Ld1.f19_f19_mg17.FHIST_BGC.cheyenne_intel.cam-outfrq9s.cesm2_1_test

ekluzek: @fvitt can you point us to the CASEDIR for both the simulation that this is pointing to for this case, and your simulation that you ran this to show that use_init_interp=T was required? We prefer avoiding having to set this if it isn't needed. The other way around it would be to create a new "_v2" version of the REFCASE that already has the interpolation done. We did that for a bunch of these so that this wouldn't need to be done.

fvitt:
Erik,
See these cases:

/glade/scratch/fvitt/SMS_D_Ln9.f19_f19_mg17.FHIST.cheyenne_intel.cam-outfrq9s.cesm2_1_test
/glade/scratch/fvitt/SMS_D_Ln9.f19_f19_mg17.FWmaHIST.cheyenne_intel.cam-outfrq9s.cesm2_1_test
/glade/scratch/fvitt/SMS_Ld1.f19_f19_mg17.FHIST_BGC.cheyenne_intel.cam-outfrq9s.cesm2_1_test

cime_config/config_compsets.xml
cacraigucar: For future reference, what was the final decision on the use of use_init_interp=T?  Could you give a little guidance for future projects which might encounter this problem?

ekluzek: OK, the first two cases need use_init_interp=T, because they are non-crop simulations being run from a simulation with crop on. So if that's what's intended then that is required.

I think the reason the last one needs it is because the glacier behavior was changed over Antarctica (multiple to virtual). So again if that's what's intended then it's required.

PR #55, opened 2019-12-31
cam_cesm2_1_rel_40: AM conservation updates

### LINKED ISSUE #54: AM conservation updates
Updates from NorESM (Toniazzo) should be applied to release code.  These updates are for the angular momentum correction and fixer code which is implemented in the FV dycore.  Answers are only affected when the AM conservation options are turned on.  They are off by default.

These mods were applied to the development code at tag cam6_1_029.

brian-eaton:
Fixes #54

### DIFF
--- a/bld/namelist_files/namelist_definition.xml
+++ b/bld/namelist_files/namelist_definition.xml
@@ -563,6 +563,18 @@ Chooses level of velocity diffusion.
 Default: 3.0e5
 </entry>
 
+<entry id="fv_high_order_top" type="logical"  category="dyn_fv"
+       group="dyn_fv_inparm" valid_values="" >
+Flag to extend standard 4th-order PPM scheme to model top.
+Default: .false.
+</entry>
+
+<entry id="fv_am_geom_crrct" type="logical"  category="dyn_fv"
+       group="dyn_fv_inparm" valid_values="" >
+Flag to turn on corrections in FV geometry and/or pressure terms.
+Default: .false.
+</entry>
+
 <entry id="fv_am_correction" type="logical"  category="dyn_fv"
        group="dyn_fv_inparm" valid_values="" >
 Flag to turn on corrections that improve angular momentum conservation.
--- a/src/dynamics/fv/cd_core.F90
+++ b/src/dynamics/fv/cd_core.F90
@@ -12,7 +12,8 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
                    mlt, ncx, ncy, nmfx, nmfy, iremote,            &
                    cxtag, cytag, mfxtag, mfytag,                  &
                    cxreqs, cyreqs, mfxreqs, mfyreqs,              &
-                   kmtp, am_correction, am_fixer, dod, don ,high_order_top)
+                   kmtp, am_correction, am_geom_crrct, am_fixer,  &
+                   dod, don, high_order_top)
 
    ! Dynamical core for both C- and D-grid Lagrangian dynamics
    !
@@ -82,6 +83,7 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
    real(r8), intent(in) :: del2coef
    integer,  intent(in) :: kmtp           ! range of levels (1:kmtp) where order is reduced
    logical,  intent(in) :: am_correction  ! logical switch for correction (applied here)
+   logical,  intent(in) :: am_geom_crrct  ! logical switch for correction (applied here)
    logical,  intent(in) :: am_fixer       ! logical switch for fixer (generate out args)
    logical,  intent(in) :: high_order_top ! use uniform 4th order everywhere (incl. model top)
 
@@ -168,7 +170,7 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
    real(r8), intent(out) ::   &
       ptk(grid%im,grid%jfirst:grid%jlast,grid%kfirst:grid%klast)
 
-! C.-C. Chen, omega calculation
+   ! Omega calculation
    real(r8), intent(out) ::   &
       cx_om(grid%im,grid%jfirst:grid%jlast,grid%kfirst:grid%klast)   ! Courant in X
    real(r8), intent(out) ::  &
@@ -229,7 +231,7 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
    integer :: npes_yz
 
    integer i, j, k, ml
-   integer js1g1, js2g0, js2g1, jn2g1  ,js4g0,jn3g0
+   integer js1g1, js2g0, js2g1, jn2g1, js4g0, jn3g0 
    integer jn2g0, jn1g1
    integer iord , jord
 
@@ -289,7 +291,10 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
    ! Used to allow the same code to execute with or without the AM correction
    real(r8) :: ptr(grid%im,grid%jfirst-1:grid%jlast+1,grid%kfirst:grid%klast+1)
 
-   logical :: sw_am_corr
+   logical  :: sw_am_corr
+   logical  :: am_press_crrct
+   real(r8) :: wg_hiord
+   real(r8) :: tpr, acap
 
    !******************************************************************
    !******************************************************************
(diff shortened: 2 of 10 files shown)


bld/namelist_files/namelist_definition.xml
```diff
@@ -563,6 +563,18 @@ Chooses level of velocity diffusion.
 Default: 3.0e5
 </entry>
 
+<entry id="fv_high_order_top" type="logical"  category="dyn_fv"
+       group="dyn_fv_inparm" valid_values="" >
+Flag to extend standard 4th-order PPM scheme to model top.
+Default: .false.
+</entry>
+
+<entry id="fv_am_geom_crrct" type="logical"  category="dyn_fv"
```
gold2718: crrct? Is this a war on vowels?

doc/ChangeLog
```diff
@@ -1,5 +1,105 @@
...
+Date:
+One-line Summary: Updates to AM correction and fixer code.
+
+Purpose of changes:
+
+. Updates to AM correction and fixer code (FV dycore only)
+  - The settings that are being used in NorESM are:
+    fv_am_diag        = .true.
+    fv_high_order_top = .false.
+    fv_am_geom_crrct  = .false.
```
gold2718: If they are not using this, why does it exist? Who needs it?

doc/ChangeLog
```diff
@@ -1,5 +1,105 @@
...
+
+components/cam/src/dynamics/fv/dp_coupling.F90
+. use am_geom_crrct in place of am_correction
+
+components/cam/src/dynamics/fv/dyn_comp.F90
+. add namelist vars fv_high_order_top and fv_am_geom_crrct
+. the parameter attribute of high_order_top has been removed to allow
+  setting via namelist
+. add am_geom_crrct to cd_core calling args
+. replace am_correction by am_geom_crrct in te_map calling args
```
gold2718: I would love an explanation of what is going on here.

doc/ChangeLog
```diff
@@ -1,5 +1,105 @@
...
+platform, and checkin with these failures has been OK'd by the gatekeeper,
+then copy the lines from the td.*.status files for the failed tests to the
+appropriate machine below.  All failed tests must be justified.
+
+cheyenne/intel: PASS
+
+cheyenne/intel/aux_cam: PASS
+
+hobart/nag: PASS except:
+045 bl369 TBL.sh f10c6aqdm outfrq3s_am 9s .....................................................FAIL! rc= 7 at Mon Dec 30 13:35:07 MST 2019
```
gold2718: Has anyone looked at the new answers? Thomas Toniazzo?

brian-eaton: The change set was relative to cam_cesm2_1_rel_20, so I think we can assume that Thomas validated in that tag which has no answer changes from the current tag.

doc/ChangeLog
brian-eaton: CAM contains many namelist variables for controlling various options which are not on by default in the production configuration.  The code should support options that are useful to researchers.

brian-eaton:
The request to get these changes in the release branch came from Thomas at the same time he requested comments on the paper submitted to JAMES.  I assume the changed code represents what is described in that paper.

doc/ChangeLog
gold2718: Exactly but given the nature of this change, please identify the class of researchers who could ever make use of this. We are not using it and neither is NorESM. If we are providing it, who has vouched for that combination. You assume that Thomas tested the code and that it conforms to the paper but that only goes for the NorESM configuration, not the new combinations you introduce with this change. Have you had any communication about the new but untested combinations?

brian-eaton: I'm not sure why you claim that NorESM researchers aren't exploring these options.  Why would Thomas ask for changes that they have no use for?

gold2718: Okay, I retract my claim. Please provide the communications which refute it. That would be very useful in evaluating whether these new options should be kept and why there should be no test which includes them.

brian-eaton: The paper submitted to JAMES discusses keeping the advection scheme at the same order all the way to the top of the model which appears to be the purpose of the fv_high_order_top variable.  The paper also discusses the treatment of geometric terms in lat/lon coordinates which is presumably relevent to the use of the fv_am_geom_crrct variable.

brian-eaton: I just found in a correspondence with Thomas from last June that the NorESM model sets fv_am_geom_crrct=.true. which means that the comment I added in the ChangeLog file is not correct.  Thomas had sent sample namelist settings for experiments that were reported on in the paper and I must have mistaken one of those for the default NorESM settings.  Since I'm not sure now what those default settings are I will remove that comment since it's not relevant to CAM anyway (NorESM has other modifications to CAM).

doc/ChangeLog
```diff
@@ -1,5 +1,100 @@
...
+Originator(s): Thomas.Toniazzo, eaton
+Date:
+One-line Summary: Updates to AM correction and fixer code.
+
+Purpose of changes:
+
+. Updates to AM correction and fixer code (FV dycore only)
+  - These updates have been documented in a paper submitted to GMD which is
+    currently in discussion phase:
+    https://www.geosci-model-dev-discuss.net/gmd-2019-254/
```
nusbaume: I noticed that here there is a reference to a GMD article, but in the PR discussions there is a reference to a JAMES article.  Does the GMD article contain all of the information on the code-relevant changes?  If not then maybe it would help to add a mention of the JAMES article here as well.

src/dynamics/fv/cd_core.F90
```diff
@@ -1324,23 +1354,30 @@ subroutine cd_core(grid,   nx,     u,   v,   pt,                  &
             do k = kfirst, klast
                do j = js2g0, jlast
                   do i = 1, im
-                     ddu(i,j,k)=ddu(i,j,k)* D0_5*(dpo(i,j,k)+dpn(i,j,k)*3._r8)*D0_5
+                     ddu(i,j,k) = ddu(i,j,k)*0.5_r8*(dpo(i,j,k) + dpn(i,j,k))
```
nusbaume: I realize they are equivalent, but is there a reason why the "D0_5" parameter is replaced by "0.5_r8"?  If this is just a code standards issue, then it should probably be added to the "CAM Coding Standards" wiki page (which I can do).

doc/ChangeLog
gold2718: The paper was originally submited to JAMES but was rejected as not aligned with that journal. Rather than refocus the paper, Thomas (at the suggestion of Peter and myself) submitted it to GMD where it was accepted.

nusbaume: Makes sense.  In that case  I am happy with these ChangeLog lines.

src/dynamics/fv/cd_core.F90
gold2718: They are not necessarily equivalent. The `D` means `double precision` which just means 'greater precision than single precision'. There is no guarantee in Fortran that double (or single) precision has any particular bit counts.
Although it seems that all chip sets are using the same math these days, the past is littered with times when this assumption caused incorrect results. Stick with the `kind` mechanism which at least guarantees a minimum precision.
Yes, adding this to the coding standards page is a good idea -- thanks for offering!

brian-eaton: The original intent of "D0_5" was to follow a coding standard that said something like "constants should have names and the values documented -- avoid literal constants".  In retrospect I think we took that too far in replacing things like 0.5_r8 with the variable D0_5.  I think that literal constants for things that are just numbers rather than physical constants  are easier to read and understand.

brian-eaton: I agree with Steve's comment in general, but in this instance the D is only meant to invoke the meaning of double precision in the mind of the reader:)

gold2718: Sorry, my bad (maybe time for an eye exam?) I agree with @brian-eaton 's remark that restricts the named-constant rule to physical or tuning quantities. Numerical quantities which show up in creating a mathematical expression should be entered as literals.

nusbaume: Sounds good!  I went ahead and added a rule to the Fortran coding standards wiki stating that all real variables, including literals, must have a specified kind, even though that particular concern doesn't quite apply in this case.

gold2718: Right. FYI, this requirement is currently enforced via the TR8 test.

Issue #64, opened 2020-01-22
part of cam6_2_014: Remove unused "use" statement


cacraigucar:
After testing of cam_snapshot was completed, an unused use statement was discovered and it should be removed.

src/physics/cam/physics_types.F90:  "use cam_history,  only: outfld'  is not used and should be removed.

fvitt:
There are several things that should be cleaned up in physics_types:

Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 199: Unused dummy variable PSETCOLS
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 199: Unused local variable STATE
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 199: Unused local variable TEND
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 481: BEGCHUNK explicitly imported into PHYSICS_UPDATE but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 481: CNST_MW explicitly imported into PHYSICS_UPDATE but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 481: ENDCHUNK explicitly imported into PHYSICS_UPDATE but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 481: Unused local variable I
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 481: OUTFLD explicitly imported into PHYSICS_UPDATE but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 671: PI explicitly imported into PHYSICS_STATE_CHECK but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 671: SHR_ASSERT explicitly imported into PHYSICS_STATE_CHECK but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 671: SHR_INFNAN_INF_TYPE explicitly imported into PHYSICS_STATE_CHECK but not used
Warning: /home/fvitt/camdev/waccm_efield/src/physics/cam/physics_types.F90, line 975: Unused local variable M
[...]

PR #70, opened 2020-02-10
cam6_2_013: Changes to efield module to workaround an issue with intel 19

### LINKED ISSUE #68: WACCM physics efield module compiled with intel/19.0.2
WACCM physics efield module produces NaNs in the electric field components over the poles in some configurations compiled with intel/19.0.2 at O2 optimization level on cheyenne.  Example configuartions: FW2000climo compset on grids ne30_ne30_mg17 and ne30pg3_ne30pg3_mg17.

fvitt:
A proposed workaround for an issue with intel/19.0.2 on cheyenne

New tests added :
aux_cam:
 ERP_Ln9.ne30pg3_ne30pg3_mg17.FW2000climo.cheyenne_intel.cam-outfrq9s_wcm_ne30
prealpha:
 SMS_Ln9.ne30_ne30_mg17.FW2000climo.cheyenne_intel.cam-outfrq9s_wcm_ne30

fixes #68

### DIFF
--- a/cime_config/testdefs/testlist_cam.xml
+++ b/cime_config/testdefs/testlist_cam.xml
@@ -572,13 +572,13 @@
       <machine name="cheyenne" compiler="intel" category="waccm"/>
     </machines>
   </test>
-  <test compset="FW1850" grid="f09_f09_mg17" name="SMS_Ld1" testmods="cam/reduced_hist1d">
+  <test compset="FW1850" grid="f09_f09_mg17" name="SMS_Ln9" testmods="cam/reduced_hist3s">
     <machines>
       <machine name="cheyenne" compiler="intel" category="aux_cam"/>
       <machine name="cheyenne" compiler="intel" category="prealpha"/>
     </machines>
     <options>
-      <option name="wallclock">00:30:00</option>
+      <option name="wallclock">00:10:00</option>
     </options>
   </test>
   <test compset="FWHIST_BGC" grid="f09_f09_mg17" name="ERP_Ld3" testmods="cam/reduced_hist1d">
--- a/cime_config/testdefs/testlist_cam.xml
+++ b/cime_config/testdefs/testlist_cam.xml
@@ -986,6 +986,23 @@
       <machine name="cheyenne" compiler="intel" category="waccm"/>
     </machines>
   </test>
+  <test compset="FW2000climo" grid="ne30pg3_ne30pg3_mg17" name="ERP_Ln9" testmods="cam/outfrq9s_wcm_ne30">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="waccm"/>
+      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+    </machines>
+  </test>
+  <test compset="FW2000climo" grid="ne30pg3_ne30pg3_mg17" name="SMS_D_Ln9" testmods="cam/outfrq9s_wcm_ne30">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="waccm"/>
+    </machines>
+  </test>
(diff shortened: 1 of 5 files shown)


src/physics/waccm/efield.F90
```diff
@@ -1481,26 +1506,14 @@ subroutine DerivPotential
 !-----------------------------------------------------------------------
 ! Poles:
 !-----------------------------------------------------------------------
-!$omp parallel do private(i, ip1f,ip2f,ip3f)
+!$omp parallel do private(i)
```
gold2718: I take it this is where the bug was? Does it only show up with threading turned on?

src/physics/waccm/efield.F90
```diff
@@ -199,6 +209,21 @@ subroutine efield_init(efield_lflux_file, efield_hflux_file, efield_potential_ma
 
       epotential_max = efield_potential_max
 
+      do i = 0,nmlon
```
gold2718: Out of curiosity, what happens if you try to thread this loop?

cime_config/testdefs/testlist_cam.xml
```diff
@@ -986,6 +986,23 @@
       <machine name="cheyenne" compiler="intel" category="waccm"/>
     </machines>
   </test>
+  <test compset="FW2000climo" grid="ne30pg3_ne30pg3_mg17" name="ERP_Ln9" testmods="cam/outfrq9s_wcm_ne30">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="waccm"/>
+      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
```
cacraigucar: I'm questioning the addition of an aux_cam test for this configuration.  Compiler changes will happen at the cime level and will be first tested with prealpha tests.  Since there is already an aux_cam test for f09_f09_mg17, is this sufficient to test changes which might occur during modifications to the CAM code?  If this test if moved to prealpha, then I believe the prealpha test below can be removed.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -578,7 +578,7 @@
       <machine name="cheyenne" compiler="intel" category="prealpha"/>
     </machines>
     <options>
-      <option name="wallclock">00:10:00</option>
+      <option name="wallclock">00:30:00</option>
```
cacraigucar: If this test is taking longer than 10 minutes, perhaps it should be shortened from 1 day for aux_cam testing.

cacraigucar:
Two more minor items.
1. Brian Eaton is out of town until next Tuesday, so you don't need to wait for his review.
2. It's looking like you may be getting the next CAM tag slot (Chris is still working on the release).  When I give it to you, please make the change for #64 and edit your PR description to include "closes #64"

src/physics/waccm/efield.F90
fvitt: Not always. One of the configurations was a pure MPI configuration (NTHRDS=1).
It seemed that the branching in the "Poles" loop was sometimes causing issues with intel/19, regardless of the threading.

src/physics/waccm/efield.F90
fvitt: This loop is executed only at init time.

cime_config/testdefs/testlist_cam.xml
fvitt: This is different than ne30_ne30.  This tests cslam.

cacraigucar: Okay - this test is appropriate

cime_config/testdefs/testlist_cam.xml
fvitt: I shortened this test

PR #71, opened 2020-02-13
cam6_2_019: Mods for ccpp kessler

cacraigucar:
Bring in ccpp-ized kessler 

NOTE - the current 9 time step test is insufficient to test kessler as it takes over a  month to get spun up enough to show errors.  I can introduce an updated kessler test which uses a spun-up case, if this is desired.


Externals_CAM.cfg
```diff
@@ -33,6 +33,13 @@ repo_url = https://github.com/ESCOMP/SILHS_CESM
 tag = silhs_ncar_backwards_compat_20181205_c20191001
 required = True
 
+[atmos_phys]
+tag = top_to_bottom_004
+protocol = git
+repo_url = https://github.com/cacraigucar/atmospheric_physics
+required = True
```
cacraigucar: Note that this will point to the NCAR/atmospheric_physics repository after it is updated

cacraigucar:
I have updated the external to use the newly committed NCAR/atmospheric_physics commit.

Please review these changes for CAM code to utilize the CCPP-ized version of Kessler.

I also need to know if I should modify the Kessler test to use a spun-up input, so that the test is actually meaningful (and not able to hide introduced bugs).

src/physics/simple/kessler_cam.F90
```diff
@@ -89,32 +122,36 @@ subroutine kessler_tend(state, ptend, ztodt, pbuf)
...
     integer :: i
 
     !
     !-----------------------------------------------------------------------
     !
+    errflg = 0
     lchnk = state%lchnk
     ncol  = state%ncol
 
+    lyr_surf = 30
```
gold2718: Shouldn't this be pver?

cacraigucar: agreed

cacraigucar:
This is ready for review again.  Please note that I have a temporary tag to my atmospheric_physics fork, but it will point to the NCAR/atmospheric_physics repo once those changes have been approved and before I do the merge.

src/physics/simple/kessler_cam.F90
```diff
@@ -56,9 +65,24 @@ subroutine kessler_init(pbuf2d)
     call add_default(cnst_name(ixrain),   1, ' ')
 
     ! Initialize Kessler with CAM physical constants
-    call kessler_set_const(rair, cpair, latvap, pstd/100.0_r8, rhoh2o)
 
-  end subroutine kessler_init
+      if (errflg == 0) then
```
nusbaume: Given that `errflg` will always be zero here (at least as currently written), is this check necessary?

src/physics/simple/kessler_cam.F90
```diff
@@ -123,72 +159,59 @@ subroutine kessler_tend(state, ptend, ztodt, pbuf)
...
-    ! qr:    Rain  water mixing ratio (gm/gm) (inout)
-    ! precl: Precipitation rate (m_water / s) (out)
-    ! errmsg: Error string if error found
-    call kessler(ncol, pver, ztodt, rho(:ncol,:), z(:ncol,:), pk(:ncol,:),    &
-         th(:ncol,:), qv(:ncol,:), qc(:ncol,:), qr(:ncol,:), prec_sed, errmsg)
-
-    if (len_trim(errmsg) > 0) then
-      call endrun(trim(errmsg))
+    
+    if (errflg == 0) then
```
nusbaume: Given that `errflg` will always be zero here (at least as currently written), is this check necessary?

src/physics/simple/kessler_cam.F90
cacraigucar: My procedure for creating the calls for this file was to cut/paste the code from the ccpp generated file.  I saw that initial check (which does nothing), but decided that having the code match the generated code was more important.

src/physics/simple/kessler_cam.F90
cacraigucar: Same as previous

src/physics/simple/kessler_cam.F90
```diff
@@ -74,9 +98,17 @@ subroutine kessler_tend(state, ptend, ztodt, pbuf)
     use physics_types,      only: physics_ptend_init
     use constituents,       only: pcnst, cnst_name, cnst_type
 
+    use kessler,          only: kessler_run
+    use state_converters, only: wet_to_dry_run
+    use state_converters, only: dry_to_wet_run
+    use geopotential_t,   only: geopotential_t_run
```
gold2718: This is not called, please remove.

src/physics/simple/kessler_cam.F90
gold2718: Why did you mark this resolved if it is not fixed?

src/utils/ccpp_kinds.F90
```diff
@@ -0,0 +1,30 @@
...
+! FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+! THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+! IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+! CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+!>
+!! @brief Auto-generated kinds for CCPP
+!!
+!
```
gold2718: This module, as written, is not part of the CCPP Framework and is not auto-generated. Please remove these comments and replace with a statement that this file is a placeholder for the auto-generated version used by CCPP host models.

src/utils/ccpp_kinds.F90
```diff
@@ -0,0 +1,30 @@
...
+
+module ccpp_kinds
+
+!   use ISO_FORTRAN_ENV, only: kind_phys => REAL64
```
gold2718: No need for this comment, please remove.

src/physics/simple/kessler_cam.F90
cacraigucar: I didn't make any new "resolve conversations" - this was old.

cime_config/testdefs/testmods_dirs/cam/outfrq9s_kessler/shell_commands
```diff
@@ -0,0 +1,2 @@
+./xmlchange ROF_NCPL=\$ATM_NCPL
+./xmlchange GLC_NCPL=\$ATM_NCPL
```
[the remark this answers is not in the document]
gold2718: It looks like this file was removed.

cime_config/testdefs/testmods_dirs/cam/outfrq9s_kessler/user_nl_clm
```diff
@@ -0,0 +1,27 @@
...
+! Set finidat            with RUN_REFCASE/RUN_REFDATE/RUN_REFTOD options for hybrid or branch cases
+!                        (includes $inst_string for multi-ensemble cases)
+! Set glc_grid           with CISM_GRID                          option
+! Set glc_smb            with GLC_SMB                            option
+! Set maxpatch_glcmec    with GLC_NEC                            option
+! Set glc_do_dynglacier  with GLC_TWO_WAY_COUPLING               env variable
+!----------------------------------------------------------------------------------
+hist_nhtfrq = 9
+hist_mfilt  = 1
+hist_ndens  = 1
```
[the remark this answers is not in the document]
gold2718: It looks like this file was removed.

src/physics/simple/kessler_cam.F90
gold2718: Hmm. I thought I had 'unresolved' it. No worries.

cime_config/testdefs/testmods_dirs/cam/outfrq9s_kessler/user_nl_clm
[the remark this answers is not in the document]
cacraigucar: That was the request, correct?   My testing indicated it indeed wasn't needed.

src/physics/simple/kessler_cam.F90
```diff
@@ -56,9 +65,24 @@ subroutine kessler_init(pbuf2d)
     call add_default(cnst_name(ixrain),   1, ' ')
 
     ! Initialize Kessler with CAM physical constants
-    call kessler_set_const(rair, cpair, latvap, pstd/100.0_r8, rhoh2o)
 
-  end subroutine kessler_init
+      if (errflg == 0) then
+         call kessler_init(cpair, latvap, pstd, rhoh2o, errmsg, errflg)
+      else
+         return
```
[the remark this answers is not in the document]
gold2718: Fixed!

fvitt:
Don't have time review this

gold2718:
I removed you as a reviewer. In the future, you can remove yourself following the guidelines in step 11 of the [CAM  SE Workflows](https://github.com/NCAR/CAM/wiki/CAM-SE-Workflows) wiki page.

cacraigucar:
Needed to have kessler.F90 compiled so that the use statement in the simple models was available (intel complained).  There are other, in my opinion less optimal ways to solve this, which is why I'm requesting a review of the minor change I made to configure with the e18adb1 commit.

src/physics/simple/kessler_cam.F90
```diff
@@ -56,9 +66,24 @@ subroutine kessler_init(pbuf2d)
     call add_default(cnst_name(ixrain),   1, ' ')
 
     ! Initialize Kessler with CAM physical constants
-    call kessler_set_const(rair, cpair, latvap, pstd/100.0_r8, rhoh2o)
 
-  end subroutine kessler_init
+      if (errflg == 0) then
+         call kessler_init(cpair, latvap, pstd, rhoh2o, errmsg, errflg)
+      else
+         call endrun('kessler_cam_init error: Error returned from kessler_init')
```
nusbaume, line 73: I think the endrun messages here are all off by one subroutine call.  For example, if this endrun was triggered, then it means `errflg` was not equal to zero before `kessler_init` was called, which means the `kessler_init` call is not the source of the error.

src/physics/simple/kessler_cam.F90
```diff
@@ -56,9 +66,24 @@ subroutine kessler_init(pbuf2d)
...
+         call pres_to_density_dry_init(cpair, rair, errmsg, errflg)
+      else
+         call endrun('kessler_cam_init error: Error returned from pres_to_density_dry_init')
+      end if
+      if (errflg == 0) then
+         call calc_exner_init(errmsg, errflg)
+      else
+         call endrun('kessler_cam_init error: Error returned from calc_exner_init')
+      end if
+
```
nusbaume, line 85: Do we need to add one more `if (errflg == 0)` check here, for the `calc_exner_init` call?

src/physics/simple/kessler_cam.F90
```diff
@@ -123,72 +159,59 @@ subroutine kessler_tend(state, ptend, ztodt, pbuf)
...
-    call kessler(ncol, pver, ztodt, rho(:ncol,:), z(:ncol,:), pk(:ncol,:),    &
-         th(:ncol,:), qv(:ncol,:), qc(:ncol,:), qr(:ncol,:), prec_sed, errmsg)
-
-    if (len_trim(errmsg) > 0) then
-      call endrun(trim(errmsg))
+    
+    if (errflg == 0) then
+       call calc_exner_run(ncol, pver, cpair, rair, state%pmid, pk, errmsg, errflg)
+    else
+       call endrun('kessler_tend error: Error returned from calc_exner_run')
```
nusbaume, line 175: Same issue here. I think the actual text for the `endrun` calls in this code section should be shifted down by one `errflg` check.

src/physics/simple/kessler_cam.F90
```diff
@@ -123,72 +159,59 @@ subroutine kessler_tend(state, ptend, ztodt, pbuf)
...
+    end if
+    if (errflg == 0) then
+       call potential_temp_to_temp_run(ncol, pver, th, pk, temp, errmsg, errflg)
+    else
+       call endrun('kessler_tend error: Error returned from potential_temp_to_temp_run')
+    end if
+    if (errflg == 0) then
+       call dry_to_wet_run(ncol, pver, state%pdel, state%pdeldry, qv, qc, qr, errmsg, errflg)
+    else
+       call endrun('kessler_tend error: Error returned from dry_to_wet_run')
```
nusbaume, line 206: Do we need to add one more `if (errflg == 0)` check here, for the `dry_to_wet_run` call?

cacraigucar:
@nusbaume Good catch! 

I'd like to get @gold2718 to comment on these error messages.  It feels very dangerous to have the error message in a different if block, but with trying to allow cut/paste from the generated code as much as possible, following Jesse's suggestion seems reasonable. 

An alternative would be to delete the "else" blocks and put in an additional if (errflg ==0) block after every call with the error message.  This too doesn't seem like the perfect solution.

What procedure do we want to adopt for copying the generated code back to CAM, and dealing with the error messages?  Note that the generated code had a return instead of an error message, so it didn't have this problem.  The error messages were put in at the request of @gold2718 (though put in incorrectly due to the different structure for the error handling that the generated code is using).

gold2718:
Ouch! Can't believe I missed that, nice catch!
I suggest a slightly different pattern:
```
   if (errflg == 0) then
      call dry_to_wet_run(ncol, pver, state%pdel, state%pdeldry, qv, qc, qr, errmsg, errflg)
      if (errflg /= 0) then
         call endrun('kessler_tend: Error returned from dry_to_wet_run: '//trim(errmsg))
      end if
   end if
```

This keeps call and error blocks together. I think it is also important to output the error message!
We could also decide that our `errmsg` strings include information about the CCPP scheme where the error was found (e.g., "Error returned from dry_to_wet_run:" string) to simplify the `endrun` call.

src/physics/simple/kessler_cam.F90
gold2718: I think it is handled now, please review and resolve this conversation if you agree (and let us know if you do not agree).

PR #75, opened 2020-02-14
part of cam6_2_014: mg utils comment change, mg3 graupel order change

### LINKED ISSUE #74: Microphysics Cleanup
Greetings,

I've found two mistakes in the mg3 microphysics: neither answer changing, but they should be fixed.

1. micro_mg_utils.F90: There are two comments on the size distribution calculation for liquid that are incorrect, because the formula was updated for CAM6 but the comment was not: The comment on eta calculation is changed to refer to Liu and Rostayn, 2003 NOT martin.

2. micro_mg3_0.F90: mg3 actually will work with do_hail and do_graupel both true. However, there was a mix of which was the LAST one called. I switched size distribution calls to be graupel last so if both do_graupel and do hail are set, answers will be identical to do_graupel.

Neither changes answers.  I have fixed these on a fork in the repository. Let's see if I have done it right.

andrewgettelman:
This changes 2 files; comments in micro_mg_utils.F90.

In micro_mg3_0.F90, the code will tolerate both do_graupel and do_hail = true (CAM at higher levels does not allow it, but the microphysics does permit it).  The changes move the places where  do_graupel and do_hail appear separately so that graupel is always last. This will give an equivalent result to do_graupel = true, do_hail = false when both are true.

It also needs to propagate to the PUMAS repository for the microphysics (@Katetc ) can handle that hopefully....

Fixes #74

### DIFF
--- a/src/physics/cam/micro_mg_utils.F90
+++ b/src/physics/cam/micro_mg_utils.F90
@@ -431,7 +431,7 @@ elemental subroutine size_dist_param_liq_line(props, qcic, ncic, rho, pgam, lamc
      ! arguments.)
      props_loc = props
 
-     ! Get pgam from fit to observations of martin et al. 1994
+     ! Get pgam from fit Rotstayn and Liu 2003 (changed from Martin 1994 for CAM6)
      pgam = 1.0_r8 - 0.7_r8 * exp(-0.008_r8*1.e-6_r8*ncic*rho)
      pgam = 1._r8/(pgam**2) - 1._r8
      pgam = max(pgam, 2._r8)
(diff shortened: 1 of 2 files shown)


andrewgettelman:
This fixes issue #74

src/physics/cam/micro_mg_utils.F90
```diff
@@ -431,7 +431,7 @@ elemental subroutine size_dist_param_liq_line(props, qcic, ncic, rho, pgam, lamc
      ! arguments.)
      props_loc = props
 
-     ! Get pgam from fit to observations of martin et al. 1994
+     ! Get pgam from fit Rotstayn and Liu 2003 (changed from Martin 1994 for CAM6)
```
fvitt: Is the "changed from Martin..." comment needed?

andrewgettelman: I would like to leave the Martin reference. The use of the new equation is not published anywhere, so this helps traceability.

Issue #86, opened 2020-03-11
Cheyenne jobs randomly failing with deallocation


cacraigucar:
In some instances cheyenne jobs just fail with an unhelpful message that a deallocation is invalid as the pointer has not been allocated.  In some instances the job can fail on the first timestep and other times it can be later.  This different behavior can all happen with the exact same code base.

The jobs which fail with a traceback indicate that the deallocate at line 267 in rrtmg_state.F90 is failing.  While this appears to be valid code, there is something which is causing cheyenne to have random issues.

A couple of suggestions that could solve the random issue:
 - Remove the pointer attribute for the rstate variable
 - Introduce an "if allocated(rstate)" prior to the deallocate.

brian-eaton:
I've taken a look at this code and don't see any reason for r_state to be a pointer in radiation_tend.  Should be able to remove the pointer attribute and use it as a local object with fairly minimal changes required downstream.

PR #87, opened 2020-03-11
part of cam6_2_016: Workaround rrtmg_state pointer issue on cheyenne

fvitt:
modified:   doc/ChangeLog
        modified:   src/physics/rrtmg/radiation.F90
        modified:   src/physics/rrtmg/rrtmg_state.F90

I ran 2 cases (f19 and f09 resolutions) for 10 days successfully on cheyenne, both with DEBUG=TRUE and NTHRDS=3.

Fixes issue #86


gold2718:
I will add this to my upcoming cam6_2_016 tag so you do not have to do any testing.

Issue #92, opened 2020-03-15
error in using ESMF library in subroutine set_time_float_from_date


mvertens:
The following error occurs when using the NUOPC/CMEPS cap for the prealpha test 
`ERS_Ld3_Vnuopc.f09_f09_mg17.FCfireHIST.cheyenne_intel.cam-outfrq1d
`
```
ESMCI_Calendar.C:1161 ESMCI::Calendar::convertToTime() Input argument out of range  - ; NoLeap: for month 2, dd=29 > 28 days in the month.
 ESMCI_Time.C:333 ESMCI::Time::set() Input argument out of range  - Internal subroutine call returned Error
```

This error is triggered by the call to routine solar_irrad_init in module solar_data
   ```
call solar_irrad_init( solar_irrad_data_file, fixed_solar, solar_data_ymd, solar_data_tod, &
                           solar_const, solar_htng_spctrl_scl )
```
and in turn solar_irrad_init calls 
   ```
call time_coord%initialize( filepath, fixed=fixed, fixed_ymd=fixed_ymd, fixed_tod=fixed_tod, &
                                force_time_interp=.true., try_dates=.true. )
```
which calls

`call set_time_float_from_date( ref_time, ref_yr, ref_mon, ref_day, tod )
`

where the following code block extists:

```
call ESMF_TimeSet( date, yy=year, mm=month, dd=day, s=sec, calendar=tm_cal, rc=rc)
  !
  ! If the subroutine returned error, check if it is Feb 29 of a non-leap year
  ! (legitimately used by the time-interpolation routines in tracer_data.F90)
  ! in which case, substitute Feb 28 for the day
  !
  if ( rc .ne. ESMF_SUCCESS ) then
     if ( ( month .eq. 2 ) .and. ( day .eq. 29 ) ) then ! assume the failure is because it is leap day
        useday = 28
        call ESMF_TimeSet( date, yy=year, mm=month, dd=useday, s=sec, calendar=tm_cal, rc=rc)
     else  ! legitimate error, let the model quit
        call chkrc(rc, sub//': error return from ESMF_TimeSet for set_time_float_from_date')
     endif
  endif
```
The problem is that the following call 
**call ESMF_TimeSet( date, yy=year, mm=month, dd=day, s=sec, calendar=tm_cal, rc=rc)**
returns an error when the ESMF library is used.  

[...]

PR #99, opened 2020-03-18
check for leap day before ESMF_TimeSet is invoked

fvitt:
modified:   src/utils/time_manager.F90

Hopefully this fixes issue #92 

Not able to test with NUOPC coupling

### DIFF
--- a/src/utils/time_manager.F90
+++ b/src/utils/time_manager.F90
@@ -257,20 +257,21 @@ subroutine set_time_float_from_date( time, year, month, day, sec )
   type(ESMF_TimeInterval) :: diff
   integer :: useday
 
-  call ESMF_TimeSet( date, yy=year, mm=month, dd=day, s=sec, calendar=tm_cal, rc=rc)
+  if ( (calendar==shr_cal_noleap) .and. (month==2) .and. (day==29) ) then ! workaround leap days for NOLEAP cal
+     useday = 28
+  else 
+     useday = day
+  endif
+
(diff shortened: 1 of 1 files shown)


src/utils/time_manager.F90
```diff
@@ -257,20 +257,21 @@ subroutine set_time_float_from_date( time, year, month, day, sec )
...
 
-  call ESMF_TimeSet( date, yy=year, mm=month, dd=day, s=sec, calendar=tm_cal, rc=rc)
+  if ( (calendar==shr_cal_noleap) .and. (month==2) .and. (day==29) ) then ! workaround leap days for NOLEAP cal
+     useday = 28
+  else 
+     useday = day
+  endif
+
+  call ESMF_TimeSet( date, yy=year, mm=month, dd=useday, s=sec, calendar=tm_cal, rc=rc)
   !
```
mvertens: I think this comment is no longer needed - since at this point ESMF_TimeSet should never return an error with your above introduction of useday.
If you get an error at this point you should abort - not return (which is the case here).

fvitt: agree

cacraigucar:
Which CAM branch or branches does this change go into?

fvitt:
See PR #103

cacraigucar:
Thanks - I missed that.  
PR #103 will need indicate that it closes this PR in it's description so this PR is closed when #103 is closed.

Issue #91, opened 2020-03-13
Improvements to aerosol wet removal scheme


fvitt:
The Wang code is trying to do convective transport and wet removal of aerosols at the same time, so that the wet removal can happen on the aerosols activated by convection. The changes by Yunpeng are mostly to change the way aerosols are released when rain evaporates. In Wang a fraction of aerosol is released each time some of the rain evaporates and it is released into the same mode it came from. In Yunpeng’s changes, the aerosols are not released until all the rain has evaporated and it is released into the largest size mode.

PR #103, opened 2020-03-24
cam6_2_021: Enhancements to wet removal of aerosols

fvitt:
These new features are turned off by default.  Baselines should be bit-for-bit unchanged.

closes issue #91
closes issue #92 -- bug fix for noleap calendar with ESMF lib
closes PR #99

### DIFF
--- a/src/chemistry/aerosol/wetdep.F90
+++ b/src/chemistry/aerosol/wetdep.F90
@@ -29,11 +29,13 @@ public :: wetdep_inputs_set
 real(r8), parameter :: cmftau = 3600._r8
 real(r8), parameter :: rhoh2o = 1000._r8            ! density of water
 real(r8), parameter :: molwta = 28.97_r8            ! molecular weight dry air gm/mole
+real(r8), parameter :: omsm = 1._r8-2*epsilon(1._r8) ! used to prevent roundoff errors below zero
 
 type wetdep_inputs_t
    real(r8), pointer :: cldt(:,:) => null()  ! cloud fraction
    real(r8), pointer :: qme(:,:) => null()
    real(r8), pointer :: prain(:,:) => null()
+   real(r8), pointer :: bergso(:,:) => null()
    real(r8), pointer :: evapr(:,:) => null()
    real(r8) :: cldcu(pcols,pver)     ! convective cloud fraction, currently empty
    real(r8) :: evapc(pcols,pver)     ! Evaporation rate of convective precipitation
@@ -48,6 +50,7 @@ end type wetdep_inputs_t
 integer :: cld_idx             = 0
 integer :: qme_idx             = 0 
 integer :: prain_idx           = 0 
+integer :: bergso_idx          = 0 
 integer :: nevapr_idx          = 0 
 
 integer :: icwmrdp_idx         = 0 
@@ -70,9 +73,12 @@ subroutine wetdep_init()
   use physics_buffer, only: pbuf_get_index
   use constituents,   only: cnst_get_ind
 
+  integer :: ierr
+
   cld_idx             = pbuf_get_index('CLD')    
   qme_idx             = pbuf_get_index('QME')    
   prain_idx           = pbuf_get_index('PRAIN')  
+  bergso_idx          = pbuf_get_index('BERGSO', errcode=ierr )  
   nevapr_idx          = pbuf_get_index('NEVAPR') 
 
   icwmrdp_idx         = pbuf_get_index('ICWMRDP') 
@@ -133,6 +139,15 @@ subroutine wetdep_inputs_set( state, pbuf, inputs )
   call pbuf_get_field(pbuf, dp_frac_idx,     dp_frac )
   call pbuf_get_field(pbuf, nevapr_shcu_idx, evapcsh )
   call pbuf_get_field(pbuf, nevapr_dpcu_idx, evapcdp )
+
+  if (bergso_idx>0) then
+     call pbuf_get_field(pbuf, bergso_idx, inputs%bergso )
+  else
+     if (.not. associated(inputs%bergso)) then
+        allocate(inputs%bergso(pcols,pver))
+        inputs%bergso(:,:) = 0.0_r8
+     endif
+  endif
 
   inputs%cldcu(:ncol,:)  = dp_frac(:ncol,:) + sh_frac(:ncol,:)
   cldst(:ncol,:)          = inputs%cldt(:ncol,:) - inputs%cldcu(:ncol,:)       ! Stratiform cloud fraction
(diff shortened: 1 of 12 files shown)


src/chemistry/modal_aero/modal_aero_convproc.F90
```diff
@@ -108,6 +120,58 @@ subroutine ma_convproc_register
...
+        if (ierr /= 0) then
+           call endrun(subname // ':: ERROR reading namelist')
+        end if
+     end if
+     close(unitn)
+     call freeunit(unitn)
+  end if
+
+  ! Broadcast namelist variables
+  call mpi_bcast( convproc_do_gas,  1, mpi_logical, masterprocid, mpicom, ierr)
```
nusbaume: Should these calls have error checks, for example:

`if (ierr /= 0) call endrun(sub//": FATAL: mpi_bcast: convproc_do_gas")`

I realize the odds of this happening are low, but it couldn't hurt.

src/chemistry/aerosol/wetdep.F90
```diff
@@ -447,6 +475,13 @@ subroutine wetdepa_v2(                                  &
          !                 which evaporates within this layer
          fracev(i) = evaps(i,k)*pdog(i) &
                      /max(1.e-12_r8,precabs(i))
+                     
+         ! If resuspending aerosol only when all the rain has totally
+         ! evaporated then zero out any aerosol tendency for partial
+         ! evaporation.
+         if (convproc_do_evaprain_atonce .and. (evaps(i,k)*pdog(i).ne.precabs(i))) then
```
gold2718: Still using `.ne.` over `/=`?

src/chemistry/modal_aero/aero_model.F90
```diff
@@ -1603,7 +1613,36 @@ subroutine aero_model_wetdep( state, dt, dlf, cam_out, ptend, pbuf)
...
+                         mm = numptrcw_amode(m)
+                      endif
+                   else if (lspec <= nspec_amode(m)) then ! non-water mass
+                      if (lphase == 1) then
+                         mm = lmassptr_amode(lspec,m)
+                      else
+                         mm = lmassptrcw_amode(lspec,m)
+                      endif
+                   endif
+                   if (lphase .eq. 2) then
```
gold2718: Please be consistent (e.g., you* use '==' on line 1624).

*I have no idea where this code came from so I'm blaming you :)

src/chemistry/modal_aero/modal_aero_convproc.F90
gold2718: I've gone back and forth on this over the decades. I used to think that catching stuff like this* was a good idea. On the other hand, in 2020, an error in a place like this will crash and burn with little likelihood that control will ever return to process this check.
Now, for fancy MPI code (e.g., asynchronous sends and receives), it is often worth trying to catch these errors but I now feel that error checks in places like this is a waste of code. To be clear, this is because of the evolution of network fabric (e.g., Infiniband) and MPI libraries, in general, it is good SE practice.

*Note that, by default, MPI routines such as MPI_bcast will not return on error. You have to set an error handler for that.

src/chemistry/modal_aero/modal_aero_convproc.F90
```diff
@@ -108,6 +120,58 @@ subroutine ma_convproc_register
...
+  integer :: unitn, ierr
+  character(len=*), parameter :: subname = 'ma_convproc_readnl'
+
+  namelist /aerosol_convproc_opts/ convproc_do_gas, deepconv_wetdep_history, convproc_do_deep, &
+       convproc_do_shallow, convproc_do_evaprain_atonce, convproc_pom_spechygro, convproc_wup_max
+
+  ! Read namelist
+  if (masterproc) then
+     unitn = getunit()
+     open( unitn, file=trim(nlfile), status='old' )
```
gold2718: `getunit` and `freeunit` are not needed and should be deprecated. Please get rid of them and use:
```
open( newunit=unitn, file=trim(nlfile), status='old' )
```

src/chemistry/modal_aero/modal_aero_convproc.F90
```diff
@@ -1143,6 +1213,15 @@ subroutine ma_convproc_tend(                                           &
...
       ktop = jt(i)
       kbot = mx(i)
+! REASTER 08/05/2015 BEGIN
+! usually the updraft ( & downdraft) start ( & end ) at kbot=pver, but sometimes kbot < pver
+! transport, activation, resuspension, and wet removal only occur between kbot >= k >= ktop
+! resuspension from evaporating precip can occur at k > kbot when kbot < pver
+! in the first version of this routine, the precp evap resusp tendencies for k > kbot were ignored, 
+!    but that is now fixed
+! this was a minor bug with quite minor affects on the aerosol, 
+!    because convective precip evap is (or used to be) much less than stratiform precip evap )
```
cacraigucar: This is a lengthy comment of what "used to be", including the developers name and a date over 4 years ago?

Should some (all?) of this comment be moved to the ChangeLog?

src/chemistry/modal_aero/modal_aero_convproc.F90
fvitt: If checking for error is a waste of code then I will leave as it is, if there are no objections.

src/chemistry/aerosol/wetdep.F90
fvitt: see below

src/chemistry/modal_aero/aero_model.F90
fvitt: Many hands have modified this code so you see different styles.. I tried to make the styles of conditional operators consistent.

src/chemistry/modal_aero/modal_aero_convproc.F90
fvitt: OK good to know.  I made the change as you suggest.

src/chemistry/modal_aero/modal_aero_convproc.F90
fvitt: I moved some of this to ChangeLog.  Please review.

fvitt:
I believe I have addressed the reviewer's issues.  I have included a small change to src/chemistry/mozart/mo_extfrc.F90 to improve error checking on input files.  Please review.

src/chemistry/modal_aero/modal_aero_convproc.F90
nusbaume: That's fine with me, given @gold2718's explanation.  I'll go ahead and resolve this conversation.

Issue #131, opened 2020-05-05
modal_aero_rename opens too many files 


jedwards4b:
In file [cam/src/chemistry/modal_aero/modal_aero_rename.F90](https://github.com/ESCOMP/CAM/blob/cam_development/src/chemistry/modal_aero/modal_aero_rename.F90) there is code that I believe to be incorrectly opening excessive numbers of files.   In my testing of compset f19_f19.QPC6 I am getting an error "Too many open files" that I have traced to the open statement at [line 367](https://github.com/ESCOMP/CAM/blob/cam_development/src/chemistry/modal_aero/modal_aero_rename.F90#L367) of this file.   This code is repeated at line 996 of the file.

In each case the file is being opened on all tasks but primarily is only used on master.
In my sandbox i have commented the calls at line 367 and 996 and it solves the problem.

jedwards4b:
Commenting these open statements causes the total number of open files reported to go from several thousand to around 12.

cacraigucar:
@fvitt I believe this is your code.  If it is a bug, could you add which projects (cam_development and/or release code) need the fix.  If you agree that Jim's fix is correct, you can also add the "next tag" label and we can add it to the next tags in the respective branches as this change can go in with others.

fvitt:
Yikes!  It looks like something leftover from someone's development (not mine) -- trying to dump stuff out to file... I can do a quick cleanup if you like.

cacraigucar:
@fvitt  Sorry - I didn't mean to imply you wrote it  - you are the recipient of many people's development efforts.  Yes, if you can propose a fix, we can put it into the next CAM tag(s) - unless this needs to go in ASAP.

jedwards4b:
Now that I've found a work around, I don't see this issue as urgent - the next tag is fine.

PR #132, opened 2020-05-05
cam6_2_024: Remove open file statements in MAM exchange code

fvitt:
Remove open file statements which are not needed 
Remove commented out debugging code

Closes issue #131

I have confirmed that the modal_aero_rename_no_acc_crs_sub routine has been exercised in our regression tests.  I don't know why this has not been a problem before.


cacraigucar:
@fvitt I know this needs to go into the development branch (so I added it to that project).  If it needs to go on the release branch as well (which I suspect is true), please add it to that project as well.

Issue #130, opened 2020-04-30
SCAM6 namelist definitions inconsistent


mikaelwitte:
### Summary of Issue:
The namelist variable scm_iop_lhflxshflxTg is specified as a logical, but string variables are given as valid values in namelist_definitions.xml. This is also the case for several other SCAM-specific namelist variables (scm_use_obs_qv,scm_force_latlon).
### Expected behavior and actual behavior:
An error is thrown at runtime that a logical (either .true. or .false.) is not a valid value for the variable.
### Steps to reproduce the problem. Include externals description file(s) and link to public repository):
Set scm_iop_lhflxshflxtg=.true. in user_nl_cam for any build using FSCAM compset.
### Screen output or log file showing the error message and context:
![Screen Shot 2020-04-30 at 2 58 44 PM](https://user-images.githubusercontent.com/11082590/80763364-350dbc00-8af3-11ea-8987-9c7eb56940c1.png)

gold2718:
@cacraigucar, these items have your name on them (probably erroneously)
@jtruesdal, do you know anything about how these variables got this way?
@mikaelwitte is using an older release branch tag but these entries seem to still be on the release branch and on cam_development.

jtruesdal:
Looks like a bug that's been in the code for a while.  These variables just must not be used much.  I can coordinate with cheryl and get a fix in.

PR #134, opened 2020-05-07
cam6_2_025?: Mods to resolve issue #130 SCAM6 namelist definitions inconsistent

jtruesdal:
The PR takes closes issue #130 where there were erroneous default values in the namelist definition file.  I also included some minor cleanup to the scam iop variables as well as corrected a bug for the upcoming SAS IOP file.

### DIFF
--- a/src/control/scamMod.F90
+++ b/src/control/scamMod.F90
@@ -164,11 +164,11 @@ logical, public ::  have_asdir             = .false. ! dataset contains asdir
 logical, public ::  have_asdif             = .false. ! dataset contains asdif
 logical, public ::  use_camiop             = .false. ! use cam generated forcing 
 logical, public ::  use_3dfrc              = .false. ! use 3d forcing
-logical, public ::  use_userdata           = .false. 
 logical, public ::  isrestart              = .false. ! If this is a restart step or not
   
 ! SCAM namelist defaults
 
+logical, public ::  scm_backfill_iop_w_init = .false. ! Backfill missing IOP data from initial file
 logical, public ::  scm_relaxation         = .false. ! Use relaxation
 logical, public ::  scm_crm_mode           = .false. ! Use column radiation mode
 logical, public ::  scm_cambfb_mode        = .false. ! Use extra CAM IOP fields to assure bit for bit match with CAM run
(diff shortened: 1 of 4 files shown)


jtruesdal:
I made a scam test on izumi with and without one of the problematical scam namelist variables and it built and ran without error.  The other mods only affect SCAM files or are within a if(single_column) statement.  Let me know what other testing you would like me to do.

cacraigucar:
I added this to the Projects for "CAM development branch".  Does this bug also need to be fixed on the "Release branch"?  If so, please add to the "Projects" section, the "release" project.

src/control/scamMod.F90
```diff
@@ -236,7 +236,7 @@ subroutine scam_readnl(nlfile,single_column_in,scmlat_in,scmlon_in)
        scm_cambfb_mode,scm_crm_mode,scm_zadv_uv,scm_zadv_T,scm_zadv_q,&
        scm_use_obs_T, scm_use_obs_uv, scm_use_obs_qv, &
        scm_relax_linear, scm_relax_tau_top_sec, &
-       scm_relax_tau_bot_sec, scm_force_latlon, scm_relax_fincl
+       scm_relax_tau_bot_sec, scm_force_latlon, scm_relax_fincl, scm_backfill_iop_w_init
```
gold2718, line 239: I would like to see more of these values logged (i.e., written to the log file). I realize that this has not been happening for SCAM over the years but now is as good a time as ever to make up for lost time.

nusbaume:
@cacraigucar It looks like issue #130 is already listed on the "release" project page.  So I don't think any new project info needs to be added until a PR is made to the release branch itself.  Unless of course I am missing something?

src/control/scamMod.F90
jtruesdal: Agreed.  I will update readnl to log the SCAM namelist parameters and take a look through the routines for other opportunities to refactor.

jtruesdal: @gold2718  SCAM namelist parameters now written to log file at end of readnl.

nusbaume:
@cacraigucar I removed the public release project link, as a PR can only be pushed to one branch on the repo, which in this case is `cam_development`.  @jtruesdal will need to make a second PR in order to add these changes to the release branch.  I should note that this is a fundamental git/github limitation, so there isn't much we can do about it without putting in a serious coding effort.  In fact the issue-closing script will throw an error if more than one project is associated with a particular PR.  I forgot to add that detail to the wiki, so I will do so now.

Issues, on the other hand, can be associated with as many projects as we want, which is the case with issue #130, which this PR is attempting to fix for the development branch.  

Anyways, if you are confused by any of this just let me know.  Thanks!

PR #135, opened 2020-05-08
Hotfix for issue closing script

### LINKED ISSUE #82: Need github action to automatically close issues
The current repository configuration for CAM means that most pull requests (PRs) will not automatically close issues via keywords, which is usually a standard github feature, because most pull requests are to the non-default branch.

To alleviate this problem, and thus make the CAM repository easier to manage, a github action should be created which, using the github API, automatically closes issues and PRs when referenced via keywords in pull requests that are merged to a non-default branch.

nusbaume:
I forgot that there are some project cards that are neither issues nor PRs, which caused the script to fail (I believe these were brought over when we first transitioned to github).

This quick fix allows the script to keep running even when it runs into those kind of cards.

closes #82

### DIFF
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -395,7 +395,7 @@ def _main_prog():
                     card_content = card.get_content()
 
                     #Next, check if card issue number matches any of the "close" issue numbers from the PR:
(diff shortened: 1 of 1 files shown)


nusbaume:
Should be good now.  Fingers crossed!

PR #146, opened 2020-06-04
cam6_2_029: Updates for using latest izumi compilers

### LINKED ISSUE #145: Updates to handle Izumi compiler and Cheyenne updates
Need to update the cime external and test_driver.sh to handle the updated libraries

cacraigucar:
cime and test_driver.sh need to be updated to handle new izumi compilers

Fixes #145 

Note that the EQ tests are failing   The summary is:
SUMMARY of cprnc:
 A total number of    297 fields were compared
          of which      0 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
 A total number of      2 fields could not be analyzed
 A total number of      5 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files DIFFER only in their field lists

My theory is that cprnc (which was recompiled for the new libraries) is giving a DIFF where before it was saying IDENTICAL when the number of fields differ.

Does anyone know if this statement is true?  If so, with Chris' testing tag coming soon, I am inclined for us to "live with this failure" or maybe even remove these tests from our testing for the next couple of tags before his

### DIFF
--- a/test/system/test_driver.sh
+++ b/test/system/test_driver.sh
@@ -207,7 +207,7 @@ cat > ${submit_script_cb} << EOF
 #PBS -q $CAM_BATCHQ
 #PBS -A $CAM_ACCOUNT
 #PBS -l walltime=2:00:00
-#PBS -l nodes=i007:select=1:ncpus=36:mpiprocs=36
+#PBS -l select=1:ncpus=36:mpiprocs=36
 #PBS -j oe
 #PBS -l inception=login
 
--- a/test/system/test_driver.sh
+++ b/test/system/test_driver.sh
@@ -207,7 +207,7 @@ cat > ${submit_script_cb} << EOF
 #PBS -q $CAM_BATCHQ
 #PBS -A $CAM_ACCOUNT
 #PBS -l walltime=2:00:00
-#PBS -l select=1:ncpus=36:mpiprocs=36
+#PBS -l nodes=i007:select=1:ncpus=36:mpiprocs=36
 #PBS -j oe
 #PBS -l inception=login
 
@@ -545,7 +545,7 @@ if [ "\$CAM_FC" = "INTEL" ]; then
     input_file="tests_pretag_izumi_nag"
     export CCSM_MACH="izumi_intel"
 elif [ "\$CAM_FC" = "NAG" ]; then
-    module load compiler/nag/6.2
+    module load compiler/nag/6.2-8.1.0
 
     export CFG_STRING="-cc mpicc -fc mpif90 -fc_type nag "
     export INC_NETCDF=\${NETCDF_PATH}/include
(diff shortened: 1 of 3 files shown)


fvitt:
Why is the cime update needed for this?

fischer-ncar:
Yes, cprnc was updated at some time to show this message.

cacraigucar:
> Why is the cime update needed for this?

@fischer-ncar will need this for the CESM testing tag, so I was just thinking to bring it in as well as it has the updated compilers for CESM tests.  

If I don't bring in cime, then I don't need to rerun the cheyenne tests.  I'm open to what everyone wants.

fischer-ncar:
I trust the new cime tag now.  So I would have no issues putting it in my tag.

fvitt:
Will the EQ tests continue to fail?

cacraigucar:
The EQ tests fail because of cprnc and will fail whether cime is updated or not.  The cprnc on izumi was recompiled to use the latest compilers (and was failing as it could not find the libraries which the the old compiled version used).    The problem is that the behavior of cprnc was changed.    I'm not sure if its worth the time to try to fix it, which is why I'm asking the question.

BTW, if someone wants to fix the EQ test, I am on PTO tomorrow, so someone else will need to take over making this tag if it drags out.

test/system/test_driver.sh
```diff
@@ -545,15 +545,15 @@ if [ "\$CAM_FC" = "INTEL" ]; then
     input_file="tests_pretag_izumi_nag"
     export CCSM_MACH="izumi_intel"
 elif [ "\$CAM_FC" = "NAG" ]; then
-    module load compiler/nag/6.2
+    module load compiler/nag/6.2-8.1.0
```
gold2718, line 548: I see this module on Izumi but have no idea what it means. Anyone??

cacraigucar:
@gold2718 Are you approving this with the EQ tests failing or should I remove them?

gold2718:
Which fields could not be analyzed?

fischer-ncar:
It looks like the fix to the EQ test is pretty simple.  just add the following in the right place.
elif grep -c "the two files DIFFER only in their field lists" cprnc.out > /dev/null; then
    echo "CAM_compare.sh: files are b4b"

fischer-ncar:
The right place is in CAM_compare.sh, forgot the mention that.

test/system/test_driver.sh
brian-eaton: The 8.1.0 is the version of gcc.

PR #148, opened 2020-06-05
cam6_2_030: Fix for floating point issue in GW drag.

### LINKED ISSUE #147: Fix for floating point issue in GW drag.
. Fix for a floating point issue that was responsible for instability in
  WACCM configurations with the SE dycore.

Note that this fix was originally included in issue #139, but has been
removed for testing purposes since it changes answers for configurations
other than WACCM with SE.  This fix should be added first.

brian-eaton:
fixes #147

Note: This fix is a workaround, the underlying source of the cascading error caused by small values still should be tracked down and fixed.

### DIFF
--- a/src/physics/cam/gw_common.F90
+++ b/src/physics/cam/gw_common.F90
@@ -619,6 +619,14 @@ subroutine gw_drag_prof(ncol, band, p, src_level, tend_level, dt, &
         ! new stress will be smaller than the old stress, causing stress
         ! divergence in the next layer down. This smoothes large stress
         ! divergences downward while conserving total stress.
+
+        ! Protection on SMALL gwut to prevent floating point
+        ! issues.
+        !--------------------------------------------------
+        where( abs(gwut(:,k,l)) < 1.e-15_r8 )
+           gwut(:,k,l) = 0._r8
+        endwhere   
+
         where (k <= tend_level)
            tau(:,l,k+1) = tau(:,l,k) + & 
                 abs(gwut(:,k,l)) * p%del(:,k) / gravit 


doc/ChangeLog
```diff
@@ -1,3 +1,55 @@
...
+Originator(s): juliob, eaton
+Date:
+One-line Summary: Fix for floating point issue in GW drag.
+
+Purpose of changes:
+
+. Fix a floating point issue in GW drag that seems to be responsible for
+  instabilities in WACCM configurations with the SE dycore.
+
+Bugs fixed (include bugzilla ID): none
```
gold2718, line 13: I propose changing this line in the template to: "Bugs fixed (including GitHub issue ID):"
(not for this tag but maybe for future tags).
@cacraigucar? @nusbaume?

src/physics/cam/gw_common.F90
```diff
@@ -619,6 +619,14 @@ subroutine gw_drag_prof(ncol, band, p, src_level, tend_level, dt, &
         ! new stress will be smaller than the old stress, causing stress
         ! divergence in the next layer down. This smoothes large stress
         ! divergences downward while conserving total stress.
+
+        ! Protection on SMALL gwut to prevent floating point
+        ! issues.
+        !--------------------------------------------------
+        where( abs(gwut(:,k,l)) < 1.e-15_r8 )
+           gwut(:,k,l) = 0._r8
+        endwhere
```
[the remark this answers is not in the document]
brian-eaton: Julio left a comment (which I removed according to normal procedure) to the effect of 'still need to track this down'.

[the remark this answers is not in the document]
gold2718: Perhaps a good approach is to open an issue explaining the nature of the issue (#147 does not really spell out the nature of the issue). Also, the ChangeLog should mention this.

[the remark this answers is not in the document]
brian-eaton: Not sure how to address this.  Issue #147 is complete with respect to the information that I have, and will be resolved by this PR which is to apply a fix (a workaround really) for a problem we don't understand (apparently some type of numerical instability).  I think it would be up to Julio to open an issue that would remain open if he wants to keep this problem on the radar and plans to look into it further in the future.  Does that make sense?

doc/ChangeLog
nusbaume: That sounds like a good idea to me, assuming bugzilla isn't being used anymore?  (I remember issues were moved from bugzilla to Github when we first transitioned, so I assume it's gone, but I just want to double-check).

doc/ChangeLog
```diff
@@ -1,3 +1,55 @@
...
+cheyenne/intel/aux_cam:
+
+izumi/nag:
+
+izumi/pgi:
+
+CAM tag used for the baseline comparison tests if different than previous
+tag:
+
+Summarize any changes to answers:
```
nusbaume, line 50: Doesn't this bug fix change answers?  If so, then should that be indicated here?

brian-eaton: I'll summarize answer changes when I know what they are, after testing has finished.

doc/ChangeLog
cacraigucar: Should this be changed to be Github issue and PR numbers (expanded from just bugs?)

@nusbaume Yes, all bugzilla bugs were transferred into github issues.

nusbaume: @cacraigucar Thanks for letting me know!  I've gone ahead and created an issue (#152)  that recommends updating the ChangeLog template.

PR #168, opened 2020-07-09
cam6_2_037?: Changes for Simple Models from Christiane

### LINKED ISSUE #143: Changes for Simple Models from Christiane
The existing logic allows the dycore initialization to adjust the atmosphere dry mass when simple physics is used with initial conditions read from a file (rather than being analytic).  Proposed changes are to modify EUL, FV, and SE dycores so the initialization does not change the dry mass of the atmosphere for simple physics configurations.

Modify EUL to not hardcode the dry mass for Held-Suarez.  This allows Held-Suarez configurations with topography as allowed by FV and SE.

Modify EUL to avoid the application of the energy fixer in case of tj2016_phys.

Update the NCAR/atmospheric_physics external to get fixes to kessler:
- fix relative humidity diagnostics with unphysical large values (> 100%)
- fix time step subcycling algorithm
- fix physical constants

The initialization of constituents was recently refactored to allow reading constituent fields (except Q) from an initial file even when the analytic IC option was used.  A problem arose with the FKESSLER compset because the initial file was from a cam5 simulation and contained the CLDLIQ field which overwrote the analytic IC specification for CLDLIQ.  The code should be modified so that no water specie is read from the initial file when the analytic IC option is in use.

### LINKED ISSUE #158: Update tag used for atmospheric_physics
NCAR/atmospheric_physics has a PR that needs to be brought into CAM for the release

### LINKED ISSUE #170: Bugs in SE/WACCM-X
WACCM-X using the SE dycore is under development and not supported in the CESM-2.2 release.  @PeterHjortLauritzen  has provided some bug fixes for the current code so that it now runs using a custom compset.

src/dynamics/se/dycore/prim_advance_mod.F90
. fix extents in some array args
. fix calc for grad_exner

src/dynamics/se/dycore/prim_advection_mod.F90
. fix extents in some array args

src/dynamics/se/dycore/prim_state_mod.F90
. fix some do loop extents

src/dynamics/se/dyn_comp.F90
. fix some do loop extents

src/utils/physconst.F90
. fix thermal_energy calc for variable dry air props
. fix some do loop extents

brian-eaton:
fixes #143
fixes #158
fixes #170 

Answers change for some simple physics configurations only.


gold2718:
> Bug fixes for SE/WACCM-X

Is this part of the Simpler models fix? If not, what issue is it addressing?

bld/namelist_files/use_cases/waccm_sc_2000_cam6.xml
```diff
@@ -23,6 +23,7 @@
 <!-- QBO settings -->
 <qbo_use_forcing >.true.</qbo_use_forcing>
 <qbo_use_forcing hgrid="0.9x1.25">.false.</qbo_use_forcing>
+<qbo_use_forcing dyn="se"        >.false.</qbo_use_forcing>
```
fvitt, line 26: I think this should be moved to cam_namelist_defaults.xml

src/physics/cam/constituents.F90
```diff
@@ -530,4 +531,24 @@ end function cnst_cam_outfld
...
+
+   cnst_is_water_specie = .false.
+
+   if (name == 'Q'      .or. &
+       name == 'CLDLIQ' .or. &
+       name == 'CLDICE' .or. &
+       name == 'RAINQM' .or. &
+       name == 'SNOWQM' .or. &
+       name == 'GRAUQM'      ) cnst_is_water_specie = .true.
+
```
fvitt, line 549: We will  need to remember to add to this list when more water constituents are added to the model or if different names are used.

brian-eaton:
> > Bug fixes for SE/WACCM-X
> 
> Is this part of the Simpler models fix? If not, what issue is it addressing?

These are fixes that Peter asked me to get into the release.  Since SE/WACCM-X is not supported in this release this is just for the convenience of the developers.

src/physics/cam/constituents.F90
brian-eaton: I agree.  The reason I added this routine to the constituents module is that this is where an improved implementation should occur.  There is already a cnst_species_class there and adding a class for water species would be a natural extension.  Then the cnst_add calls from the microphysics would be able to set the class.

gold2718:
Okay, then open an issue that describes the bugs and have this PR 'fix' them.

bld/namelist_files/use_cases/waccm_sc_2000_cam6.xml
brian-eaton: I would like to do that, but the way things are set up now it won't work.  All WACCM use case files currently set qbo_use_forcing and that setting takes precedence over cam_namelist_defaults.xml.  The way the WACCM use case files are set up needs to be refactored if we want to take advantage of being able to set qbo_use_forcing=.false. for all SE configurations.  I don't know whether that's what we want.  What do you think?

fvitt: The QBO nudging needs zonal means of winds.  Currently we know how to compute zonal means only if the FV dycore is used.  So qbo_use_forcing should be true only if the FV dycore is used.  I think all the use case files should be changed to include the dyn="fv' modifier for setting qbo_use_forcing to true.   Maybe this should be addressed in a future PR and leave this as you have it.

brian-eaton: I have opened issue #169 to address this.  In the meantime I've also added an SE specific entry to the waccmx_ma_2000_cam6 use case file as this is also a test that Peter has been using for SE/WACCM/WACCM-X development.

Externals_CAM.cfg
```diff
@@ -41,9 +41,9 @@ tag = pumas_cam-release_v1.3
 required = True
 
 [atmos_phys]
-tag = version0_00_005
+hash = d24df04
 protocol = git
-repo_url = https://github.com/NCAR/atmospheric_physics
+repo_url = https://github.com/cjablono/atmospheric_physics
```
cacraigucar, line 46: As has been pointed out, this needs to be brought into NCAR/atmospheric_physics.  What is the plan to get this accomplished?

src/dynamics/fv/dyn_comp.F90
```diff
@@ -45,7 +45,8 @@ module dyn_comp
 use physconst,          only: pi
 
 use pmgrid,             only: plon, plat
-use constituents,       only: pcnst, cnst_name, cnst_read_iv, qmin, cnst_type
+use constituents,       only: pcnst, cnst_name, cnst_read_iv, qmin, cnst_type, &
+                              cnst_is_water_specie
```
cacraigucar, line 49: At first glance, I thought there was a typo as I expected an "s" at the end of species.  I did a quick google search and there is a "nonstandard" use of specie which is singular for species.  Most places I looked said that species is both singular and plural.  Not being a scientist, perhaps our field has adopted this "nonstandard", in which case I'm okay with it.  Otherwise, I propose using the most common spelling and including the "s".   If indicating that this is singular is important, I would propose "cnst_is_a_water_species"

gold2718: On the other hand, the usual meaning of specie is "money in the form of coins rather than notes" so maybe we have constituents that are water coins? A new form of graupel?
`</snark>`
I second @cacraigucar's proposal.

Externals_CAM.cfg
brian-eaton: I assume the same procedure that resulted in the version0_00_006 tag (the relhum fix) will eventually produce the version0_00_007 tag (subcycle and constant fixes) that is needed here.

src/dynamics/fv/dyn_comp.F90
brian-eaton: I'll make this change.  I thought it sounded right to call CLDLIQ a specie in the class of water species.  So much for being a native speaker:(

brian-eaton: grep'ing the code for 'specie' turns up quite a few occurences, and they're probably entirely due to me:(  But I'm reformed now.

doc/ChangeLog
```diff
@@ -1,5 +1,152 @@
...
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+
+Fixes issue #143 Changes for Simple Models from Christiane
+Fixes issue #158 Update tag used for atmospheric_physics
+Fixes issue #170 fixes for SE/WACCM-X
+
+. The existing logic allows the dycore initialization to adjust the
+  atmosphere dry mass when simple physics is used with initial conditions
+  read from a file (rather than being analytic).  Proposed changes are to
```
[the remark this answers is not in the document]
brian-eaton: Done.

doc/ChangeLog
```diff
@@ -1,5 +1,152 @@
...
+
+. The initialization of constituents was recently refactored to allow
+  reading constituent fields (except Q) from an initial file even when the
+  analytic IC option was used.  A problem arose with the FKESSLER compset
+  because the initial file was from a cam5 simulation and contained the
+  CLDLIQ field which overwrote the analytic IC specification for CLDLIQ.
+  The code has been modified so that no water species is read from the
+  initial file when the analytic IC option is in use.
+
+. Bug fixes for SE/WACCM-X
```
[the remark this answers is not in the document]
brian-eaton: The line in the ChangeLog file that addresses this is:

Fixes issue #170 fixes for SE/WACCM-X

That's the issue I created in response to your request that the bullet in question be associated with an issue.  Not sure what you're asking for here.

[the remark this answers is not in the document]
gold2718: Just looking for some reference, e.g.
"Bug fixes for SE/WACCM-X (GitHub issue #170)"
or
"Bug fixes for SE/WACCM-X (see above)"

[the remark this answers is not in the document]
brian-eaton: Done.

Externals_CAM.cfg
brian-eaton: Updated to version0_00_007 in atmospheric_physic repo.

gold2718:
@cjablono, Do any of the Kessler tests in this PR exercise your relative humidity fix? If not, what would it take? The current tests are:
- f19_f19_mg17, 9 time steps
- ne5pg3_ne5pg3_mg37, 27 time steps
- f09_f09_mg17, 9 time steps

I believe all three tests are using the default configuration (e.g., analytic initial conditions). Any ideas?

PR #210, opened 2020-09-03
Modify Github Action push workflow so that it runs on all future branches.

### LINKED ISSUE #209: Github Action workflow only works on cam_development
Currently the Github Action workflow done after pull requests is set to only run on the `cam_development` branch.  However, we ideally want it to run on all branches.  This should be doable by simply removing the check for `cam_development` in the workflow YAML file.  

However, this will require that all branch project columns have the same column titles that are used in `cam_development`, which I will make sure of ahead of the pull request.

nusbaume:
Removes the branch check from the workflow, so that the action will run regardless of which branch the code changes are being pushed to.

Fixes #209

### DIFF
--- a/.github/workflows/branch_push_workflow.yml
+++ b/.github/workflows/branch_push_workflow.yml
@@ -1,16 +1,13 @@
 name: Pushed commit workflow
 
-on:
-  #For some reason GitHub does not allow secrets
-  #when a PR from a fork is closed.  Thus instead
-  #the workflow must activate whenever a commit
-  #is pushed to the repo.  Theoretically this should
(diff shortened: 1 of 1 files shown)


cacraigucar (approved):
This doesn't break when an issue resides on more than one project page, does it?  Probably the best test is to just put it in and monitor it.

nusbaume:
It shouldn't.  In theory it should simply move the issue card to the correct project column for each branch until there is only one branch remaining, at which point the issue should officially be closed.  I am hoping these PRs will act as a good test of this.

gold2718 (approved):

PR #212, opened 2020-09-03
Add Github Action push workflow to CESM2.1 release branch.

### LINKED ISSUE #209: Github Action workflow only works on cam_development

nusbaume:
Adds the Github Action push workflow and issue closing script to the CESM2.1 release branch.  This should allow for issues to be closed automatically via keywords for any PRs that are associated with CESM2.1.

Fixes #209

### DIFF
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -1,0 +1,477 @@
+#!/usr/bin/env python
+
+"""
+Script name:  branch_PR_issue_closer.py
+
+Goal:  To check if the newly-merged PR's commit message attempted to close an issue.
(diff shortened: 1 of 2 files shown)


gold2718 (approved):
Now that you are putting this in, Murphy says we will never touch this branch again :)

nusbaume:
Since @cacraigucar is on PTO I am merging this with the assumption that she is ok with it.

PR #227, opened 2020-09-23
 Misc bugs #1: rrtmg_state pointer, SCAM namelist, izumi compiler and TGIT test

### LINKED ISSUE #86: Cheyenne jobs randomly failing with deallocation

### LINKED ISSUE #130: SCAM6 namelist definitions inconsistent

### LINKED ISSUE #145: Updates to handle Izumi compiler and Cheyenne updates

cacraigucar:
Fix rrtmg_state pointer bug, correct SCAM6 namelist inconsistencies and update izumi compilers and TGIT test

Closes #86 
Closes #130 
Closes #145

Also updates TGIT regression test

### DIFF
--- a/src/control/scamMod.F90
+++ b/src/control/scamMod.F90
@@ -1,4 +1,23 @@
 module scamMod
+  !----------------------------------------------------------------------
+  !
+  ! This module provides control variables and namelist functionality
+  ! for SCAM.
+  !
+  ! As with global CAM, SCAM is initialized with state information
+  ! from the initial and boundary files. For each succeeding timestep
+  ! SCAM calculates the physics tendencies and combines these with
+  ! the advective tendencies provided by the IOP file to produce
+  ! a forcast. Generally, the control variables in this module
+  ! determine what data and parameterizations will be used to make
+  ! the forecast. For instance, many of the control variables in
+  ! this module provide flexibility to affect the forecast by overriding
+  ! parameterization prognosed tendencies with observed tendencies
+  ! of a particular field program recorded on the IOP file.
+  !
+  ! Public functions/subroutines:
+  !   scam_readnl
+  !-----------------------------------------------------------------------
 
 use shr_kind_mod,   only: r8 => shr_kind_r8
 use pmgrid,         only: plon, plat, plev, plevp
@@ -27,8 +46,6 @@ integer, parameter :: max_path_len = 128
 
 logical, public ::  single_column         ! Using IOP file or not
 logical, public ::  use_iop               ! Using IOP file or not
-logical, public ::  use_analysis
-logical, public ::  use_saveinit
 logical, public ::  use_pert_init         ! perturb initial values
 logical, public ::  use_pert_frc          ! perturb forcing 
 logical, public ::  switch(num_switches)  ! Logical flag settings from GUI
@@ -164,11 +181,11 @@ logical, public ::  have_asdir             = .false. ! dataset contains asdir
 logical, public ::  have_asdif             = .false. ! dataset contains asdif
 logical, public ::  use_camiop             = .false. ! use cam generated forcing 
 logical, public ::  use_3dfrc              = .false. ! use 3d forcing
-logical, public ::  use_userdata           = .false. 
 logical, public ::  isrestart              = .false. ! If this is a restart step or not
   
 ! SCAM namelist defaults
 
+logical, public ::  scm_backfill_iop_w_init = .false. ! Backfill missing IOP data from initial file
 logical, public ::  scm_relaxation         = .false. ! Use relaxation
 logical, public ::  scm_crm_mode           = .false. ! Use column radiation mode
 logical, public ::  scm_cambfb_mode        = .false. ! Use extra CAM IOP fields to assure bit for bit match with CAM run
@@ -220,14 +237,11 @@ subroutine scam_readnl(nlfile,single_column_in,scmlat_in,scmlon_in)
 
   ! Local variables
   character(len=*), parameter :: sub = 'scam_readnl'
-  integer :: unitn, ierr
+  integer :: unitn, ierr, i
   integer  :: ncid
-  integer  :: latdimid, londimid
-  integer  :: latsiz, lonsiz
-  integer  :: latid, lonid
   integer  :: iatt
-  integer  :: ret
   integer  :: latidx, lonidx
+  logical  :: adv
   real(r8) :: ioplat,ioplon
 
 ! this list should include any variable that you might want to include in the namelist
@@ -236,7 +250,7 @@ subroutine scam_readnl(nlfile,single_column_in,scmlat_in,scmlon_in)
        scm_cambfb_mode,scm_crm_mode,scm_zadv_uv,scm_zadv_T,scm_zadv_q,&
        scm_use_obs_T, scm_use_obs_uv, scm_use_obs_qv, &
        scm_relax_linear, scm_relax_tau_top_sec, &
-       scm_relax_tau_bot_sec, scm_force_latlon, scm_relax_fincl
+       scm_relax_tau_bot_sec, scm_force_latlon, scm_relax_fincl, scm_backfill_iop_w_init
 
   single_column=single_column_in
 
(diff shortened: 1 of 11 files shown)


fvitt (approved):

brian-eaton (approved):

nusbaume (approved):
Looks good, and thanks for fixing the TGIT error codes!  Just one question regarding a print statement.

src/control/scamMod.F90
```diff
@@ -303,6 +317,46 @@ subroutine scam_readnl(nlfile,single_column_in,scmlat_in,scmlon_in)
...
+              if (adv) then
+                 write (iulog, "(A18)") "'"//trim(scm_relax_fincl(i))//"',"
+              else
+                 write (iulog, "(A18)", ADVANCE="NO") "'"//trim(scm_relax_fincl(i))//"',"
+              end if
+           else
+              exit
+           end if
+        end do
+        print *
```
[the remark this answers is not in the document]
cacraigucar: Exactly - I saw that too!   I decided that because I am applying the exact changes from cam_development, I would leave it in.  We all missed it the first time around, but not now!

jtruesdal (approved):

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
 ===============================================================
 
+Tag name:cam_cesm2_1_rel_46
+Originator(s): cacraig, fvitt, jet
+Date:
+One-line Summary: Fix rrtmg_state pointer bug, correct SCAM6 namelist inconsistencies and update izumi compilers and TGIT test
+
+Purpose of changes:
+Closes:
+    Workaround rrtmg_state pointer issue on cheyenne (issue #86)
```
gold2718, line 11: The new ChangeLog template says to include the issue text, but that is not what is written here. In fact, this text should be down in the modified section where an explanation of the changes should be.
Speaking of the new ChangeLog template, it would not hurt to update that in this branch for future maintenance tags.

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
+    SCAM6 namelist definitions inconsistent (issue #130)
+    Updates for using latest izumi compilers (issue #146)
```
gold2718, line 13: This is a PR, not an issue. There should be an issue for these changes.

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
+Code reviewed by:
+
+List all files eliminated:
+
+List all files added and what they do:
+
+List all existing files that have been modified, and describe the changes:
+        modified:   src/physics/rrtmg/radiation.F90
+        modified:   src/physics/rrtmg/rrtmg_state.F90
+          - Mods for issue #86
```
gold2718, line 35: A summary of the change would be much better here.

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
+
+        modified:   bld/namelist_files/namelist_definition.xml
+        modified:   src/control/scamMod.F90
+        modified:   src/dynamics/eul/iop.F90
+        modified:   src/physics/cam/physpkg.F90
+          - Mods for issue #130
```
gold2718, line 41: A summary of the change would be much better here.

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
+
+        modified:   Externals.cfg
+        modified:   test/system/CAM_compare.sh
+        modified:   test/system/test_driver.sh
+          - Mods for issue #146
```
gold2718, line 46: A summary of the change would be much better here.

doc/ChangeLog
```diff
@@ -1,6 +1,94 @@
...
+
+        modified:   TGIT.sh
+          - Update the TGIT CAM regression test
```
gold2718, line 49: Why did it need an update? Is there an issue on that?

src/control/scamMod.F90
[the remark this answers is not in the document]
gold2718: So is it going away? I think it should.

test/system/TGIT.sh
```diff
@@ -67,16 +68,26 @@ EOF
...
     check_code "$?" "Problem running 'find' command for multi-git file check."
 
-    if [ "${git_file_num}" -gt 2 ]; then
+    if [ "${git_file_num}" -gt 3 ]; then
         cat <<EOF
-More than two ".git*" files or sub-directories present in this CAM git repo.
+More than three ".git*" files or sub-directories present in this CAM git repo.
 EOF
-        rc=4
+        rc=5
```
gold2718, line 90: Note for the future. This seems like a losing game. Perhaps a better approach would be to take a peek in Externals.cfg and maybe also Externals_CAM.cfg and check that those directories, and only those directories contain .git files.
Another check could be to use `manage_externals/checkout_externals --status` to make sure that all the .git files that are supposed to be there are actually working clones in a clean state.

gold2718 (changes requested):
Mostly documentation issues but I think that stray print statement should go away.

doc/ChangeLog
cacraigucar: I am only addressing the open issues on the CESM2.1 branch.  Since there was no issue to update the ChangeLog_template, I did not bring it in.   I will make an issue to bring it in with my next tag in this branch.

doc/ChangeLog
cacraigucar: I will copy the issue text here and for the following.

doc/ChangeLog
cacraigucar: I will copy the issue text here

doc/ChangeLog
cacraigucar: I will copy the issue text here

doc/ChangeLog
cacraigucar: The test failed when I ran it.  I had looked for an issue that I could reference, but it was buried in issue #82.  I decided to not reference that issue as the thrust of the issue was not this modification.  I also thought since I was fixing the test, I wouldn't bother with making an issue for it.

In hindsight, since issue #82 included a change to the CAM source tree, it should not have been committed without making a tag and an updated ChangeLog.  I was unable to find this change via the ChangeLog, but rather had to use the github log.  I believe all CAM source changes should be documented in the ChangeLog.

test/system/TGIT.sh
cacraigucar: This is what is currently on cam_development.   If a change is needed then an issue needs to be opened.   I will not be making any changes now, as the test is passing.

src/control/scamMod.F90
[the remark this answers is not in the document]
cacraigucar: I will remove the print statement.  If someone wants it removed from cam_development, then an issue needs to be opened for it.

doc/ChangeLog
nusbaume: @cacraigucar if you can also change the PR number to the correct issue number in the keyword text above that would be great, as it will help us avoid having to close the issue manually later.  Thanks!

test/system/TGIT.sh
nusbaume: @gold2718 I might disagree with that assessment.  Right now this test only examines the `.git*` files and directories in the top-level CAM directory, which should be quite limited (right now it is only `.git`, `.gitignore`, and `.github`.  I know there are a few other `.git*` file types out there, but I can't imagine there are too many more that we would add to the repo?  

Also, I can of course modify the script to examine the externals as well, but I wasn't sure if this test should cover them, or if they should be the responsibility of their own repo's tests?  Either way is fine with me.

doc/ChangeLog
```diff
@@ -1,6 +1,100 @@
 
 ===============================================================
 
+Tag name: cam_cesm2_1_rel_46
+Originator(s):  cacraig, fvitt, jet
+Date:
+One-line Summary:  Fix rrtmg_state pointer bug, correct SCAM6 namelist inconsistencies and update izumi compilers and TGIT test
+Github PR URL:
```
gold2718, line 8: Where is the URL?

gold2718 (approved):
Either the ChangeLog template is wrong or you need to include a GitHub PR URL near the top.

doc/ChangeLog
```diff
@@ -1,6 +1,100 @@
...
+   Misc bugs #1: rrtmg_state pointer, SCAM namelist, izumi compiler and TGIT test
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+    cheyenne jobs randomly failing with deallocation [rrtmg_state pointer bug] (issue #86)
+    SCAM6 namelist definitions inconsistent (issue #130)
+    Updates to handle izumi compiler updates (issue #145)
+
```
gold2718, line 15: Do you need to include #228 here?

nusbaume (approved):
Should we create an issue to remind us to remove the erroneous `print` statement from the other branches as well?

doc/ChangeLog
cacraigucar: Sorry - first time filling out the template and didn't register the "URL".  Added

doc/ChangeLog
cacraigucar: Forgot to add that documentation.  Added both here and to the PR text.

PR #232, opened 2020-09-26
cam6_3_012: Add Horizontal Momentum Tendency Budget Variables

### LINKED ISSUE #226: Add Horizontal Momentum Tendency Budget Variables
I'd like to add modifications I've already made that output U and V tendencies for all parameterizations and the dynamical core. While a temperature and moisture budget can be output by setting history_budget = .true., existing horizontal momentum tendency variables available for output have inconsistent names and by no means make a complete, closed budget. Pull request to come soon.

Mods will allow outputing UTEND_TOT, UTEND_CORE, UTEND_PHYSTOT, UTEND_MACROP, UTEND_VDIFF, UTEND_DCONV, UTEND_SHCONV, UTEND_GWDTOT, UTEND_NDG, UTEND_RAYLEIGH, UTEND_QBORLX, UTEND_LUNART, and UTEND_IONDRG. Same for V as well.

Features:

- Some of the tendency variables above may be zero, depending on the CAM configuration. However, setting history_budget = .true. will output all of them so that the sum of all the UTEND_* variables (except for UTEND_PHYSTOT and UTEND_TOT) is equal UTEND_TOT regardless of configuration. That said, I have only tested this in SCAM and with in nudged runs with the FV dycore.

- As mentioned, by default, these variables will not be output. However, with history_budget = .true., a closed U and V tendency budget will be output in addition to the existing, untouched T and Q budgets.

Caveats:

- Some of these tendencies can already be output (e.g. UTGW_TOTAL = UTEND_GWDTOT, UTEND_CLUBB = UTEND_MACROP for default CAM6, ZMMTU = UTEND_DCONV). I have not altered existing variables, but rather added variables with a consistent naming convention. Some of the added variables will be redundant.

- Some parameterizations currently do not influence horizontal momentum, but may in the future (e.g. microphysics, dry adjustment). Additional variables would need to added should such development ever occur.

- Important: time integrating UTEND_TOT **will not** give the instantaneous U. However, time integrating UTEND_TOT **will** give instantaneous UAP (U after physics). This is because U is output in the middle of physics (after deep convection but before radiation). UAP is essentially U at the end of the time step loop (at least for the FV core...).

siouxyalie:
Mods allow outputing UTEND_TOT, UTEND_CORE, UTEND_PHYSTOT, UTEND_MACROP, UTEND_VDIFF, UTEND_DCONV, UTEND_SHCONV, UTEND_GWDTOT, UTEND_NDG, UTEND_RAYLEIGH, UTEND_QBORLX, UTEND_LUNART, and UTEND_IONDRG. Same for V as well.

Features:

Some of the tendency variables above may be zero, depending on the CAM configuration. However, setting history_budget = .true. will output all of them so that the sum of all the UTEND_* variables (except for UTEND_PHYSTOT and UTEND_TOT) is equal UTEND_TOT regardless of configuration. That said, I have only tested this in SCAM and with in nudged runs with the FV dycore.

As mentioned, by default, these variables will not be output. However, with history_budget = .true., a closed U and V tendency budget will be output in addition to the existing, untouched T and Q budgets.

Caveats:

Some of these tendencies can already be output (e.g. UTGW_TOTAL = UTEND_GWDTOT, UTEND_CLUBB = UTEND_MACROP for default CAM6, ZMMTU = UTEND_DCONV). I have not altered existing variables, but rather added variables with a consistent naming convention. Some of the added variables will be redundant.

Some parameterizations currently do not influence horizontal momentum, but may in the future (e.g. microphysics, dry adjustment). Additional variables would need to added should such development ever occur.

Important: time integrating UTEND_TOT will not give the instantaneous U. However, time integrating UTEND_TOT will give instantaneous UAP (U after physics). This is because U is output in the middle of physics (after deep convection but before radiation). UAP is essentially U at the end of the time step loop (at least for the FV core...).

Default averaging flag for UAP, VAP (U,V after physics) is "A". I think it makes more sense for these variables to be output instantaneously. This way, time integrating the time averaged UTEND_TOT will give UAP. However, it looks like the the averaging flag set by avgflat_pertape overrides the default set by addfld, making the potential change from "A" to "I" in UAP, VAP addfld calls irrelevant. So, solely setting history_budget = .true. will output a closed budget, but time integrating this budget will not give U,V or UAP,VAP. This is potentially very confusing. I did not want to alter any existing code, nor did I know how to somehow make it so UAP, VAP are only ever output instantaneously, so I'll leave it up to the real developers to decide.

resolves #226


cacraigucar (changes requested):
Once you make the requested changes and push them back to your fork, please ask for a rereview from me (hit the two circular arrows beside my name in the reviewer section of the PR).  At that point, I will ask for review from the other CAM SE's.

cacraigucar (approved):

src/physics/cam/check_energy.F90
```diff
@@ -193,12 +199,19 @@ subroutine check_energy_init()
     call addfld('TEFIX',  horiz_only,  'A', 'J/m2', 'Total energy after fixer')
     call addfld('EFIX',   horiz_only,  'A', 'W/m2', 'Effective sensible heat flux due to energy fixer')
     call addfld('DTCORE', (/ 'lev' /), 'A', 'K/s' , 'T tendency due to dynamical core')
+    call addfld('UTEND_CORE', (/ 'lev' /), 'A', 'm/s2' , 'U tendency due to dynamical core')
+    call addfld('VTEND_CORE', (/ 'lev' /), 'A', 'm/s2' , 'V tendency due to dynamical core')
+    call register_vector_field('UTEND_CORE','VTEND_CORE')
```
fvitt, line 204: Inside the check_energy module seems like an odd place for these addfld calls for dycore tendencies diagnostics.  Should those be with the other added addfld calls in cam_diagnostics or in phys_init (in the some module source file as the corresponding outfld calls)?

cacraigucar: Good catch.  @gold2718 and @nusbaume  - what do you think?

nusbaume: I agree, and would probably vote to move them to `cam_diagnostics`, mostly because I think `physpkg` has enough going on, and I don't want to start adding extra `addfld` calls to it if we can avoid it.

However, would this move include `DTCORE`, which I suspect the added tendencies were modeled after?  Does anyone know if there is a reason for `DTCORE` to be located in `check_energy`?

gold2718: Our usual standard is that the `addfld` call should be in the same module as the corresponding `outfld` call. Otherwise, it risks being removed in a cleanup pass.

nusbaume: @gold2718 that's a good point!  However, in that case I might vote to move most of the `addfld` and `outfld` calls to their respective physics parameterization interfaces.  For example, I would move `UTEND_RAYLEIGH` inside `rayleigh_friction_tend`, instead of having it exist in `tphysac`.  

Obviously some of the quantities, like `UTEND_TOT`, may not have a clear location, in which case I am happy to put them in either `cam_diagnostics` or somewhere in `physpkg` (whatever we decide).

gold2718: @nusbaume, right. If an output quantity is already being computed (or something very close to it is being computed) in a physics parameterization, the `addfld` and `outfld` calls should go there. That will be especially for items derived from parameterization tendencies. For the 'total tendency' quantities, I'm tempted to say they should be in their own module -- cam_diagnostics is pretty full of all kinds of stuff.

siouxyalie: @fvitt, @nusbaume - Yeah. I largely followed the existing code for the temperature tendency budget. I could not tell you why DTCORE is in check_energy.

Whatever is done with ?TEND_* variables should probably be done to the temperature and q tendency variables, I'd guess.

I put the outfld calls before nearly every physics_update. This just seemed like the fastest/easiest place to do so. Another perk of having outfld calls there allows the physics tendency to be output regardless of the actual scheme chosen (e.g. if you use CLUBB or not, you still output the macrophysics tendency) and with a consistent name. This way, if a user decided to go nuts and change schemes, you still get a tendency for the particular parameterized processes. Maybe this is extremely unlikely to occur. But, coming from the WRF world, I thought it would be much more sensible to output tendencies named by the parameterized process as opposed to being named by each scheme chosen.

But yeah. These tendencies are sometimes redundant (e.g. UTEND_MACROP = UTEND_CLUBB when you use CLUBB).

Whatever you decide goes, though. Y'all are the experts.

nusbaume: @siouxyalie I think having the tendency named after the process instead of the particular parameterization makes complete sense, at least for this particular diagnostic. 

However, you can still go one-level down, at least for many of these processes, as subroutines like `convect_shallow_tend` and `vertical_diffusion_tend` are designed to manage multiple different parameterizations (e.g. between CAM4 and CAM5 physics) in the same location, so the diagnostic can still be independent of the exact parameterization chosen.  I agree that the same should be done for the temperature and humidity tendency diagnostics as well, although we don't have to do that for this PR unless you are really feeling up to it (as we can always just make an issue instead and save it for later).

For `?TEND_MACROP`, the argument could be made that labeling CLUBB just "MACROP" is not really accurate, as CLUBB also includes shallow convection and vertical diffusion. So if someone was comparing CAM4 to CAM6 and wasn't knowledgeable it would look as if CAM6 had no shallow convection but a huge cloud macrophysics tendency, when in reality this isn't actually the case.

So, I might argue that `?TEND_MACROP` should only represent the tendencies from `macrop_driver_tend`, and just use     `TEND_CLUBB` for CLUBB (as that would remove the redundancy).  I also believe you could pass in the `macmic` sub-stepping info into `macrop_driver_tend` similar to what is done for `clubb_tend_cam` and calculate the (properly scaled) tendencies there, again like what is done for `?TEND_CLUBB`.  However, I don't know this section of the code that well, so it is entirely possible that I am missing something (maybe the other SEs can chime in if they think that is wise or not)?

Finally, if I am the only one pushing for this larger change and the other SEs don't think it is necessary then I am happy to concede, especially if it holds up this PR for too long.

cacraigucar: I believe very strongly that the budget terms should all be consistently written out in the tphysac/tphysbc routines. 
 
One constraint of these budget terms is that adding up all the individual budget terms should yield the total budget term.  After much discussion between @siouxyalie and I, we determined that this is up to the user to verify - setting up a test was becoming very cumbersome.  He also thought it was standard practice for users to do this verification.

With that in mind, if a new parameterization is added and the individual budget terms for that parameterization are not output the total budget will not match.  At that point unless the outputting of the budget terms are in a single routine, the user will have a more difficult task of trying to identify the parameterization which is missing the term.  Now it is a simple matter of scanning through the calls and seeing which is missing the outfld calls.  Also, as people love to "find examples", this approach will prompt them to create budget terms for the new parameterization.

With that in mind, if the outfld calls are in tphysac/tphysbc, @gold2718 where would you like the addfld calls to reside?  The  outfld calls I see at this level have their addfld calls in check_energy (as @siouxyalie) pointed out.

src/physics/cam/physpkg.F90
```diff
@@ -2020,6 +2059,10 @@ subroutine tphysbc (ztodt, state,  &
 
     ifld   =  pbuf_get_index('DTCORE')
     call pbuf_get_field(pbuf, ifld, dtcore, start=(/1,1,itim_old/), kount=(/pcols,pver,1/) )
+    ifld   =  pbuf_get_index('UTEND_CORE')
```
gold2718, line 2062: `pbuf_get_index` is slow and should be done at initialization time (stored in a module variable for run time use).

gold2718 (changes requested):
Not a complete review but a couple of items that need to be addressed.

src/physics/cam/check_energy.F90
cacraigucar: No problem other than it shows my indecisiveness of which rule we were going to break - the addfld being in the same module or keeping the interfaces in physpkg "clean".  After thinking about it more,  I agree with your confirmation of the proper place for the addfld calls.  My alternate approach doesn't look so good in the light of Monday morning.

@siouxyalie - this means that you will pull all of your addfld calls for your new outfld calls out of the places where you put them and put them all in the phys_init routine inside physpkg.F90.  I would suggest you make a section at the very end of the routine with them.  You will also need to add at the top of phys_init a "use cam_history, only: addfld".

siouxyalie: Thanks for this guidance. On it.

src/physics/cam/physpkg.F90
siouxyalie: Done and done.

src/physics/cam/physpkg.F90
```diff
@@ -1789,6 +1898,8 @@ subroutine tphysac (ztodt,   cam_in,  &
     ! store T in buffer for use in computing dynamics T-tendency in next timestep
```
nusbaume, line 1898: Might be worth updating this comment to include wind as well.

src/physics/cam/physpkg.F90
```diff
@@ -1521,6 +1602,10 @@ subroutine tphysac (ztodt,   cam_in,  &
          (trim(cam_take_snapshot_before) == trim(cam_take_snapshot_after))) then
        call cam_snapshot_ptend_outfld(ptend, lchnk)
     end if
+    if ( ptend%lu .and. ptend%lv ) then
```
nusbaume, line 1605: For this and the other `ptend%lx` checks below, I believe you should split them like so:

```
if (ptend%lu) then
   call outfld( 'UTEND_XXXX`, ptend%u, pcols, lchnk)
end if
if (ptend%lv) then
   call outfld(`VTEND_XXXX`, ptend%v, pcols, lchnk)
end if
```

That way if for some reason a future parameterization has only one wind direction being modified (which would be weird, but you never know), the diagnostic would still work as expected.

src/physics/cam/physpkg.F90
```diff
@@ -2085,7 +2201,11 @@ subroutine tphysbc (ztodt, state,  &
     ! T tendency due to dynamics
```
nusbaume, line 2201: Might be worth updating this comment to include wind as well.

src/physics/cam/physpkg.F90
```diff
@@ -2273,6 +2402,10 @@ subroutine tphysbc (ztodt, state,  &
        prec_pcw_macmic = 0._r8
        snow_pcw_macmic = 0._r8
 
+       ! initialize ptend structures where macro and microphysics tendencies are
+       ! accumulated over macmic substeps
+       call physics_ptend_init(ptend_macp_all,state%psetcols,'macrophysics',ls=.false.,lu=.true.,lv=.true.)
```
nusbaume, line 2407: I believe the `ls=.false.` is un-necessary here (it is automatically assumed to be false if not present).

nusbaume (changes requested):
Mostly looks good to me,  but I do have some (hopefully minor) requests.

src/physics/cam/cam_diagnostics.F90
```diff
@@ -195,6 +199,8 @@ subroutine diag_init_dry(pbuf2d)
 
     ! State before physics
     call addfld ('TBP',     (/ 'lev' /), 'A','K',             'Temperature (before physics)')
+    call addfld ('UBP',     (/ 'lev' /), 'A','m/s2',          'Zonal wind (before physics)')
+    call addfld ('VBP',     (/ 'lev' /), 'A','m/s2',          'Meridional Wind (before physics)')
```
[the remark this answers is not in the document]
siouxyalie: Yes. Good catch.

src/physics/cam/cam_diagnostics.F90
```diff
@@ -2043,19 +2073,31 @@ subroutine diag_phys_tend_writeout_dry(state, pbuf,  tend, ztodt)
...
 
     !! calculate and outfld the total temperature tendency
     ftem3(:ncol,:) = (state%t(:ncol,:) - t_ttend(:ncol,:))/ztodt
     call outfld('TTEND_TOT', ftem3, pcols, lchnk)
+    ftem3(:ncol,:) = (state%u(:ncol,:) - t_utend(:ncol,:))/ztodt
+    call outfld('UTEND_TOT', ftem3, pcols, lchnk)
+    ftem3(:ncol,:) = (state%v(:ncol,:) - t_vtend(:ncol,:))/ztodt
+    call outfld('VTEND_TOT', ftem3, pcols, lchnk)
 
     !! update physics buffer with this time-step's temperature
```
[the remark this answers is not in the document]
siouxyalie: Done

src/physics/cam/cam_diagnostics.F90
```diff
@@ -2043,19 +2073,31 @@ subroutine diag_phys_tend_writeout_dry(state, pbuf,  tend, ztodt)
...
 
     ! Total (physics+dynamics, everything!) tendency for Temperature
 
     !! get temperature stored in physics buffer
     itim_old = pbuf_old_tim_idx()
     call pbuf_get_field(pbuf, t_ttend_idx, t_ttend, start=(/1,1,itim_old/), kount=(/pcols,pver,1/))
+    call pbuf_get_field(pbuf, t_utend_idx, t_utend, start=(/1,1,itim_old/), kount=(/pcols,pver,1/))
+    call pbuf_get_field(pbuf, t_vtend_idx, t_vtend, start=(/1,1,itim_old/), kount=(/pcols,pver,1/))
 
     !! calculate and outfld the total temperature tendency
```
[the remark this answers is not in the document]
siouxyalie: Done

src/physics/cam/check_energy.F90
```diff
@@ -173,7 +179,7 @@ subroutine check_energy_init()
 ! Initialize the energy conservation module
 !
 !-----------------------------------------------------------------------
-    use cam_history,       only: addfld, add_default, horiz_only
+    use cam_history,       only: addfld, add_default, horiz_only, register_vector_field
```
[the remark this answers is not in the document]
siouxyalie: Removed. Good catch.

src/physics/cam/physpkg.F90
siouxyalie: Done

src/physics/cam/physpkg.F90
```diff
@@ -952,6 +961,74 @@ subroutine phys_init( phys_state, phys_tend, pbuf2d, cam_in, cam_out )
...
+    call addfld ( 'VTEND_LUNART', (/ 'lev' /), 'A', 'm/s2', 'Meridional wind tendency by lunar tides')
+    call register_vector_field ( 'UTEND_LUNART', 'VTEND_LUNART')
+    call addfld ( 'UTEND_IONDRG', (/ 'lev' /), 'A', 'm/s2', 'Zonal wind tendency by ion drag')
+    call addfld ( 'VTEND_IONDRG', (/ 'lev' /), 'A', 'm/s2', 'Meridional wind tendency by ion drag')
+    call register_vector_field ( 'UTEND_IONDRG', 'VTEND_IONDRG')
+    call addfld ( 'UTEND_NDG', (/ 'lev' /), 'A', 'm/s2', 'Zonal wind tendency by nudging')
+    call addfld ( 'VTEND_NDG', (/ 'lev' /), 'A', 'm/s2', 'Meridional wind tendency by nudging')
+    call register_vector_field ( 'UTEND_NDG', 'VTEND_NDG')
+    call addfld('UTEND_CORE', (/ 'lev' /), 'A', 'm/s2' , 'U tendency due to dynamical core')
+    call addfld('VTEND_CORE', (/ 'lev' /), 'A', 'm/s2' , 'V tendency due to dynamical core')
```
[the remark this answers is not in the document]
siouxyalie: Done.

src/physics/cam/physpkg.F90
siouxyalie: I think if one component is output, the other should be as well. If for some odd reason a parameterization only modifies one component, having the other output as well, even if all zeros, would be more clear than having it missing. I suppose this would require changing the ".and."  to ".or."?

nusbaume: @siouxyalie I completely agree with you.  The problem is that the `ptend`  variables are only allocated if the logical is true (for example, `ptend%v` only exists if `ptend%lv` is true).  So if an `.or.` is used and one of them isn't allocated it will most likely result in a runtime failure.  The code above was the simplest way I could think of to get around that while still allowing for the existing tendency to be output. 

If you really wanted to have zeroes output when the logical is false then you could instead create a new variable in `physpkg` with the same dimensions as `ptend%x` that is all zeroes, and then use that in the `outfld` call.  I am happy with either choice, so I'll let you decide.

siouxyalie: Gotchya. Thanks. I think I will go with your original suggestion, then.

nusbaume (approved):
Looks great now, thanks!  Just a couple last comments which appear to have not been updated.

src/physics/cam/cam_diagnostics.F90
```diff
@@ -2043,19 +2074,31 @@ subroutine diag_phys_tend_writeout_dry(state, pbuf,  tend, ztodt)
...
-    !! calculate and outfld the total temperature tendency
+    !! calculate and outfld the total temperature, U, and V tendencies
     ftem3(:ncol,:) = (state%t(:ncol,:) - t_ttend(:ncol,:))/ztodt
     call outfld('TTEND_TOT', ftem3, pcols, lchnk)
+    ftem3(:ncol,:) = (state%u(:ncol,:) - t_utend(:ncol,:))/ztodt
+    call outfld('UTEND_TOT', ftem3, pcols, lchnk)
+    ftem3(:ncol,:) = (state%v(:ncol,:) - t_vtend(:ncol,:))/ztodt
+    call outfld('VTEND_TOT', ftem3, pcols, lchnk)
 
     !! update physics buffer with this time-step's temperature
```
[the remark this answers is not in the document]
siouxyalie: Missed that. Fixed.

src/physics/cam/check_energy.F90
```diff
@@ -66,6 +66,8 @@ module check_energy
 
   integer  :: teout_idx  = 0       ! teout index in physics buffer
   integer  :: dtcore_idx = 0       ! dtcore index in physics buffer
+  integer  :: ducore_idx = 0       ! dtcore index in physics buffer
+  integer  :: dvcore_idx = 0       ! dtcore index in physics buffer
```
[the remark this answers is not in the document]
siouxyalie: Don't know how I missed this! Fixed.

nusbaume (approved):
Thanks, everything looks good to me now!

gold2718 (approved):
Looks good now, thanks!

fvitt (approved):

PR #240, opened 2020-10-05
Replace serial reads with infld calls

### LINKED ISSUE #237: Replace deprecated physics grid interfaces in nudging input
nudging.F90 currently uses deprecated physics grid interfaces (e.g., `scatter_field_to_chunk`). These non-scalable interfaces will be removed so nudging.F90 should convert its input to use `infld`.

patcal:
This is an incremental revision to the nudging.F90 module which replaces serial reads with infld() calls so that PIO is used to read values on known model grids from an open netcdf file. This update does not address the need for in-line horizontal/vertical interpolations. 

fixes #237


gold2718:
@patcal,
Before reviewing this code, please close this PR and open a new one with the base being ESCOMP:cam_development (ESCOMP:cam-cesm2_2_rel is for CESM2.2 support, cam_development is where all our current development takes place).

PR #241, opened 2020-10-06
Replace serial reads with infld calls.

### LINKED ISSUE #237: Replace deprecated physics grid interfaces in nudging input

patcal:
This is an incremental revision to the nudging.F90 module which replaces serial reads with infld() calls so that PIO is used to read values on known model grids from an open netcdf file. This update does not address the need for in-line horizontal/vertical interpolations.

Fixes #237

### DIFF
--- a/src/physics/cam/nudging.F90
+++ b/src/physics/cam/nudging.F90
@@ -1395,7 +1395,7 @@ contains
    use pio             ,only: PIO_BCAST_ERROR,PIO_INTERNAL_ERROR
    use pio             ,only: pio_closefile,pio_seterrorhandling,file_desc_t
    use ncdio_atm       ,only: infld
-   use cam_grid_support,only: cam_grid_id,cam_grid_get_dim_names
+   use cam_grid_support,only: cam_grid_id,cam_grid_get_dim_names,DLEN=>max_hcoordname_len
 
    ! Arguments
    !-------------
@@ -1403,12 +1403,12 @@ contains
 
    ! Local values
    !-------------
-   type(file_desc_t):: fileID
-   integer          :: nn,Nindex
-   logical          :: VARflag
-   integer          :: grid_id
-   character(len=8) :: dim1name,dim2name
-
+   type(file_desc_t)  :: fileID
+   integer            :: nn,Nindex
+   logical            :: VARflag
+   integer            :: grid_id
+   character(len=DLEN):: dim1name,dim2name
+   integer            :: err_handling
 
    real(r8),allocatable:: Tmp3D(:,:,:)
    real(r8),allocatable:: Tmp2D(:,:)
@@ -1436,13 +1436,11 @@ contains
    ! Open the file and get the fileID.
    !-------------------------------------
    call cam_pio_openfile(fileID,trim(anal_file),0)
-   call pio_seterrorhandling(fileID,PIO_BCAST_ERROR   )
-   call pio_seterrorhandling(fileID,PIO_INTERNAL_ERROR)
+   call pio_seterrorhandling(fileID,PIO_BCAST_ERROR,oldmethod=err_handling)
    if(masterproc) write(iulog,*)'PIO_OPEN: file=',trim(anal_file)
 
    grid_id = cam_grid_id('physgrid')
    call cam_grid_get_dim_names(grid_id,dim1name,dim2name)
-!   if(masterproc) write(iulog,*)'PIO: DIM1NAME=',dim1name,' DIM2NAME=',dim2name
 
    allocate(Tmp3D(pcols,pver,begchunk:endchunk))
    allocate(Tmp2D(pcols,begchunk:endchunk))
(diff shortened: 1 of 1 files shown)


src/physics/cam/nudging.F90
```diff
@@ -1429,154 +1433,57 @@ subroutine nudging_update_analyses_se(anal_file)
...
-     istat=nf90_get_var(ncid,varid,Lon_anal)
-     if(istat.ne.NF90_NOERR) then
-       write(iulog,*) nf90_strerror(istat)
-       call endrun ('UPDATE_ANALYSES_SE')
-     endif
+   ! Open the file and get the fileID.
+   !-------------------------------------
+   call cam_pio_openfile(fileID,trim(anal_file),0)
+   call pio_seterrorhandling(fileID,PIO_BCAST_ERROR   )
+   call pio_seterrorhandling(fileID,PIO_INTERNAL_ERROR)
```
brian-eaton, line 1440: Why are the error handling calls back to back?  Internal error is the default, and normally you'd change that to broadcast error while you do checks that may fail, and later switch back to the default.  This code doesn't seem to be doing anything.

patcal: I used the code in metdata.F90 as a guide for my modifications. There were originally a couple of pio_inq_var() calls between them in the section of code I was referencing. The pio_inq calls were not relevant for my usage, but because I am not familiar with the PIO error handling functionality, I thought it best to include both of the pio_seterrorhandling() calls. Since they do nothing, they can certainly be removed.

gold2718:
Murphy ensures that whenever you select some code as a guide, you will select the worst possible example :-)
Also, the interface has changed so the best approach is to include these at the beginning of your routine:
```
   integer err_handling
...
       call pio_seterrorhandling(File, PIO_BCAST_ERROR, oldmethod=err_handling)
```
and this at the end of your routine.
```
    ! Back to old error handling
    call pio_seterrorhandling(File, err_handling)
```
This will reset the error handling back to whatever it was which is important if your routine is called from a region that has already changed the default PIO error handling method.

src/physics/cam/nudging.F90
```diff
@@ -1429,154 +1433,57 @@ subroutine nudging_update_analyses_se(anal_file)
...
-     if(istat.ne.NF90_NOERR) then
-       write(iulog,*) nf90_strerror(istat)
-       call endrun ('UPDATE_ANALYSES_SE')
-     endif
-   endif ! (masterproc) then
-   call scatter_field_to_chunk(1,1,1,Nudge_ncol,PSanal,           &
-                               Nobs_PS(1,begchunk,Nudge_ObsInd(1)))
+   ! Read in, U,V,T,Q, and PS  
+   !----------------------------------
+   call infld('U',fileID,dim1name,'lev',dim2name,     &
```
nusbaume, line 1452: It looks like there are no checks here for whether or not these variables are actually present on the file.  I believe you need to add something like this after the `infld` call:

```
if (VARflag) then 
   Nobs_U(:,:,begchunk:endchunk,Nudge_ObsInd(1)) = Tmp3D(:,:,begchunk:endchunk)
else
   call endrun('Variable "U" is missing from '//trim(anal_file))
end if
```
Something similar should be added to the rest of the `infld` calls as well (I think).

nusbaume (changes requested):
Mostly looks good to me, beyond the `seterrorhandling` issue found previously.  However, I do have a few concerns which I would like to see addressed, if possible.  Thanks!

src/physics/cam/nudging.F90
```diff
@@ -1429,154 +1433,57 @@ subroutine nudging_update_analyses_se(anal_file)
...
-       call endrun ('UPDATE_ANALYSES_SE')
-     endif
-     istat=nf90_get_var(ncid,varid,Lat_anal)
-     if(istat.ne.NF90_NOERR) then
-       write(iulog,*) nf90_strerror(istat)
-       call endrun ('UPDATE_ANALYSES_SE')
-     endif
+   grid_id = cam_grid_id('physgrid')
+   call cam_grid_get_dim_names(grid_id,dim1name,dim2name)
+!   if(masterproc) write(iulog,*)'PIO: DIM1NAME=',dim1name,' DIM2NAME=',dim2name
```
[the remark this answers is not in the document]
gold2718: Another technique is to define a `debug` private logical symbol at module scope which you can set to `.true.` when you want these messages.
```suggestion
   if(masterproc .and. debug) then
      write(iulog,*)'PIO: DIM1NAME=',dim1name,' DIM2NAME=',dim2name
   end if
```

gold2718 (changes requested):
Mostly supporting previous change requests (just added a suggestion).

fvitt:
What issue does this PR fix?  Does this address scalability issues?

src/physics/cam/nudging.F90
```diff
@@ -1624,202 +1551,83 @@ subroutine nudging_update_analyses_eul(anal_file)
      end do
      Nudge_ObsInd(1)=Nindex
      inquire(FILE=trim(anal_file),EXIST=Nudge_File_Present(Nudge_ObsInd(1)))
+     write(iulog,*)'NUDGING: Nudge_ObsInd=',Nudge_ObsInd
+     write(iulog,*)'NUDGING: Nudge_File_Present=',Nudge_File_Present
    endif
 #ifdef SPMD
    call mpibcast(Nudge_File_Present, Nudge_NumObs, mpilog, 0, mpicom)
    call mpibcast(Nudge_ObsInd      , Nudge_NumObs, mpiint, 0, mpicom)
```
fvitt, line 1559: Should mpi_bcast (without ifdef SPMD) be used rather than mpibcast?

gold2718:
> What issue does this PR fix? Does this address scalability issues?

#237, thanks for noticing, I updated the description.

fvitt:
Are we expecting this regression test to be restored.   I am hoping someone checks that it is bit-for-bit compared to cam6_3_006.  

FAIL ERP_Ln9.ne30_ne30_mg17.FCnudged.cheyenne_intel.cam-outfrq9s RUN time=39
Nudging code is using global interfaces not supported for weak scaling.
This error should be fixed with PR #241

gold2718:
> Are we expecting this regression test to be restored

Yes, I will make sure of that.

nusbaume (approved):
Looks good to me!  Just one typo in all the new `endrun` calls.

gold2718:
From around line 772:

> `       call endrun('NUDGING IS CURRENTLY ONLY CONFIGURED FOR CAM-SE, FV, or EUL')`

Is this still true? Should it work with MPAS and/or FV3?

patcal:
The module originally needed different interfaces to read in  the data for 
each dycore. FV3 and MPAS are new so there is no interface set up for them, 
but if you notice, with the change to infield calls all of the nudging_update_analyses_*()
subroutines are now identical. So I think that adding dycore_is() cases for the 2 
new dycores to the if-elseif block and hang them call nudging_update_analyses_se() 
will allow for FV3 and MPAS.  I left the degeneracy for the time being in case any
dycore-dependent differences showed up during testing and needed to be added 
back in. An alternative would be to remove the if-block and the redundant routines.

nusbaume (approved):

gold2718:
The changes from this PR were merged to cam_development as part of cam6_3_008 (#310).

PR #243, opened 2020-10-07
Fix bug in issue-closing script

### LINKED ISSUE #242: Issue closing github action fails with tag creation that references PR with open issue
If a user merges a PR, and then pushes a tag with the same commit message as the PR merge commit itself, then the issue closing script will run through the already merged PR's Github text, and will attempt to close any open issues that are present.  However, given that the PR has already been run through by the script once, open issues will likely have had their project cards moved to the `closed issues` column, which will then cause the issue-closing script to fail on a second pass, as it is unable to find them.

This can be fixed by simply having the script check the `closed issues` project column for the relevant issue numbers.  In the meantime, if you create a tag, I would recommend not including the `#` symbol when referencing the PR number in the tag's commit message.  Otherwise you'll get an email from Github complaining about a failed action (even though everything is actually fine).

nusbaume:
Fix bug in issue-closing script so that creating a tag with the same commit message as the original PR merge commit doesn't result in a failure.  Also improve the error message when a non-existent issue number is referenced.

Fixes #242

## Tests performed:

Tests were run on a Github test repo to ensure that the action runs correctly when multiple pushes occur with the same commit message and PR reference (similar to what a tag commit would do).

### DIFF
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -470,10 +470,10 @@ def _main_prog():
             print("Issue #{} has been closed.".format(issue_num))
         else:
             #Extract card id from id dictionary:
-            try:
+            if issue_num in proj_issue_card_ids:
                 card_id = proj_issue_card_ids[issue_num]
-            except KeyError:
-                #If there is a key error, then it means the issue
+            else:
+                #If issue isn't in dictionary, then it means the issue
                 #number was never found in the "To do" column, which
                 #likely means the user either referenced the wrong
                 #issue number, or the issue was never assigned to the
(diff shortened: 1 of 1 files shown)


.github/scripts/branch_pr_issue_closer.py
```diff
@@ -447,7 +470,20 @@ def _main_prog():
             print("Issue #{} has been closed.".format(issue_num))
         else:
             #Extract card id from id dictionary:
-            card_id = proj_issue_card_ids[issue_num]
+            try:
+                card_id = proj_issue_card_ids[issue_num]
+            except KeyError:
+                #If there is a key error, then it means the issue
```
gold2718, line 476: Why use `try` / `except` here instead of a test? E.g., 
```suggestion
            if issue_num in proj_issue_card_ids:
                card_id = proj_issue_card_ids[issue_num]
            else:
                #If there is a key error, then it means the issue
```

nusbaume: Good call.  Fixed.

gold2718 (approved):
Looks good, thanks.

PR #245, opened 2020-10-07
Fix bug in issue-closing script for CESM2.1 branch

### LINKED ISSUE #242: Issue closing github action fails with tag creation that references PR with open issue

nusbaume:
This is the same as PR #243 , except it is for the `cam_cesm2_1_rel` branch.

Fixes #242

### DIFF
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -470,10 +470,10 @@ def _main_prog():
             print("Issue #{} has been closed.".format(issue_num))
         else:
             #Extract card id from id dictionary:
-            try:
+            if issue_num in proj_issue_card_ids:
                 card_id = proj_issue_card_ids[issue_num]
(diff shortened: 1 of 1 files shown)


gold2718 (approved):
Looks good, thanks.

PR #272, opened 2020-11-13
contrail parameterization

chihchen24:
We have added a contrail parameterization to CAM which will not be triggered if aviation emission files are not included in the user_nl_cam and thus will not, by default, alter the answer of the simulation.

We will also need disk space to store the aviation emission files as input files for the parameterization. 

Thanks!
@andrewgettelman

### DIFF
--- a/src/chemistry/utils/aircraft_emit.F90
+++ b/src/chemistry/utils/aircraft_emit.F90
@@ -40,16 +40,16 @@ module aircraft_emit
 
   type(forcing_air),allocatable :: forcings_air(:)
 
-  integer, parameter :: N_AERO = 10 
-  character(len=11)    :: aero_names(N_AERO) = (/'ac_HC      ','ac_NOX     ','ac_PMNV    ',&
+  integer, parameter :: N_AERO = 13
+  character(len=13)    :: aero_names(N_AERO) = (/'ac_SLANT_DIST','ac_TRACK_DIST','ac_HC      ','ac_NOX     ','ac_PMNV    ',&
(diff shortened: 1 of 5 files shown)


andrewgettelman:
This code has been lying around for a while, and written up in several papers. We have updated the code to CAM6, and are getting requests from collaborators and CESM users to get the code. Given that it does not change answers if not used, we would like to get it checked in, and put the emission files needed into inputdata

gold2718:
We need more information on this including which CAM versions are targets.
Please open an issue that spells out the details of the changes, data needs, and CAM targets.
See step 2 of [How to submit code changes to be included in ESCOMP/CAM](https://github.com/ESCOMP/CAM/wiki/CAM-Development-Workflow-in-GitHub#how-to-submit-code-changes-to-be-included-in-escompcam).

andrewgettelman:
Step 2 just says 'open a pull request' 
Sorry for speaking a different language: this should go onto CAM_development (what used to be the cam trunk).
This is an open issue.  What specifically do you need for Data needs? Jack can detail the input files need. 

Thanks!

Andrew

gold2718:
@andrewgettelman, that is step 4. Step 2 talks about opening an issue where we describe the needs and proposed solutions. It is also how we prioritize incoming work.
This PR is not set up for cam_development. To effect this, this PR will need to be closed and a correct one opened (another reason to have an issue that survives situations like incorrect PRs).
Step 4 does talk about how to open a correct PR (i.e., one that will merge correctly).

andrewgettelman:
Jack: you need to put the issue in the CAM_development. We don't want it on the release branch, we want it on the CAM_development branch so it moves forward with the model. So Steve says you have to close it and open a new issue with the CAM_development branch and do the commit in there. Does that make sense?

Issue #274, opened 2020-11-17
contrail parameterization


chihchen24:
A contrail parameterization was implemented and we will like it to get into the CAM development tag.
The parameterization is triggered only if user_nl_cam includes aviation emission files needed for the parameterization.
Otherwise, the parameterization will not be activated and thus will not change the answer of the simulation.

In addition to the parameterization, we will need to request disk space in inputdata directory to store the aviation emission files for the parameterization.

@andrewgettelman @cacraigucar

nusbaume:
Closed manually as we forgot to add this to the CAM development project.

PR #275, opened 2020-11-17
Cam cesm2 1 rel cchen

chihchen24:
#274 

This PR is for the contrail parameterization.

@andrewgettelman @cacraigucar


cacraigucar:
@chihchen24 This PR won't work either.  You can look at the number of files changed and see that it is referencing maybe every file in CAM!  The problem is that your branch in your fork (where your changes are located) is not the proper place to start this PR in order to merge it into cam_development.

PR #276, opened 2020-11-17
cam6_3_005: Enable "-chem none" configure option with cam6 physics

### LINKED ISSUE #261: Get -chem none option (prescribed bulk aerosols) functional with CAM6
A bulletin board user has discovered that the -chem none option does not work with CAM6:
https://bb.cgd.ucar.edu/cesm/threads/how-to-close-atmospheric-chemistry-in-fhist_bgc-case.5678/

This was reported by @cecilehannay. 

The error indicates the appearance of pointing to some mam files (_m3 seems to be a mode), which might be the unused 'prescribed modal aerosols' which we DO NOT want to use.

A discussion thread developed on this, including @brian-eaton, @nusbaume and @fvitt 

It would be nice to figure out how to get the prescribed bulk aerosols to work with -chem none for CAM6 as they do for CAM5. This might not be that bad:

As reported by @cacraigucar We have two CAM regression tests for -chem none which are run with every CAM tag we make.  One is with ne5 aquaplanet CAM5 and the other is f10 aquaplanet CAM5.   They only run for 9 timesteps, but they run.  

It would be great if CAM6 could get to the same level.

fvitt:
Do not set use_hetfrz_classnuc if chemistry does not include MAM species.

Added aux_cam test which uses aquaplanet cam6 physics and -chem none.

closes #261

### DIFF
--- a/cime_config/testdefs/testlist_cam.xml
+++ b/cime_config/testdefs/testlist_cam.xml
@@ -785,6 +785,15 @@
       <option name="wallclock">00:30:00</option>
     </options>
   </test>
+  <test compset="QPC6" grid="f10_f10_mg37" name="SMS_D_Ln9" testmods="cam/outfrq3s_ba">
+    <machines>
+      <machine name="izumi" compiler="nag" category="aux_cam"/>
+    </machines>
+    <options>
+      <option name="comment" >cam6 physics with prescribed bulk aerosols</option>
+      <option name="wallclock">00:30:00</option>
+    </options>
+  </test>
 <!--              003                        -->
   <test compset="QPSCAMC5" grid="T42_T42_mg17" name="SMS_D_Ln7" testmods="cam/scmarm">
     <machines>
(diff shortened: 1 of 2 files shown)


cime_config/testdefs/testlist_cam.xml
```diff
@@ -445,6 +445,14 @@
...
     </options>
   </test>
+  <test compset="QPC6" grid="f10_f10_mg37" name="SMS_D_Ln9" testmods="cam/outfrq3s_ba">
+    <machines>
+      <machine name="izumi" compiler="nag" category="aux_cam"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:30:00</option>
+    </options>
+  </test>
```
gold2718, line 455: Why is this test here instead of after the QPC5 version of this test (see line 677)?
Also, why does `outfrq3s_ba` mean no chemistry?

gold2718 (approved):
Looks okay to me (assuming you add a ChangeLog entry), just one question.

nusbaume (approved):

cacraigucar (approved):

cime_config/testdefs/testlist_cam.xml
brian-eaton: "ba" --> bulk aerosols which is what you get for '-chem none'.

brian-eaton (approved):

PR #285, opened 2020-12-02
Remove support for CISM1

billsacks:
As of cism2_1_74, CISM1 is no longer supported, so we don't need to have a PE layout section for it.

I haven't done any testing of this simple change. I am assuming you'll want to just combine it with some other changes at some point.


billsacks:
I also see references to CISM1 in commented out lines in cime_config/config_compsets.xml - related to old refcases. I imagine those are old things that should just be removed, but I didn't want to do that removal myself in case those commented-out lines are serving some purpose.

PR #294, opened 2020-12-10
Update External_CAM.cfg

KapilS25:
https://github.com/CFMIP/COSPv2.0/tags doesnt exist  hence replacing with https://github.com/CFMIP/COSPv2.0/tree


cacraigucar:
@KapilS25 Please see the discussion in #295 and let us know if we need to keep this PR open or if we can close it without merging.  If you believe it still needs to be changed, please indicate your reasoning in the issue.

PR #296, opened 2020-12-14
Fix cam_snapshot not working with CLUBB and rename output fields to contain structure

### LINKED ISSUE #286: cam_snapshot is not working with clubb_tend_cam
There is an error with cam_snapshot when clubb_tend_cam is set.  The error occurs in build-namelist.

cacraigucar:
When cam_snapshot was called using CLUBB, build-namelist was not returning the correct information.

Also, the structure name has been prepended to the variables, and "_snapshot" has been removed.

Closes #286

### DIFF
--- a/src/control/cam_snapshot.F90
+++ b/src/control/cam_snapshot.F90
@@ -147,19 +147,19 @@ use time_manager,   only: is_first_step
 
    lchnk = state%lchnk
 
-   call outfld('flx_heat_snapshot', flx_heat, pcols, lchnk)
-   call outfld('cmfmc_snapshot', cmfmc, pcols, lchnk)
-   call outfld('cmfcme_snapshot', cmfcme, pcols, lchnk)
-   call outfld('pflx_snapshot', pflx, pcols, lchnk)
-   call outfld('zdu_snapshot', zdu, pcols, lchnk)
-   call outfld('rliq_snapshot', rliq, pcols, lchnk)
-   call outfld('rice_snapshot', rice, pcols, lchnk)
-   call outfld('dlf_snapshot', dlf, pcols, lchnk)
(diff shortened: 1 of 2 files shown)


cacraigucar:
Tested by running a case with and without the mods.  I ended up with the same number of fields in both the with and without mods files.  Also, made sure there were no fields with _snapshot left in the name.

src/control/cam_snapshot.F90
```diff
@@ -1771,11 +1771,7 @@ subroutine fill_pbuf_info(pbuf_info, pbuf, const_cname)
 !          pbufname,                 'unset',        &
 !          pbufname,                 'unset',        &
 !          pbuf_names(i),            'unset',        &
-!          rad_name1,                'unset',        &
-!          rad_name2,                'unset',        &
-!          rad_name3,                'unset',        &
-!          rad_name,                 'unset',        &
-!          sad_name,                 'cm2/cm3',      &
+!          rad_name1,                'unset',        & !          rad_name2,                'unset',        & !          rad_name3,                'unset',        & !          rad_name,                 'unset',        & !          sad_name,                 'cm2/cm3',      &
```
nusbaume, line 1774: Why was this change made?  It looks like it might have been done on accident?

nusbaume (approved):
Looks good!  Just one concern with a comment modification.

src/control/cam_snapshot.F90
cacraigucar: Thought I undid all the changes made with an accidental caps lock.  ("j" in vi to move becomes "J" to join lines)  Obviously I missed some!

cacraigucar:
Ran one more test case (Cecile's setup) and am getting unexpected errors.  Please wait on final review until this is sorted out tomorrow.

PR #330, opened 2021-02-16
part of cam6_3_010: Solutions to issue of configure option "-chem none" with cam6 physics

### LINKED ISSUE #169: refactor WACCM settings for qbo_use_forcing
Currently all the WACCM use case files set qbo_use_forcing.  This overrides the ability to make the setting .false. for all SE grids in the namelist_defaults_cam.xml file.  This should be refactored so that the SE setting doesn't need to appear in every WACCM use case file.

### LINKED ISSUE #261: Get -chem none option (prescribed bulk aerosols) functional with CAM6

### LINKED ISSUE #334: Remove reprosum namelist settings from FSD test
During regression testing for cesm2_3_alpha02b, it was discovered that the FSD test was not handling the reprosum namelist settings correctly.  These settings will be removed for the FSD tests.

fvitt:
By default, turn off volcanic forcings when configure option "-chem none" is used with cam6 physics.  

Change aerosol names in the physics property files of prescribed stratospheric aerosols so the the 3 modes have unique names and there are no namespace issues with history fields added in ndrop_bam if the user chooses to turn on the volcanic forcings.

The default configuration with "-chem none" is to have no volcanic forcings.  One can turn on the cam6 volcanic forcings by setting prescribed_strataero_feedback=.true.

Added an aux_cam test since it seems that there is a desire to support this configuration.

Refactor how WACCM physics QBO nudging is turned on.

Remove reprosum settings from FSD prealpha test.

Closes #169
Closes #261
Closes #334


andrewgettelman:
I think this solution sounds like it will meet our needs, and will be happy to get back to a working -chem none configuration.  Thanks Francis!

andrewgettelman (approved):
I think the proposed solution will work scientifically. Happy to have this functionality back. Thanks Francis.

nusbaume (approved):

cacraigucar (approved):

gold2718 (approved):
Thanks for bringing this up.

PR #332, opened 2021-02-18
Add Readme auto-update Github Action

### LINKED ISSUE #253: Need Github Action to update listed tag in README
Currently the `README.md` file on CAM's master branch needs to be updated manually whenever a new `cam_development` branch tag is created.  This can result in the README file being out-of-sync with the actual repo.  

To fix this, a Github action should be created that automatically updates the README file whenever a new `cam_development` tag is pushed.

nusbaume:
Fixes #253 

One question that still needs to be determined is if we want the ability to push the tag regardless of what  branch one has checked out locally, or if we want to require the user to checkout `cam_development` locally before pushing the tag.  If we want this auto-update to be independent of the local branch then I will need to push these files to the other CAM branches as well (please note that the README file will still only be updated for development tags, just that the workflow will trigger if someone pushes the development tag while sitting in, say, the default branch).

## Tests run:

Ran multiple tests in the CAM test repository (https://github.com/NCAR/repoForTestingCAM).  The script behaves as expected, at least as far I could tell.

### DIFF
--- a/.github/workflows/branch_push_workflow.yml
+++ b/.github/workflows/branch_push_workflow.yml
@@ -31,7 +31,7 @@ jobs:
       if: github.repository == 'ESCOMP/CAM' # Only run on main repo
       run: |
         python -m pip install --upgrade pip  # Install latest version of PIP
-        pip install PyGithub                 # Install PyGithub pythong package
+        pip install PyGithub                 # Install PyGithub python package
     # Run CAM issue-closing script
     - name: Run python Github issue-closing script
       if: github.repository == 'ESCOMP/CAM' # Only run on main repo
--- a/.github/workflows/tag_push_workflow.yml
+++ b/.github/workflows/tag_push_workflow.yml
@@ -25,7 +25,7 @@ jobs:
       if: github.repository == 'ESCOMP/CAM' # Only run on main repo
       run: |
         python -m pip install --upgrade pip  # Install latest version of PIP
-        pip install PyGithub                 # Install PyGithub pythong package
+        pip install PyGithub                 # Install PyGithub python package
     # Run CAM README tag update script
     - name: Run python README tag-updating script
       if: github.repository == 'ESCOMP/CAM' # Only run on main repo
--- a/.github/workflows/tag_push_workflow.yml
+++ b/.github/workflows/tag_push_workflow.yml
@@ -1,21 +1,15 @@
-name: Pushed branch commit workflow
+name: Pushed tag workflow
 
 on:
-  #For some reason GitHub does not allow secrets
-  #when a PR from a fork is closed.  Thus instead
-  #the workflow must activate whenever a commit
-  #is pushed to the repo.  Theoretically this should
-  #behave the same way as if triggered from a pull
-  #request, just as long as no user ever pushes
-  #directly to the repo.
   push:
-    branches:
-      - '**'  # Trigger workflow on push to any branch or branch hierarchy.
+    tags:
+      - '**'  # Trigger workflow on push of any tag or tag hierarchy.
+
 
 jobs:
-  #This job is designed to close any issues or pull requests specified
-  #in the body of a pull request merged into a non-default branch.
-  issue_closer:
+  #This job will update the README file on the default branch
+  #if the push created a new development branch tag.
+  update_readme_dev_tag:
     runs-on: ubuntu-latest
     steps:
     # Acquire github action routines
@@ -32,10 +26,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip  # Install latest version of PIP
         pip install PyGithub                 # Install PyGithub pythong package
-    # Run CAM issue-closing script
-    - name: Run python Github issue-closing script
+    # Run CAM README tag update script
+    - name: Run python README tag-updating script
       if: github.repository == 'ESCOMP/CAM' # Only run on main repo
       env:
         ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
-      run: .github/scripts/branch_pr_issue_closer.py --access_token $ACCESS_TOKEN --trigger_sha $GITHUB_SHA
+      run: .github/scripts/readme_tag_update.py --access_token $ACCESS_TOKEN --trigger_sha $GITHUB_SHA
 
--- a/.github/scripts/readme_tag_update.py
+++ b/.github/scripts/readme_tag_update.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env python
+
+"""
+Script name:  readme_tag_update.py
+
+Goal:  To determine if a recent push to the "development" branch
+       was for a tag, and if so, to update the README.md file on
+       the default branch to display the new development tag.
+
+
+Written by:  Jesse Nusbaumer <nusbaume@ucar.edu> - February, 2020
+"""
+
+#+++++++++++++++++++++
+#Import needed modules
+#+++++++++++++++++++++
+
+import re
+import sys
+import argparse
+
+from github import Github
+
+#################
+#HELPER FUNCTIONS
+#################
+
+#++++++++++++++++++++++++++++++
+#Input Argument parser function
+#++++++++++++++++++++++++++++++
+
+def parse_arguments():
+
+    """
+    Parses command-line input arguments using the argparse
+    python module and outputs the final argument object.
+    """
+
+    #Create parser object:
+    parser = argparse.ArgumentParser(description='Close issues and pull requests specified in merged pull request.')
+
+    #Add input arguments to be parsed:
+    parser.add_argument('--access_token', metavar='<GITHUB_TOKEN>', action='store', type=str,
+                        help="access token used to access GitHub API")
+
+    parser.add_argument('--trigger_sha', metavar='<GITHUB SHA>', action='store', type=str,
+                        help="Commit SHA that triggered the workflow")
+
+    #Parse Argument inputs
+    args = parser.parse_args()
+    return args
+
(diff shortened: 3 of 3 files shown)


cacraigucar:
Two comments/question:

1. Shouldn't the script automatically update the tag for whichever branch is being pushed?
2. I typically push the changes and then make my tag after it is pushed.  Will this break your proposed enhancement?

nusbaume:
@cacraigucar In response to your questions:

**1.  Shouldn't the script automatically update the tag for whichever branch is being pushed?**

I was told that we will only want to auto-update the development tags, as the release branch tags are associated with a particular CESM tag, and so should be updated manually?  Of course if we decide to update them automatically then I can adjust the script to do so.  The other issue is that it is possible to push a tag from any branch regardless of whether the commit that is being tagged is actually on that branch (it just has to be in the repo somewhere).  

If this script/Action was pushed to every branch then this scenario is managed properly.  If it is just left in `cam_development` however then one will have to locally checkout `cam_development`  before pushing the tag (which is fine, it just means there is another step involved for the user).

**2.  I typically push the changes and then make my tag after it is pushed. Will this break your proposed enhancement?**

That is exactly the workflow this script is expecting.  Basically this script only checks if a tag is pushed, and doesn't really care when the tagged commit is pushed.  

In other words, I designed this script with the [SE workflow](https://github.com/NCAR/CAM/wiki/CAM-SE-Workflows#primary-assigned-ses-duties-for-when-code-from-scientists-are-being-brought-into-cam) in mind.  So, I imagine if you did something really weird and outside the workflow norms then you could probably break the script, but I am hoping we all just won't try and do something like that.

Anyways, I hope that helps, and if you still have any questions or concerns please let me know!

nusbaume:
@goldy I finally got around to adding the remaining features you requested.  Specifically, the script that is triggered whenever someone pushes a tag (`tag_check_readme_update.py`) will now check the `ChangeLog` to make sure that the first tag listed is the one that was pushed.  It also does some basic `ChangeLog` sanity checks, which I imagine might be beneficial if there is ever a situation where the file is mistakenly deleted or corrupted (especially given that it is frequently modified after code review).

Along with the extra checks, the script will now also send an email from the "cam.engineering.works" account that notifies everyone whenever a problem has occurred that would be worth investigating.  I decided to only include the core CAM SEs on the email list instead of the entire "CCM Core group" mailing list, as I imagine people wouldn't want their mistakes broadcasted to everyone.  However, I can certainly change this if preferred.

Thanks!

nusbaume:
@peverwhee I decided to add you as well given that you also know python, but of course feel free to remove yourself if this isn't your cup of tea.  I just figured the more eyes the better.  Thanks!

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+
+    #Convert changelog blob to unicode string:
+    changelog = base64.b64decode(changelog_git_blob.content).decode("UTF-8")
+
+    #+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+    #Search for tag in ChangeLog, to ensure log was properly updated
+    #+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+    #Make ChangeLog all lowercase, so that searching is case independent:
+    changelog = changelog.lower()
```
gold2718, line 299: Do we need to make another copy of a > 5MB string here (we already copied it once to do the decode)? The tag better be in the first 100 characters, maybe live large and go for 256?
```suggestion
    changelog = changelog[0:256].lower()
```

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+    if not dev_tag_match:
+        endmsg = "No text matches expected development tag pattern in README.md file, "
+        endmsg += "so no further action will be taken."
+        end_script(endmsg)
+
+    #Extract start and end indices of matched tag text:
+    dev_tag_idx = dev_tag_match.span()
+
+    #Add new tag to README text:
+    new_content = readme_text[:dev_tag_idx[0]] + tag_name + readme_text[dev_tag_idx[1]:]
```
gold2718, line 359: Why not just use `replace`?
```suggestion
    #Add new tag to README text:
    new_content = readme_text.replace dev_tag_match.group(0), tag_name)
```

.github/workflows/branch_push_workflow.yml
```diff
@@ -1,38 +1,39 @@
...
-#directly to the repo.
-on: push
+on:
+  #For some reason GitHub does not allow secrets
+  #when a PR from a fork is closed.  Thus instead
+  #the workflow must activate whenever a commit
+  #is pushed to the repo.  Theoretically this should
+  #behave the same way as if triggered from a pull
+  #request, just as long as no user ever pushes
+  #directly to the repo.
```
gold2718, line 10: Is there a PyGithub command to send that user an electrical shock or maybe shame them on Instatwokbook?

.github/workflows/tag_push_workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+    if: github.repository == 'ESCOMP/CAM'
+    runs-on: ubuntu-latest
+    steps:
+    # Acquire github action routines
+    - uses: actions/checkout@v2
+    # Acquire specific version of python
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.6' # Semantic version range syntax or exact version of a Python version
```
gold2718, line 22: It's been a while since you opened this PR. Time to update this version?

.github/workflows/branch_push_workflow.yml
```diff
@@ -1,38 +1,39 @@
...
   #in the body of a pull request merged into a non-default branch.
   issue_closer:
+    if: github.repository == 'ESCOMP/CAM' # Only run on main repo
     runs-on: ubuntu-latest
     steps:
-    # acquire github action routines
+    # Acquire github action routines
     - uses: actions/checkout@v2
-    # acquire specific version of python
+    # Acquire specific version of python
```
gold2718, line 24: It's been a while since you opened this PR. Time to update the python version?

gold2718 (changes requested):
A few suggestions.

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+        for receiver in receiver_list:
+            #Send email
+            server.sendmail("cam.engineering.works@gmail.com", receiver,
+                            msg_email)
+
+#++++++++++++++++++++++++++++++++
+#Script message and exit function
+#++++++++++++++++++++++++++++++++
+
+def end_script(msg):
```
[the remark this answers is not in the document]
nusbaume: Done.

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+
+    #+++++++++++++++++++++++++++++++
+    #Search for associated PR number
+    #+++++++++++++++++++++++++++++++
+
+    #Extract tag commit message:
+    commit_message = tag_commit.commit.message
+
+    #Compile Pull Request merge text expression:
+    pr_merge_pattern = re.compile(r'Merge pull request ')
```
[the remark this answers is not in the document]
nusbaume: Done (along with suggested simplification below).

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+        end_script(endmsg)
+    else:
+        print("Script found tag name of '{}'".format(tag_name))
+
+    #+++++++++++++++++++++++++++++++
+    #Search for associated PR number
+    #+++++++++++++++++++++++++++++++
+
+    #Extract tag commit message:
+    commit_message = tag_commit.commit.message
```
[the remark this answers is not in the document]
nusbaume: Agreed, it can't hurt!  `strip` has been added.

.github/scripts/tag_check_readme_update.py
```diff
@@ -0,0 +1,385 @@
...
+        print("Merged PR associated with tag: {}".format(first_word))
+
+        try:
+            #Try assuming the word is just a number:
+            pr_num = int(first_word[1:]) #ignore "#" symbol
+        except ValueError:
+            #If the conversion fails, then this is likely not a real PR merge, so end the script:
+            endmsg = "No Pull Request number was found in the tagged commit message.\n"
+            endmsg += "This is a non-normal tag commit, and so would be worth double-checking."
+            end_script_fail(cam_pword, endmsg)
```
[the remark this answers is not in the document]
nusbaume: Thanks for this suggestion!  The  code has been cleaned-up now.

Also, I changed the print message wording slightly to make it clear that it is the PR number that is being printed, and not the tag name.

.github/scripts/tag_check_readme_update.py
nusbaume: Good call!  Applied suggested change.

.github/workflows/branch_push_workflow.yml
nusbaume: I was going to implement that, but _sadly_ UCAR Legal said it wasn't allowed.

.github/workflows/branch_push_workflow.yml
nusbaume: Updated to python 3.9

nusbaume:
@goldy  Actually, It looks like gmail is again blocking me, so I'll need to figure out an alternative to get the email feature working.  Since I am not 100% sure if an easy solution exists, I will go ahead and close this PR, and then re-open it if I find an adequate solution.

PR #333, opened 2021-02-18
Update 'nhtfrq' namelist definition to remove incorrect text.

### LINKED ISSUE #265: Need to update "nhtfrq" namelist definition
Currently the definition for the namelist variable `nhtfrq` states that only the first file can contain monthly averages.  However, this is not actually the case, and appears to be causing some confusion among certain users.  Thus the namelist variable definition should be updated to remove this "first file" statement.

nusbaume:
Fixes #265 

@cacraigucar If you could incorporate this change into the upcoming "miscellaneous" tag PR that would be super-helpful.  Thanks!


nusbaume:
I have also added the relevant `ChangeLog` additions in case that helps with the Misc. PR documentation as well.

cacraigucar (approved):

PR #348, opened 2021-03-29
part of cam6_3_016: support for nuopc threading

jedwards4b:
This is a port for CMEPS threading in CAM.  The change in physics_types.F90 is due to a stack size issue in the call to flush.  It could also be solved via an OMP master region, but I think that that call in the code is really old and crusty and no longer required.  
Fixes issue #349


cacraigucar (approved):

gold2718:
> I think that that call in the code is really old and crusty and no longer required.

So if you don't like it, it is "old and crusty"? ;)

Seriously, this module is going away in the new infrastructure.

nusbaume (approved):

gold2718 (changes requested):
This PR still needs a ChangeLog entry and an issue describing the problem this is addressing.

Issue #278, opened 2020-11-18
add clubb time-step to the atm.log


adamrher:
The clubb's time-step is not currently outputted anywhere. As clubb is probably the 2nd most expensive piece of the CAM code, it is important to keep track of its time-step as it related to model performance. 

At present, the clubb timestep, "dtime" is computed in clubb_intr.F90. It internally computes dtime from dt_macmic = dtphys/cld_macmic_num_steps. It solves for the smallest integer "nadv" one needs to divide dt_macmic by such that dtime <= clubb_timestep namelist var. So there is no guarantee that dtime = clubb_timestep namelist var.

adamrher:
Working on a PR to clean-up some clubb issues and am requesting help from the SE's on how best to output the clubb time-step to atm.log. Looking through the atm.log, It looks like ancillary info is written out in the readnl or init subroutines. How can I pull the physics time-step into the init or readnl subroutines in clubb_intr?

gold2718:
Isn't this a fun game? I think one confusion is a comment in physpkg:
```
    real(r8), intent(in) :: ztodt            ! physics time step unless nstep=0
```

I think this comment is incorrect. The physics timestep is set in `stepon_run1` which means that it is not available at namelist read or init time. However, `phys_run1` is always called after `stepon_run1` so that is the place to add the report. Something like:
```
if ((is_first_step() .or. is_first_restart_step()) .and. masterproc) then
  write(iulog, *) 'Yeehaw, we know the time step size! ', cld_macmic_ztodt
end if
```
added to the top of `clubb_tend_cam` or maybe even in `tphysbc` after `cld_macmic_ztodt` is calculated.

adamrher:
purrfect. FYI I'm running some 5 year simulations to test mods in this issue and https://github.com/ESCOMP/CAM/issues/336. If all goes well I plan on opening a PR to close these two issues this evening/tmrw. OK by you?

adamrher:
"physics time step unless nstep=0" is confusing so might as well just remove the "unless nstep=0" part?

gold2718:
> OK by you?

Opening an issue, easy, getting the SEs to do anything about it, priceless ;)

> just remove the "unless nstep=0" part?

Yes please. As far as I can tell, it is not true. I think the history has to do with the fact that `tphysbc` is called at model initialization time. Perhaps this comment was true when that was first set up but I think it has been fixed long ago (if anyone invents a regression test for comments, they should get the Turing Award).

[...]

Issue #351, opened 2021-04-02
CLUBB supersaturation issue


adamrher:
CLUBB is supposed to remove super-saturation from the solution, but it is not very good at it (see attached pdf of RH computed just after exiting the first macmic iteration of CLUBB. Green 'cntl' is out of the box cam6).

The cause of the supersaturation in CLUBB is primarily due to the use of the trapezoidal rule to produce a layer mean cloud liquid. That is, CLUBB does its pdf closure/saturation adjustment to provide a cloud liquid at the mid-points. It then calls the pdf closure a second time to get a cloud liquid at the interfaces. The midpoint and 2 surrounding interface ql's are used by the trapezoidal rule to construct a piecewise linear subgrid reconstruction, which is integrated over the layer to give a layer-mean cloud liquid. This more often than not reduces the cloud liquid from the value produced from the pdf closure at the mid-points. When converting back to T from thetal using this new ql systematically lowers T from the value it would have had, had we used the mid-point ql from the pdf closure. The cooler T's result in a supersaturated state.

Taking the mid-point ql from the pdf closure as the layer mean ql solves this problem (light blue lines in pdf "notrap"), which is invoked by turning off the trapezoidal rule via the namelists:

clubb_l_trapezoidal_rule_zm = .false.
clubb_l_trapezoidal_rule_zt = .false.

It increases low level cloud liquid just about everywhere though ... but that is in a direction more in line with observations.

[AMWG diagnostics of run with the two namelists set to false vs. cntl](https://webext.cgd.ucar.edu/F2000climo/cam6_3_006_F2000climo_f09_f09_mg17_1800pes_210324_postipdf_6macmic_notrap_Nx5yrs/atm/cam6_3_006_F2000climo_f09_f09_mg17_1800pes_210324_postipdf_6macmic_notrap_Nx5yrs.2_10-cam6_3_006_F2000climo_f09_f09_mg17_1800pes_201210_postipdf-6macmic-Nx10yrs.2_10/).

![temp_pdf clubb rhsens](https://user-images.githubusercontent.com/29961476/113459932-42eb5180-93d4-11eb-9f09-a94b80f1759f.png)

vlarson:
Nice work, Adam.

I recommend trying the following flag settings, which are intended to go together:
```
l_vert_avg_closure  = .false.
l_trapezoidal_rule_zt = .false.
l_trapezoidal_rule_zm = .false.
l_call_pdf_closure_twice = .false.
```
[...]

PR #352, opened 2021-04-05
bugfixes/cleanup clubb_intr, output physics timesteps to log

adamrher:
Fixes https://github.com/ESCOMP/CAM/issues/278
Fixes https://github.com/ESCOMP/CAM/issues/336

To be added to this PR (before first code review):

Fixes https://github.com/ESCOMP/CAM/issues/496
Fixes https://github.com/ESCOMP/CAM/issues/497


andrewgettelman (approved):
I've looked at the diags. I concur this is not climate changing. Fine with me.

adamrher:
In tphysac and tphysbc, Is this true:

    real(r8), intent(in) :: ztodt                  ! Two times model timestep (2 delta-t)

I have been assuming it's just 1 physics time-step.

adamrher:
@andrewgettelman you'll need to bring in these changes if you want dp1/dp2 to be part of the PPE runs. Currently they are only recognized in marcop_driver, but we don't use macrop_driver in cam6 (we use clubb_intr).

andrewgettelman:
Thanks Adam for point me to this. I'll make sure @Katetc sees it so we can bring these changes in with the PPE for dp1 and dp2

adamrher:
> In tphysac and tphysbc, Is this true:
> real(r8), intent(in) :: ztodt                  ! Two times model timestep (2 delta-t)

This appears to only be true for the Eulerian dycore. So this comment needs to be modified. Similarly, the physics time-step output to the log needs to be doubled for the Eulerian dycore in this PR (this is b/c for the first time-step, the physics time-step is not doubled; it is only doubled for nstep /= 0).

adamrher:
@andrewgettelman @JulioTBacmeister @swrneale should I include the super-saturation fixes https://github.com/ESCOMP/CAM/issues/351 in the PR?

clubb_l_trapezoidal_rule_zm = .false.
clubb_l_trapezoidal_rule_zt = .false.

It will make this PR climate changing, but seems like it might be timely to just incl. it here. Vince seems to be in agreement, that is, still calling the pdf twice (mid-point & interface levels), but turning off the trapezoidal rules since it overwrites the cloud liquid.

andrewgettelman:
I'm a big fan of including the supersaturation fixes. It's a small change, but cleans it up. Unless you don't want this PR climate changing, and just want to check those changes in in a new PR.

I'm not even sure our procedure for a PR that changes climate with switches and doesn't need SE review....

adamrher:
After speaking with Julio, I've turned off the trapezoidal rules, and added issue https://github.com/ESCOMP/CAM/issues/351 to the original PR post.

zarzycki:
First -- super helpful having the timesteps added to atm.log!

Small bug -- I think timestep logic in physpkg.F90 is missing the below block from clubb_intr that first checks if clubb_timestep is greater than the host (in CAM's case, cldmacmic timestep) and if greater, just sets the two equal.

```
   !  Now check to see if dtime is greater than the host model 
   !    (or sub stepped) time step.  If it is, then simply 
   !    set it equal to the host (or sub step) time step.  
   !    This section is mostly to deal with small host model
   !    time steps (or small sub-steps)

   if (dtime > hdtime) then
     dtime = hdtime
   endif
```

As is, the added diagnostics in physpkg.F90 will give you `clubb_timestep = dt_cldmacmic / 2.`
if you define clubb_timestep > dt_cldmacmic, when it should be `clubb_timestep = dt_cldmacmic`. You can verify this by printing the timesteps in CLUBB.

In physpkg, something like...

```
      clubb_timestep_actual = clubb_timestep
!++
      if (clubb_timestep_actual .gt. cld_macmic_ztodt) then
        clubb_timestep_actual = cld_macmic_ztodt
      endif
!--
      if (mod(cld_macmic_ztodt,clubb_timestep_actual) .ne. 0) then
        clubb_timestep_actual = cld_macmic_ztodt/2._r8
        do while (clubb_timestep_actual > clubb_timestep)
          clubb_timestep_actual = clubb_timestep_actual/2._r8
        end do
      endif
```

should work (TBH, I'd probably rather do an if/else block, but the above matches CLUBB's logic to be safe).

adamrher:
Hey Colin. Good catch. I'll fix that before this is merged.

BTW, the timeline for completing this PR is probably a month out. The other CLUBB PR: https://github.com/ESCOMP/CAM/pull/387 is a higher priority than this one, which we are hoping to begin next week.

adamrher:
@zarzycki while I'm thinking of it ... this could all be avoided if I could just export the relevant physics time-steps around, instead of trying to recreate the time-step calculations in ```physpkg```. The problem is that some of the time-steps are defined during run time proper, i.e., after the initialization phase. And I'm choosing to output the time-steps in the initialization of ```physpkg```. We'll see what Steve and Cheryl think of my approach once we begin code review.

adamrher:
we should consider adding these bug fixes and changes to energy fixer from Oksana: https://github.com/E3SM-Project/E3SM/pull/4490#issuecomment-909616841

adamrher:
closing this PR since I started over with an up to date fork. Will open a new PR.

PR #380, opened 2021-05-19
Correction to H2O2 + OH -> H2O + HO2 reaction rate constant

### LINKED ISSUE #372: gas phase reaction rate in pp_trop_mam3 and mam4 is out of date
The rate constant for the reaction:
H2O2 + OH -> H2O + HO2                                           
in the pp_trop_mam3 and 4 "prescribed oxidant" chemistry system is out of date.

In the current code base it is 2.9e-12* exp( -160 / T)
It should be 1.8e-12, according to a more recent JPL database search.

Someone (other than me) should identify the correct value, do some runs to see how much it affects results, and then we should update the code.

fvitt:
The correction is applied to trop_mam3, trop_mam4, trop_mam7, trop_mozart, and waccm_sc_mam4 chemistry mechanisms (chem_mech.in) and chemistry preprocessor generated code is included.  

Added a MAM7 aqua-planet compset for testing purposes.  Otherwise, MAM7 aerosols were not being exercised in test suites.

CAM6 compsets will change answers.  Expect many of the baseline tests to fail.  Will this require a climate test?

fixes #372


fvitt:
@AndrewJConley, I was not able to add you as a review.  Maybe @gold2718 needs to include you in this repo somehow?

AndrewJConley (approved):
Thanks.  The changes to the rate constants look correct.  
It looks like there are some other changes to some compsets.  I don't know anything about those changes.

gold2718 (approved):

andrewgettelman:
@fvitt, we discussed this a bit in the projects meeting: I assume this will affect aerosol oxidation and might change AOD. 

1. Is this a 'bug' (wrong equation) or just an updated reaction rate? Its fine if it is just updated, but it would be useful to know if it is really a bug
2. Before putting this in, can we see a simple F2000climo test of 3-5 years with AMWG diagnostics with and without. I'm guessing this is not a major issue but i'd like to make sure that SO4, AOD changes and subsequent cloud and radiation changes are not large. 

Thanks!

AndrewJConley:
This is an update.

It would be good to have a climo test to see how much it changes the
climate and aerosols.

Thanks.

lkemmons:
I ran a couple years of F2000climo cases with and without the new rate. I created a new mechanism file from the Cafe:
/glade/u/home/emmons/mechanisms/MZ290_CAM_MAM4_20210504.in
The output from the 2 cases are in:
/glade/scratch/emmons/archive/f.e22.F2000climo.f09.cesm220.orig
/glade/scratch/emmons/archive/f.e22.F2000climo.f09.cesm220.newrate/
Case directories are in /glade/u/home/emmons/cesm22_cases/
Diagnostics at: https://acomstaff.acom.ucar.edu/emmons/cesm/f.e22.F2000climo.f09.cesm220.newrate.0002_2-f.e22.F2000climo.f09.cesm220.orig.0002_2/

andrewgettelman:
Greetings,

Thanks for the explanation and the reference to the earlier issue/pr (#372). 

1. The differences are small: if we really wanted to look we would need more than a single year. But it does not look like this change is going to be significant. That's good. 

2. So it seems like this is not a bug, but an update. Do we really want to just update one rate constant? Why does this particular equation need to be updated now, or can/should this be grouped with a larger update that comprehensively moves the chemistry forward. Our process puts an opportunity cost on doing this, but I am okay if Louisa et al want to do this now.

AndrewJConley:
It is there because we will need to test against it during the development
of the CCPP version for SIMA

lkemmons:
This is the only rate in the simple CAM chemistry that is inconsistent with the TS1 and TSMLT1 mechanisms.
The user-define DMS+OH rate constant is a bit out of date, but we have not yet updated it in any of the chemistry mechanisms yet ([usr_DMS_OH]           DMS + OH  -> 0.5*SO2 + 0.5*HO2).

andrewgettelman:
So the motivation is to make TSMLT1 (which is used in part by MAM?) consistent with TS1?

That makes sense to me. Thanks.

lkemmons:
No, it is to make the very simple CAM ("trop_mam") mechanism consistent
with both TS1 and TSMLT1.

fvitt:
@lkemmons,  Thanks for pointing us to a Cafe generated mechanism file for trop_mam4.   Are you going to provide Cafe generated mechanism files for trop_mam3, trop_mam7, and waccm_sc_mam4?

lkemmons:
I was not going to because MAM3 and MAM7 are not currently set up in the
cafe, but I could if needed. I also need to see what WACCM-SC is exactly,
but could do that.

fvitt:
@lkemmons, It is up to you let us know what is needed in terms of Cafe generated mechanism.  This branch currently has hand-edited mechanisms files for trop_mam3, trop_mam7, and waccm_sc_mam4, which is fine for me.

nusbaume (approved):

cacraigucar (commented):
@fvitt and @lkemmons - Have you finalized this PR (does it have what you want in it)?

PR #386, opened 2021-06-04
Update Held_Suarez to CCPP version

### LINKED ISSUE #267: Get Held_Suarez from atmospheric_physics repository
Will use the CCPP'ized Held_Suarez code from the atmospheric_physics repository

cacraigucar:
Minor mods to bring in Held_Suarez

Ran test using the updated Held_Suarez code and using cam_snapshot, the "after" file was identical to the original "after" file.

Closes #267


src/physics/simple/held_suarez_cam.F90
```diff
@@ -40,12 +40,15 @@ subroutine held_suarez_init(pbuf2d)
     use cam_history,        only: addfld, add_default
     use physconst,          only: cappa, cpair
     use ref_pres,           only: pref_mid_norm, psurf_ref
-    use held_suarez,        only: held_suarez_1994_init
+    use held_suarez_1994,   only: held_suarez_1994_init
 
     type(physics_buffer_desc), pointer :: pbuf2d(:,:)
 
+    character(len=512)            :: errmsg
```
nusbaume, line 47: I might replace `512` here with `SHR_KIND_CX`.

src/physics/simple/held_suarez_cam.F90
```diff
@@ -81,6 +84,9 @@ subroutine held_suarez_tend(state, ptend, ztodt)
     real(r8)                           :: pmid(pcols,pver) ! mid-point pressure
     integer                            :: i, k             ! Longitude, level indices
 
+    character(len=512)            :: errmsg
```
nusbaume, line 87: I might replace `512` here with `SHR_KIND_CX`.

nusbaume (approved):
Looks good to me!  The only change I would recommend is replacing the `512` number with a CIME-provided kind.

src/physics/simple/held_suarez_cam.F90
gold2718: @nusbaume, the issue here is that we don't have symbolic sizes in the metadata. Hopefully, this will be fixed at some point.

src/physics/simple/held_suarez_cam.F90
gold2718: Same response.

gold2718 (approved):

src/physics/simple/held_suarez_cam.F90
nusbaume: @goldy that makes sense to me when it comes to CAMDEN, but why would that matter for CAM, given that it is just using the fortran code directly (so the metadata is never involved)?

gold2718: That is true but it is good if the code reflects what is in CCPP so we can easily find and fix these interfaces when the CCPP changes. Not a big deal either way IMHO.

nusbaume: Totally agree that this is not really a big deal.  If @cacraigucar just wants to merge this without worrying about these particular changes then that is fine with me!

peverwhee (approved):

Issue #381, opened 2021-05-26
Pass surface ozone concentrations to the coupler


billsacks:
CTSM would like to be able to receive surface ozone concentrations from the atmosphere. There were some discussions involving @danicalombardozzi @mvertens and @fvitt in late 2016, and a plan was developed at the time, but CESM2 release priorities interfered and this never happened. We would like to restart these discussions.

My impression / recollection is that, when running with CAM-Chem, surface ozone concentrations are readily available. Things get trickier when running CAM without Chem; a note from a few years ago suggested that CAM may still have surface ozone concentrations in that case, but that they may just be monthly averages. I'm not sure what the best approach will be, but we'd like to have some discussions about this.

@danicalombardozzi and I would like to talk with the relevant people to (re)chart a path forward on this: @mvertens @fvitt and anyone else who should be included.

fvitt:
When we run CAM without chemistry, ozone is prescribed for radiation transfer calculations.  The prescribed ozone is typically monthly averaged and are time interpolated.  It seems that we should be able to provide ozone concentrations in the bottom most layer whether the concentrations are prognosed by chemistry or prescribed.

AndrewJConley:
I agree as well.  This is not the only surface flux that needs to be
calculated by, or passed to, the surface models.  With the transition to
NUOPC, it will be good to reevaluate where and how we do surface flux
computations.

AndrewJConley:
Thanks for including me.  I expect Francis can cover the technical aspects
of the specific question of Ozone deposition to land for both prescribed
and prognostic chemistries.  I look forward to the discussion for
informational purposes as we design the next phase of our deposition work
for other chemicals and aerosols.

gold2718:
Should have been closed by #404

PR #404, opened 2021-07-17
cam6_3_045: Pass surface ozone concentrations to coupler

fvitt:
Lowest layer ozone concentrations (PPBV) passed to coupler when CMEPS mediator (NUOPC) is used.

Fixes issue #381
Fixes issue #411


bld/namelist_files/namelist_definition.xml
```diff
@@ -6805,6 +6805,17 @@ List of fluxes needed by the CARMA model, from CLM to CAM.
...
+
+<entry id="atm_ozone_frequency" type="char*64" category="ozone_coupling"
+       group="ozone_coupling_nl" valid_values="instantaneous,monthly_interpolated">
+Frequency of surface ozone field passed from CAM to surface components.
+Surface ozone is passed every coupling interval, but this namelist flag
+indicates whether the timestep-level values are interpolated from a
+coarser temporal resolution.
+Default: set by build-namelist.
+</entry>
+
```
fvitt, line 6818: @billsacks 
For CAM6 (without chemistry) the default prescribed ozone data is zonal means of 5-day average which is time interpolated.   In this case "monthly_interpolated" is not quite accurate.  However, in CTSM it would probably treated the same as time interpolated monthly averaged data.  We may want to think about might be better namelist variable name and and its valid values.

billsacks: Thanks for pointing this out. I'd like to get @danicalombardozzi 's input here. I agree we should aim to be accurate in our naming. Maybe the most important consideration is what CTSM will and won't care about. For example, is the important point just whether the diurnal cycle is resolved, in which case CTSM would treat 5-day averages the same as monthly values? Or might it be useful for CTSM to distinguish between different frequencies like 5-day averages vs. monthly averages?

If we're pretty sure that the only thing CTSM will care about is whether the diurnal cycle is resolved in CAM's field, then maybe we can go with the suggestion that I think someone made earlier, which is to just use this namelist flag to distinguish whether or not the atm-provided ozone concentrations have a diurnal cycle. We could go so far as to make this a logical, though I am always a bit hesitant about using a logical value for something unless I am nearly 100% sure that we won't need a 3rd option in the future.

Maybe a compromise solution would be: keep this a string variable, but for now the options would be something like "instantaneous" and something like "average_of_multiple_days" – with the latter encompassing 5-day averages, monthly averages, and anything similar. Then other options could be added later if the need arises.

Thoughts – especially from @danicalombardozzi ?

src/control/camsrfexch.F90
```diff
@@ -489,16 +493,21 @@ subroutine cam_export(state,cam_out,pbuf)
...
+   cam_out%co2diag(:ncol) = chem_surfvals_get('CO2VMR') * 1.0e+6_r8
    if (co2_transport()) then
       do i=1,ncol
          cam_out%co2prog(i) = state%q(i,pver,c_i(4)) * 1.0e+6_r8 *mwdry/mwco2
       end do
    end if
+
+   ! get bottom layer ozone concentrations to export to surface models
+   call rad_cnst_get_gas(0, 'O3', state, pbuf, o3_ptr)
+   cam_out%ozone(:ncol) = o3_ptr(:ncol,pver) * 1.0e+9_r8 * mwdry/mwo3 ! ppbv
```
billsacks, line 509: Given that you need to convert this anyway, I should say that I don't have any feelings myself on what units this should be in. (This is tied in with my comment here: https://github.com/ESCOMP/CMEPS/pull/212/files#r670863912)

@danicalombardozzi or others - does ppbv make the most sense for units of the O3 concentration or would something else be more standard?

billsacks (approved):
Thanks a lot @fvitt for your quick work on this!

I'm marking this as approved, with the caveats that:
1. There are still some outstanding questions we need to resolve
2. I'm not the best one to review the nuopc cap changes. I'm assigning @mvertens to take a quick look at this.

lkemmons:
My opinions:
namelist: my understanding was that CLM only needed to know whether or not to apply a diurnal cycle (no 3rd option)
units:  ozone is very commonly considered in ppb, but I think here it might be better to stick with mole/mole.

src/control/camsrfexch.F90
lkemmons: I think mol/mol is more standard and consistent with ozone output from CESM.

bld/namelist_files/namelist_definition.xml
lkemmons: I think the only flag needed is to tell clm to apply diurnal cycle or not.
But "instantaneous" and "average_of_multiple_days" is also good.

mvertens (approved):

bld/namelist_files/namelist_definition.xml
danicalombardozzi: I agree that the most important piece is whether the diurnal cycle is resolved in the data or not. Once the data are averaged to daily or multi-day average, CLM will need to apply a diurnal cycle. 

I wonder about using "instantaneous". Does this imply that the data are for each model time step? Would "sub-daily" be a better descriptor for the higher frequency data? If so, the flags would be something like "subdaily" and "multiday_average", highlighting both the data frequency and processing. 

@fvitt You mentioned the data are zonal means. Does this mean that the data are averaged through space? If so, how large are the zones?

src/control/camsrfexch.F90
danicalombardozzi: I agree that mol/mol is relatively standard, and then it's easy to convert to ppb from there.

bld/namelist_files/namelist_definition.xml
danicalombardozzi: One additional question to make sure I understand how this works: Will the default setting of this flag be tied to the compset (e.g., subdaily/instantaneous for CAM-CHEM and multi-day for CAM and CLM-only)? Or will it be fixed to one default value and the user needs to know to change the frequency in the namelist?

billsacks: @danicalombardozzi and @lkemmons - thanks a lot for your input here. I like your names, "subdaily" and "multiday_average". I'd be happy to use those names if others are. Based on input from you and @lkemmons , I could also see changing this to a logical like `atm_ozone_has_diurnal_cycle` (true/false), since it really does sound like that may be all that matters. If you have a preference between the strings vs. logical, let us know: the logical seems more clear to me, but more changes would be needed later if a 3rd distinction were needed in the future.

@danicalombardozzi and @fvitt - I missed the piece that the data are zonal means. To me that implies that there are latitudinal bands of values; @fvitt , is that right? If so, @danicalombardozzi, would you want an additional flag stating this spatial averaging (e.g., so that we could apply spatial anomalies as well as temporal anomalies), or would that be unnecessary? (Or would you go so far as to question the whole value of getting ozone from CAM if all we're getting are zonal means?)

src/control/camsrfexch.F90
billsacks: Thanks, @lkemmons and @danicalombardozzi . So @fvitt let's change this to be in units of mol/mol. I'll change the CMEPS PR.

bld/namelist_files/namelist_definition.xml
billsacks: Okay, I have changed the values of this namelist flag, both in https://github.com/ESCOMP/CESM_share/pull/4 and via a PR to this CAM PR (https://github.com/fvitt/CAM/pull/2) - @fvitt can you please merge that into here if it looks good to you?

The remaining outstanding question is around the spatial resolution and whether we need a new flag to describe that.

src/control/camsrfexch.F90
```diff
@@ -403,14 +405,15 @@ subroutine cam_export(state,cam_out,pbuf)
...
+   use physconst,        only: rair, mwdry, mwco2, gravit, mwo3
    use constituents,     only: pcnst
    use physics_buffer,   only: pbuf_get_index, pbuf_get_field, physics_buffer_desc
+   use rad_constituents, only: rad_cnst_get_gas
 
    implicit none
 
    ! Input arguments
-   type(physics_state),  intent(in)    :: state
+   type(physics_state),  intent(in) :: state
```
gold2718, line 416: Why remove the alignment?

gold2718 (changes requested):
I do not think the requested change to `cam_out%ozone(:ncol)` has been made.

cacraigucar (approved):

billsacks:
The remaining outstanding question here as far as I can tell is whether we're going to want an additional flag describing something about the spatial characteristics of the ozone coupling field. @fvitt I believe @danicalombardozzi was waiting to make a decision on this until she had some more details from you (or someone else) on her last question here: https://github.com/ESCOMP/CAM/pull/404#discussion_r678666694

bld/namelist_files/namelist_definition.xml
lkemmons: I believe the zonal mean (longitude-averaged) fields Francis mentioned are the input oxidant files to CAM.  CAM takes care of interpolating them, so there should be no need for CLM to worry about spatial regridding - CAM will pass ozone mixing ratios on the model grid to CLM, whatever the original source of ozone.

billsacks: @lkemmons thanks for that note. I should have clarified that (if I understood correctly from my conversation with @danicalombardozzi), the concern wasn't whether we'd need to regrid values, but whether a simple interpolation of zonal averages is good enough scientifically for CTSM's needs. What I'm imagining from what you and @fvitt have said is that all grid cells in a given latitude band would have the same surface ozone concentration. Is that right? If so, I'll let @danicalombardozzi comment on whether we'd like to know that in CTSM in order to possibly handle those values differently from how we'd handle values that exhibit some variation within a latitude band.

lkemmons: In general the oxidant fields read by CAM are not zonal averages.  For F2000climo compset, the oxidants file is /glade/p/cesmdata/cseg/inputdata/atm/cam/ozone/tracer_cnst_CAM6chem_2000climo_3D_monthly_c171004.nc which has monthly means at f09 resolution for O3, etc. 
The files for stratospheric ozone and aerosols are usually zonal means, which is probably what Francis was thinking of.

fvitt: @lkemmons 

F2000climo reads ozone for radiation transfer from:
 prescribed_ozone_file  = 'ozone_strataero_CAM6chem_2000climo_zm_5day_c171004.nc'

As well as ozone for chemistry from:
 tracer_cnst_file  = 'tracer_cnst_CAM6chem_2000climo_3D_monthly_c171004.nc'

Which ozone should be used for transfer to the land model? I am assuming the tracer_cnst data should be used.

lkemmons: The tracer_cnst_file which is used for the tropospheric chemistry should be the source of ozone to be passed to CLM. Sorry, I forgot that in CAM O3 is a 'fixed' species, so in a different array than when ozone is simulated (in CAM-chem).

danicalombardozzi: Thanks @lkemmons for the explanation on the zonal averages. This clarifies my question, and it's good to know that tropospheric ozone and oxidants are not zonally averaged. I think the clarification of the different files that @fvitt highlights also answers the questions about temporal resolution -- that CAM fields that will be passed to CLM are monthly averaged, not 5-day averages. 

Given this new/clarified information, it seems that the ozone fields passed from CAM to CLM will (in most circumstances) fall into two categories: 1) Gridded monthly-averaged values (from CAM); or 2) Gridded values every time step (from CAM-CHEM). If my understanding is correct, this is perhaps simpler than we were anticipating, and the namelist flags @billsacks proposed to use should still work.

billsacks: Great! Thank you all for helping to get to the bottom of this!

peverwhee (approved):

nusbaume (approved):
Looks good to me!  Just one minor, optional change requested.

bld/namelist_files/namelist_definition.xml
```diff
@@ -6805,6 +6805,17 @@ List of fluxes needed by the CARMA model, from CLM to CAM.
 Default: set by build-namelist.
 </entry>
 
+<!-- Ozone namelist variables shared between CAM and driver -->
+
+<entry id="atm_ozone_frequency" type="char*64" category="ozone_coupling"
```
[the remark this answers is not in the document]
billsacks: My general rule with things like this is to allow for significantly more space than I think I'll ever need, because being too stingy with characters can lead to some really subtle issues and hard-to-track-down bugs sometimes. Also, since this should be consistent with the length in this other repository (https://github.com/ESCOMP/CESM_share/blob/2e6406de2179391908835b51df8ab1916aed29da/cmeps/shr_ozone_coupling_mod.F90#L47), it would be even more of a pain to change it, since changes would need to be made in multiple repositories, arguing even more for making this value longer than necessary.

I'm starting to talk myself into making this 256, come to think of it ;-)  (just kidding).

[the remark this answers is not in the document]
billsacks: Oh, I should be explicit: the reasoning is in case we need other options in the future that have longer names.

billsacks:
@fvitt thank you for all of your work on this.

The relevant CMEPS PR has just been merged and is available as cmeps0.13.24.

billsacks:
Also, the relevant CESM_share tag is share1.0.5, though that shouldn't actually be needed with this CAM change (the share code change is just for reading the drv_flds_in ozone coupling namelist, and only CTSM should need to call that code, not CAM).

bld/namelist_files/namelist_definition.xml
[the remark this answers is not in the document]
nusbaume: Thanks for the explanation @billsacks!  That sounds reasonable to me.

bld/namelist_files/namelist_definition.xml
dan800: We discussed this work at the AMP priorities meeting today. I wonder if it would make sense to simply have CAM specify ozone passed at each coupling time step (presently 30 minutes), and let CAM calculate a diurnal cycle when it is using prescribed ozone. In CAM-CHEM it would simply pass the prognostic ozone values. This would simplify the coupling infrastructure.

billsacks: @dan800 - Although that would keep the coupling logic simpler, the plan to have CLM calculate the diurnal cycle was based on these reasons:
- There is no value to CAM of having anomalies just in the lowest model level
- It would be expensive to have anomalies applied in all model levels
- It could be confusing to apply anomalies in CAM just in the lowest model level

It's important to note that the anomalies in CLM will be applied via files, so the i/o burden (and disk storage burden) could be large if CAM needed to read in these anomalies for all levels, and the feeling is that it would be confusing to have CAM read & apply anomalies only for the lowest level.

That said, I don't feel like this was set in stone, so if people want to revisit that, we can do so.

gold2718: > It could be confusing to apply anomalies in CAM just in the lowest model level

One idea is that CAM could apply the diurnal cycle to just the lowest model level as part of its export step. The cost would then be the same as for CLM and it would not affect the CAM run.
@fvitt, when CAM is _not_ doing active ozone chemistry, is the species used at all? If not, we could just set up the value (read plus dirunal cycle) as part of the export step.

dan800: @danicalombardozzi Thanks for the clarification regarding calculating hourly anomalies. When running CAM, it seems there is a risk in decoupling the diurnal cycle (in one file) from the mean (in another file - presumably the one used to drive sw radiative heating). Having to change both CLM and CAM namelists to make them consistent under various emission scenarios seems like a potential source of model inconsistency. In generating the ozone dataset for RT from a CAM-CHEM run, it might be best to also include the diurnal cycle at the surface. In terms of data storage this would be no more than adding another radiatively active gas (24 hourly samples is less than the number of levels). Even if the files are split, at least having the two files in the same namelist would make it more obvious that the datasets are consistent.

gold2718: > So we'd probably want to keep the diurnal cycle dataset and code to apply it in sync between DATM and CAM. Not a huge deal, but a small downside.

Given proper coding this in DATM (CDEPS version), couldn't CAM just `use` that code to grab ozone (with diurnal cycle already applied) for export to CTSM? Alternatively, if CAM reads in the ozone, it could then call the DATM version of the diurnal cycle code.

billsacks: I wondered about that. My impression (maybe wrong) is that CAM can't use anything in DATM itself, but I agree that the apply-diurnal-cycle code could be in shared code, either in CDEPS or in the share code. I'm thinking that the specification of the dataset (namelist default) may still need to be duplicated between the two, because we wouldn't normally have a datm namelist if running with CAM.

gold2718 (approved):
I have a whitespace question but otherwise, this looks okay.

bld/namelist_files/namelist_definition.xml
billsacks: Since this PR was just merged, does that imply that this conversation was resolved among the CAM SEs in favor of the original plan of having the diurnal anomalies applied in CTSM rather than in CAM? Or is this still an open question that might be revisited in a follow-on set of changes?

PR #418, opened 2021-08-13
cam6_3_030: Update the MPAS-A dycore to MPAS-Atmosphere v7+

### LINKED ISSUE #417: Update the MPAS-A dycore in CAM to MPAS-Atmosphere v7+
The MPAS-A dycore in CAM is currently based on MPAS-Atmosphere v6, while a SIMA milestone requires that the dycore be updated to v7+. Updating the MPAS-A dycore to a stable hash on the MPAS-Model 'develop' branch will suffice.

mgduda:
Update the MPAS-A dycore in CAM to a stable hash on the MPAS-Model's 'develop' branch.

With the update to a newer MPAS-A dycore, several changes are needed in CAM-MPAS:
1. New MPAS-A namelist options need to be handled in CAM-MPAS
2. The dycore Makefile needs to correctly account for new dependencies in the dycore
3. Additional calls to MPAS-A init/finalize routines need to be made at appropriate points in CAM-MPAS
4. Acquisition of pointers to certain dycore fields need to be moved to account for new memory allocation strategies in the updated dycore

This PR closes #417.

### DIFF
--- a/src/dynamics/mpas/dyn_comp.F90
+++ b/src/dynamics/mpas/dyn_comp.F90
@@ -402,10 +402,6 @@ subroutine dyn_init(dyn_in, dyn_out)
    call mpas_pool_get_array(diag_pool,  'uReconstructZonal',      dyn_in % ux)
    call mpas_pool_get_array(diag_pool,  'uReconstructMeridional', dyn_in % uy)
 
-   call mpas_pool_get_array(tend_physics_pool, 'tend_ru_physics',     dyn_in % ru_tend)
-   call mpas_pool_get_array(tend_physics_pool, 'tend_rtheta_physics', dyn_in % rtheta_tend)
-   call mpas_pool_get_array(tend_physics_pool, 'tend_rho_physics',    dyn_in % rho_tend)
-
    ! Let dynamics export state point to memory managed by MPAS-Atmosphere
    ! Exception: pmiddry and pintdry are not managed by the MPAS infrastructure
 
@@ -465,6 +461,13 @@ subroutine dyn_init(dyn_in, dyn_out)
    end if
 
    call cam_mpas_init_phase4(endrun)
+
+   !
+   ! Set pointers to tendency fields that are not allocated until the call to cam_mpas_init_phase4
+   !
+   call mpas_pool_get_array(tend_physics_pool, 'tend_ru_physics',     dyn_in % ru_tend)
+   call mpas_pool_get_array(tend_physics_pool, 'tend_rtheta_physics', dyn_in % rtheta_tend)
+   call mpas_pool_get_array(tend_physics_pool, 'tend_rho_physics',    dyn_in % rho_tend)
 
    ! Check that CAM's timestep, i.e., the dynamics/physics coupling interval, is an integer multiple
    ! of the MPAS timestep.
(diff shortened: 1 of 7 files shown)


src/dynamics/mpas/dyn_comp.F90
```diff
@@ -1401,6 +1412,42 @@ subroutine cam_mpas_namelist_read(namelistFilename, configPool)
...
+         read(unitNumber, assimilation, iostat=ierr2)
+         if (ierr2 /= 0) then
+            call endrun(subname // ':: Failed to read namelist group &assimilation')
+         end if
+      else
+         call endrun(subname // ':: Failed to find namelist group &assimilation')
+      end if
+   end if
+
+   call mpi_bcast(mpas_jedi_da, 1, mpi_logical, masterprocid, mpicom, mpi_ierr)
```
cacraigucar, line 1447: The mpi_ierr should probably be checked for an error

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -1401,6 +1412,42 @@ subroutine cam_mpas_namelist_read(namelistFilename, configPool)
...
+         if (ierr2 /= 0) then
+            call endrun(subname // ':: Failed to read namelist group &limited_area')
+         end if
+      else
+         call endrun(subname // ':: Failed to find namelist group &limited_area')
+      end if
+   end if
+
+   call mpi_bcast(mpas_apply_lbcs, 1, mpi_logical, masterprocid, mpicom, mpi_ierr)
+
```
cacraigucar, line 1430: Should probably check the mpi_ierr for an error

src/dynamics/mpas/dyn_comp.F90
mgduda: I didn't add checks on `mpi_ierr` here specifically because we weren't checking `mpi_ierr` after other calls to `MPI_Bcast` for namelist variables in the [&nhyd_model](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/mpas/dyn_comp.F90#L1308-L1340), [&damping](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/mpas/dyn_comp.F90#L1390-L1395), [&decomposition](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/mpas/dyn_comp.F90#L1418), [&restart](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/mpas/dyn_comp.F90#L1436), or [&printout](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/mpas/dyn_comp.F90#L1457-L1459) groups.

If we add checks for the broadcast of `mpas_jedi_da` and `mpas_apply_lbcs`, I think we should include checks for all broadcasts in the `cam_mpas_namelist_read` routine for consistency. Do you have any preference for whether we should check `mpi_ierr` after each broadcast or check an aggregated error code for each namelist group?

cacraigucar: Typically in CAM physics, we do MPI checks after each MPI call (in recent code).  @gold2718 do you see any issues if they decide to aggregate the checks?

mgduda: For what it's worth, it doesn't look like the SE dycore checks errors after [broadcasting its namelist options](https://github.com/ESCOMP/CAM/blob/cam6_3_028/src/dynamics/se/dyn_comp.F90#L259-L304), so what we're doing in MPAS is at least consistent with SE.

cacraigucar: Since I am not very familiar with the dycore side of things, let's see what @gold2718 has to say about all of this.

gold2718: This is a weird topic. We have stopped putting in the error checks because, at least with the MPI versions we are currently using, they do not work from Fortran. I believe newer versions do work but we would have to try calling `MPI_Comm_set_errhandler` to check (once sysadmins get around to building compilers using the newer versions).
That said, I do not have an issue with leaving out the checks because the default behavior is to abort so I do not see a huge advantage to adding these checks.

cacraigucar: I did not know about this problem with MPI and error handling.  I am fine with not adding the checks.

fvitt (approved):

Issue #423, opened 2021-08-30
Transition tests to NUOPC


gold2718:
Since NUOPC will soon become the default CESM driver, CAM needs to transition its test suite to NUOPC. Currently, the plan is:
- Run all tests in both MCT and in NUOPC.
- Update externals to cesm2_3_alpha05b (successfully running all the NUOPC tests requires updating cime and all the NUOPC externals)
    - Comparing baselines with updated NUOPC externals is not straightforward because the test names have been changed
    - Will probably need to create new baseline directories to do straight baseline check or write a script to check the discrepancies.
- Fix existing NUOPC issues (#411, #412)
- Miscellaneous (_next tag_) fixes (e.g., #422).
- Create a minimal, `aux_cam_mct` test suite to ensure CAM/MCT is not broken
- Make NUOPC the default driver for regression tests (in test_driver.sh)

fischer-ncar:
You can easily run nuopc tests by setting the environment variable CIME_DRIVER=nuopc.  This way you won't need to edit the test lists when running both mct and nuopc tests.

gold2718:
@fischer-ncar: Our test framework already handles running tests in MCT or in NUOPC. However, currently, all of our baselines are MCT-only. This work is meant to transition our baselines to NUOPC and the purpose of creating a new testlist is to maintain only a few MCT tests once NUOPC becomes the default driver.

The issue with the new test names is that our current baselines do not have the `_Vmct` (or `_Vnuopc`) string.

gold2718:
**Update: The PGI issue with running using the NUOPC driver is not going to be fixed for CESM2.3, so that release will _not_ have PGI support.**

To move forward with CAM, I propose we move our NUOPC driver aux_cam pgi tests to gnu. To  do this, I propose the following set of steps:
1. In a checkout of cam6_3_031, change the aux_cam pgi tests to gnu and create a set of MCT baselines. Creating an archived version might need to be done by hand.
2. In the branch for this issue, change the aux_cam pgi tests to gnu
3. Change the default compiler in archive_baseline.sh to gnu (and maybe update some of the instructions).
4. Run tests and compare with baselines made using tag (we still expect some answer changes and this will allow us to note them).

@fischer-ncar, do you want me to go ahead and change the prebeta pgi test to gnu as well?

[...]

PR #427, opened 2021-09-02
cam6_3_031: Update externals to cesm2_3_alpha05c

gold2718:
Update Externals.cfg to match CESM 2.3 alpha 05c
Addresses #423 

Testing is a bit tricky since the test names are different in the new CIME. To allow checks to pass, I created a set of dummy baseline directories that is mostly a renamed version of the cam6_3_030 baselines:
- Test directories have the new name (with `_Vmct` added at the end of the test type field)
- Files inside the top level of the test directory are soft links to the cam6_3_030 baseline files (saves disk space)
- Files inside CaseDocs are copies of the cam6_3_030 baseline files _except_:
    - esp_modelio.nml: new value for `pio_netcdf_format`
    - drv_in: new `case_name` that includes driver type

Because of the updated component externals, there are numerous namelist changes (GLC, CLM, and driver) and a few baseline changes (I believe due to CLM changes but I will have to double check).

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -11,7 +11,7 @@ protocol = git
 repo_url = https://github.com/ESCOMP/CESM_CICE
 local_path = components/cice6
 externals = Externals.cfg
-required = True
+required = False
 
 [cmeps]
 tag = cmeps0.13.25
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,56 +1,104 @@
-[externals_description]
-schema_version = 1.0.0
-
 [cice]
-tag = cice5_20200430
+tag = cice5_20210802
 protocol = git
 repo_url = https://github.com/ESCOMP/CESM_CICE5
 local_path = components/cice
 required = True
 
-[cime]
-tag = cime5.8.44
+[cice6]
+tag = cesm_cice6_2_0_5
 protocol = git
-repo_url = https://github.com/ESMCI/cime.git
+repo_url = https://github.com/ESCOMP/CESM_CICE
+local_path = components/cice6
+externals = Externals.cfg
+required = True
+
+[cmeps]
+tag = cmeps0.13.25
+protocol = git
+repo_url = https://github.com/ESCOMP/CMEPS.git
+local_path = components/cmeps
+required = True
+
+[cdeps]
+tag = cdeps0.12.21
+protocol = git
+repo_url = https://github.com/ESCOMP/CDEPS.git
+local_path = components/cdeps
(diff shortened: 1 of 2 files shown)


Externals.cfg
```diff
@@ -1,60 +1,111 @@
...
-
 [cice]
-tag = cice5_20200430
+tag = cice5_20210802
 protocol = git
 repo_url = https://github.com/ESCOMP/CESM_CICE5
 local_path = components/cice
 required = True
 
+[cice6]
```
cacraigucar, line 8: Just confirming that we need both cice and cice6?  Are any of our F compsets using cice6 yet?  If not, perhaps we wait to add the cice6 external until we move over to using it.

fischer-ncar: You can also set required = False for cice6, so you have a place holder.  Then it'll be easy to do a -o checkout when cice6 is needed for development.  Then set required = True when it's ready to go prime time.

gold2718: I like the idea of making it optional. @cacraigucar?

Externals.cfg
```diff
@@ -1,60 +1,111 @@
...
+protocol = git
+repo_url = https://github.com/ESCOMP/CESM_CICE
+local_path = components/cice6
+externals = Externals.cfg
+required = True
+
+[cmeps]
+tag = cmeps0.13.25
+protocol = git
+repo_url = https://github.com/ESCOMP/CMEPS.git
```
nusbaume, line 19: Is there a reason some of these URLs end in `.git` while the rest don't?

nusbaume (approved):
Looks fine to me!  I just had one (not particularly important) question.

Externals.cfg
cacraigucar: Optional seems like a reasonable way to go

Externals.cfg
gold2718: I just copied from the CESM Externals.cfg. Using git blame, I would ask:
@jedwards4b, @ekluzek, is there any particular reason cmeps and cdeps have the '.git' extension?

jedwards4b: I believe that the extension is optional, it will work either way.

ekluzek: No there really is no reason for it. Both work. Personally I've been adding the .git because I think it's more explicit. It also serves as a way to point out the end to the repo name. But there really isn't much of a reason to prefer one over the other. You might want to go with consistency and have them all appear the same

peverwhee (approved):

cacraigucar (approved):

PR #433, opened 2021-09-10
Fix issue with input data reader when ESMF lib is used and CALENDAR is GREGORIAN

### LINKED ISSUE #317: error in mo_srf_emissions using ESMF library with leap year
When running the aux_cam testlist on cheyenne - the following test:
SMS_D_Ln9_Vnuopc.f09_f09_mg17.FCts2nudged.cheyenne_intel.cam-outfrq9s_leapday
Resulted in the following traceback in the cesm.log file
```
40:cesm.exe           000000000B757464  shr_abort_mod_mp_         114  shr_abort_mod.F90
40:cesm.exe           000000000B7572C2  shr_abort_mod_mp_          61  shr_abort_mod.F90
40:cesm.exe           0000000002A9EB37  time_manager_mp_c        1242  time_manager.F90
40:cesm.exe           0000000002A93AAD  time_manager_mp_s         271  time_manager.F90
40:cesm.exe           0000000002AEBCB9  tracer_data_mp_re        1200  tracer_data.F90
40:cesm.exe           0000000002AE121E  tracer_data_mp_ad         704  tracer_data.F90
40:cesm.exe           0000000006099E17  mo_srf_emissions_         309  mo_srf_emissions.F90
40:cesm.exe           000000000320C071  chemistry_mp_chem        1102  chemistry.F90
40:cesm.exe           000000000116DFD1  physpkg_mp_phys_t        2715  physpkg.F90
40:cesm.exe           0000000001139B93  physpkg_mp_phys_r        1036  physpkg.F90
40:cesm.exe           00000000007D8D8A  cam_comp_mp_cam_r         258  cam_comp.F90
40:cesm.exe           0000000000786E10  atm_comp_nuopc_mp         845  atm_comp_nuopc.F90
```

The atm log file contained the following output before the crash:
 ```
emiss_shipping
 READ_NEXT_TRCDATA emiss_bb
 READ_NEXT_TRCDATA emiss_res_tran
 set_time_float_from_date: error return from ESMF_TimeSet for set_time_float_from_date
 ERROR: CHKRC
```
And finally, the PET log file contained the following error message - which points to the source of the problem.
 PET0222 ESMCI_Calendar.C:1059 ESMCI::Calendar::convertToTime() Input argument out of range  - ; Gregorian: for Febru\ary 2010, dd=29 > 28 days in the month

It seems that the mct esmf_wrf_timemgr is not picking up these type of failures (the same issue arose in a CTSM test) and using the ESMF library for time management triggers the expected error.

### PR DESCRIPTION
**Pull request by fvitt**
When interp_missing_months is used the time interpolation may try to use 29 Feb on non-leap years when the Gregorian calendar is used.  Here checks are added to avoid this.

Fixes #317

### DIFF
--- a/src/chemistry/utils/tracer_data.F90
+++ b/src/chemistry/utils/tracer_data.F90
@@ -16,7 +16,7 @@ module tracer_data
   use cam_logfile,      only : iulog
 
   use physics_buffer,   only : physics_buffer_desc, pbuf_get_field, pbuf_get_index
-  use time_manager,     only : set_time_float_from_date, set_date_from_time_float
+  use time_manager,     only : set_time_float_from_date, set_date_from_time_float, is_leapyear
   use pio,              only : file_desc_t, var_desc_t, &
                                pio_seterrorhandling, pio_internal_error, pio_bcast_error, &
                                pio_char, pio_noerr, &
@@ -115,7 +115,7 @@ module tracer_data
      integer, pointer, dimension(:) :: count0_x=>null(), count0_y=>null()
      integer, pointer, dimension(:,:) :: index0_x=>null(), index0_y=>null()
      logical :: dist
-     
+
      real(r8)                        :: p0
      type(var_desc_t) :: ps_id
      logical,  allocatable, dimension(:) :: in_pbuf
@@ -751,7 +751,7 @@ contains
             enddo
            endif
         endif
-   
+
         call mpi_bcast(file%weight_x, plon*file%nlon, mpi_real8 , mstrid, mpicom,ierr)
         if (ierr /= 0) call endrun(sub//": FATAL: mpi_bcast: file%weight_x")
         call mpi_bcast(file%weight_y, plat*file%nlat, mpi_real8 , mstrid, mpicom,ierr)
@@ -1281,6 +1281,7 @@ contains
     logical :: times_found
 
     integer :: cur_yr, cur_mon, cur_day, cur_sec, yr1, yr2, mon, date, sec
+    integer :: use_day
     real(r8) :: series1_time, series2_time
     type(file_desc_t) :: fid1, fid2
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by cacraigucar — APPROVED** [review]

File: src/utils/time_manager.F90 [hunk 1]
```diff
@@ -314,6 +315,14 @@ subroutine set_date_from_time_float( time, year, month, day, sec )
 
 endsubroutine set_date_from_time_float
 
+!=========================================================================================
+
+logical function is_leapyear( yr )
+  integer, intent(in) :: yr
+  is_leapyear = mod(yr,100)/=0 .and. mod(yr,4)==0
```
nusbaume, line 322: I believe a leap year can be evenly divisible by 100 if it is also evenly divisible by 400 as well (e.g. the year 2000).  Thus I think this check should look something like:

 ```suggestion
    is_leapyear = .false.
    if (mod(yr,4)==0) then
       if (mod(yr,100)==0) then
         if(mod(yr,400)==0) then
           is_leapyear = .true.
         end if
       else
         is_leapyear = .true.
       end if
    end if
```

**Review by nusbaume — CHANGES_REQUESTED** [review]
I think the `is_leapyear` function is missing a 400 year check.  Everything else looks good, though.

File: src/utils/time_manager.F90 (hunk 1 above)
Comment (gold2718) [reply src/utils/time_manager.F90:322]: I agree with the 400 year exception but usually use a non-nested logic:
```suggestion
  if (mod(yr, 400) == 0) then
    is_leapyear = .true.
  else if (mod(yr, 100) == 0) then
    is_leapyear = .false.
  else if (mod(yr, 4) == 0) then
    is_leapyear = .true.
  else
    is_leapyear = .false.
  end if
```
If you _really_ want to put it on one line:
```suggestion
  is_leapyear = (mod(yr, 400) == 0 .or. mod(yr,100) /= 0) .and. mod(yr,4)==0
```

peverwhee (approved):

src/chemistry/utils/tracer_data.F90
```diff
@@ -1312,8 +1313,17 @@ subroutine read_next_trcdata( flds, file )
           call set_date_from_time_float(file%datatimem, yr1, mon, date, sec )
           call set_date_from_time_float(file%datatimep, yr2, mon, date, sec )
 
-          call set_time_float_from_date( series1_time, yr1, cur_mon, cur_day, cur_sec )
-          call set_time_float_from_date( series2_time, yr2, cur_mon, cur_day, cur_sec )
+          use_day=cur_day
+          if ( cur_mon==2 .and. cur_day==29 .and. .not.is_leapyear(yr1)) then
+             use_day=28
+          end if
```
gold2718, line 1319: Is there any reason to not move this logic to `set_time_float_from_date`?

gold2718 (commented):
Not quite sure if this is a question or a change request but let's start off with question.

src/chemistry/utils/tracer_data.F90
fvitt: Not sure.  I can move it to the set_time_fload_from_date routine if you think it belongs there.

src/utils/time_manager.F90
fvitt: I have included the 400 year exception.

gold2718 (approved):

nusbaume (approved):
Thanks for taking care of my concern!

PR #435, opened 2021-09-10
Transport CL and Br when mid-atmos chem is used in cesm2.1

### LINKED ISSUE #314: Need to transport CL and BR
Chemical species CL and BR need to be transported in all configurations.

We have not been transporting CL and BR in the WACCM middle atmosphere configurations, including D-region chemistry configuration, (waccm_ma and waccm_mad). The assumption that these species do not need to be transported breaks down in the upper atmosphere (above 80 km or so) where species are not well mixed.  This has lead to unrealistic concentrations of chlorine and bromine simulations that use WACCM middle atmosphere chemistry.

The chemistry mechanisms that need to change are:
-  waccm_ma
-  waccm_ma_mam4
-  waccm_mad
-  waccm_mad_mam4

fvitt:
Remove CL and BR from the not-transported lists in middle-atmosphere chemistry mechanisms.

Addresses #314 for CESM2.1


cacraigucar (approved):
All files are pp files, so easy approval

test/system/test_driver.sh
```diff
@@ -937,7 +937,7 @@ if [ -n "${submit_script_cb}" ]; then
     case $hostname in
         # cheyenne
         chey* | r* )
-            batch_queue_submit='qsub '
+            batch_queue_submit='qsub -V '
```
fvitt, line 940: @cacraigucar I need to include the `-V` for cheyenne.  Otherwise, the stand-alone cam tests don't work correctly.

cacraigucar: Just curious - what changed since you last made a CESM2_1 branch tag back in September, do you know?

fvitt: I needed the same changes for the `cam_cesm2_1_rel_56` also.  I just did not include the changes to `test_driver.sh` in the tag.

cacraigucar: In that case, I wholeheartedly approve this change.  I'm not surprised that there are changes like this needed for the branch which is so old.

gold2718: > @cacraigucar I need to include the `-V` for cheyenne. Otherwise, the stand-alone cam tests don't work correctly.

To me, this sounds like an error in the submit script. Any environmental variables from test driver that are not resolved in the script (i.e., the environmental variable is still there) should be evaluated before being written to the script. For example, if the script depends on variable, `FOO`:
```
export FOO=bar
```
then we should not see `$FOO` or `${FOO}` in the script, we should see `bar`.
I think it is a lot more robust to do this rather than depend on the environment. Thoughts?

fvitt: @gold2718 You make a good point.  Given that this is for a machine that does not have much life left and for a branch that has very little development activity, I prefer to go with my simple fix.  I don't want to spend any more time on this.  
@cacraigucar @nusbaume Thoughts?

gold2718: If it is not fixed now, I suggest opening an issue to get this fixed.

cacraigucar: I agree to move ahead with this and @fvitt can open an issue detailing what he saw.

nusbaume (approved):

cacraigucar (approved):

PR #440, opened 2021-09-27
Update Externals_CAM.cfg

Qiyu8:
The repo_url is `github `, but the protocal is `svn`, which leads to unsuccessful download.


gold2718:
This change is impractical because `git` currently downloads the entire COSP repository which is quite large (on the order of the entire CESM sandbox).  Please note that a subversion client is a prerequisite for using CESM (https://escomp.github.io/CESM/versions/cesm2.2/html/introduction.html#cesm2-software-operating-system-prerequisites).

PR #480, opened 2021-12-09
cam6_3_038: Change CISM%NOEVOLVE to SGLC

### LINKED ISSUE #204: Convert F-compsets to use stub glacier instead of CISM%NOEVOLVE
We had a meeting with @billsacks and @mvertens today to address issues with long test suite build times and calendars in compsets that use CISM%NOEVOLVE. We decided that the long term goal would be to build a NUOPC-based data glacier component that used active CISM output. But, this is likely several months to a year out. To address current issues, the LIWG is comfortable with other components moving to use stub glacier in their compsets temporarily. 

I can update the F cases (and any other CAM-owned compsets that use CISM currently) to use stub glacier. This will not go into the 2.2 release. It will change all baseline answers slightly, very likely.

Katetc:
Changes compsets that use CISM%NOEVOLVE to use SGLC in the F-compsets managed by CAM. Passes aux_cam test suite on Cheyenne but has baseline changes for all compsets changed.

Fixes #204


fvitt (approved):

adamrher (approved):
Looks good to me. The lively discussion @cacraigucar is referring to ended in agreement. Once the new CTSM surface datasets are generated (soon'ish) then the sub-grid glacier fraction's per elevation class will be derived from the same dataset as used in NO_EVOLVE runs. That is, the differences between NO_EVOLVE and SGLC will be minimal, making the glacier people happy.

billsacks (approved):

PR #481, opened 2021-12-09
cam6_3_040: MPAS pressure fix

### LINKED ISSUE #442: CAM Crashes with 58 levels and higher horiz resolution
Opening an issue to describe crashes with high vertical resolution.

So far this has only been seen with higher resolution simulations, and with CAM-MPAS. 

The basic test case is 58L CAM-MPAS aquaplanet crashes almost immediately with an error from CLUBB: 

The errors are coming out of CLUBB ( we are not necessarily convinced it's CLUBB's fault yet) in advance_windm_edsclrm_module.F90. 

The error is:

405: Fatal error solving for eddsclrm
405: Error in advance_windm_edsclrm 

The error has been seen by @skamaroc and Xingying Huang (not sure their github names yet). 

Vince Larson notes that:

> edsclrm is CLUBB's array of scalars diffused by CLUBB's eddy diffusivity.  windm is CLUBB's representation of the horizontal wind components.  ("m" in CLUBB-speak denotes "grid-mean."  So "windm" refers to the grid-mean values of u and v.)
>
> I am guessing that the wind is the problem, not the eddy scalars, which are chemical species, etc.  

The cause could be initial conditions (not initializing CLUBB variables). Or it could be upstream of CLUBB (and be the input winds).

Still trying to debug....

### LINKED ISSUE #487: MPAS total energy diagnostics in error
The MPAS history variables which capture different aspects of the MPAS energy budget are incorrect.  The variables in question are SE_dBF, SE_dAM, KE_dBF, KE_dAP, KE_dAM, WV_dBF, WV_dAP, WV_dAM.  The error is due to swapped level and cell dimensions in the rho_d and dz arrays in the tot_energy subroutine of the MPAS version of dp_coupling.

jtruesdal:
MPAS uses hydrostatic interface pressures and non-hydrostatic mid level pressures when passing state to CAM physics (this ensures that geopotential_t will exactly recover MPAS z levels from CAM physics state); unfortunately, the hydrostatic interface pressures did not bound the non-hydrostatic mid level pressure at high resolution leading to a physically inconsistent pressure field. Mid-level pressure computation changed to the arithmetic mean of hydrostatic interface pressures. This makes sure that the mid level pressure is bounded (but violates the geopotentian_t consistency for mid level height). 
This PR also addresses an error in the calculation of some MPAS energy diagnostics due to swapped array dimensions in the rhod and dz arrays.
closes #442
closes #487


src/dynamics/mpas/dp_coupling.F90
```diff
@@ -750,20 +751,22 @@ subroutine hydrostatic_pressure(nCells, nVertLevels, zz, zgrid, rho_zz, theta_m,
...
-        pmid(k,iCell)    = (rhok*rgas*thetavk*kap1)**kap2                     !mid-level pressure
-        theta            = theta_m(k,iCell)/(1.0_r8 + Rv_over_Rd * q(k,iCell))!potential temperature
-        pmiddry(k,iCell) = (rhodryk*rgas*theta*kap1)**kap2                    !mid-level dry pressure
+        pint   (k)       = pint   (k+1)       + gravity * rhok    * dz(k)
+      end do
+
+      do k = nVertLevels, 1, -1
+        !hydrostatic mid-level pressure - MPAS full pressure is (rhok*rgas*thetavk*kap1)**kap2 
+        pmid   (k,iCell) = 0.5_r8*(pint(k+1)+pint(k))       
+        !hydrostatic dry mid-level dry pressure - MPAS non-hydrostatic dry pressure is pmiddry(k,iCell) = (rhodryk*rgas*theta*kap1)**kap2
```
cacraigucar, line 768: I believe this comment might be longer than 132 characters.  If so, please break it into two lines

cacraigucar (changes requested):

src/dynamics/mpas/dp_coupling.F90
jtruesdal: Good eye 137.  Split.

cacraigucar (approved):

doc/ChangeLog
```diff
@@ -1,4 +1,62 @@
 ===============================================================
+Tag name: cam6_3_0XX
+Originator(s): jet
+Date: 10 Dec 2021
```
fvitt, line 4: Should this date be the date the new cam tag is created?

fvitt (approved):

doc/ChangeLog
jtruesdal: I believe so.  Now just a placeholder till I get a tag. Thanks for noting it though I will make sure its corrected.

peverwhee (approved):

src/dynamics/mpas/dp_coupling.F90
```diff
@@ -821,13 +825,13 @@ subroutine tot_energy(nCells, nVertLevels, qsize, index_qv, zz, zgrid, rho_zz, t
         dz(k,iCell)   = zgrid(k+1,iCell) - zgrid(k,iCell)
         zcell         = 0.5_r8*(zgrid(k,iCell)+zgrid(k+1,iCell))
         rhod(k,iCell) = zz(k,iCell) * rho_zz(k,iCell)
-        rho_dz        = (1.0_r8+q(index_qv,k,iCell))*rhod(iCell,k)*dz(iCell,k)
+        rho_dz        = (1.0_r8+q(index_qv,k,iCell))*rhod(k,iCell)*dz(k,iCell)
```
gold2718, line 828: I see this bug fix explained in the ChangeLog, however, there should be an open issue that this should address (unless there is no effect from these changes). Can you add that to this PR?

gold2718 (approved):
Code looks okay, however, there is missing issue for the bug fix. Please find the issue or create one and add it to this PR's description section.

src/dynamics/mpas/dp_coupling.F90
jtruesdal: I added issue #487 that describes the bug in these energy diagnostics and referenced it as being closed in this PR description.  These variables are not output by default and don't affect the model run.  The error was caught while I was putting this PR together.

nusbaume (approved):

Issue #461, opened 2021-11-05
DP_FRAC and SH_FRAC not initialized in clubb_intr


adamrher:
clubb_intr.F90 is basically a merged version of macrop_driver.F90 and cloud_fraction.F90, and neither of their run-time modules are being called in CAM6 anymore. It turns out the pbuf vars ```DP_FRAC``` and  ```SH_FRAC```, which are intilized to zero in cloud_fraction.F90, are not being initialized in clubb_intr.F90. This isn't a problem so much for ```DP_FRAC``` because there is a limiter that filters out bogus small values:

         if (deepcu(i,k) <= frac_limit .or. dp_icwmr(i,k) < ic_limit) then
            deepcu(i,k) = 0._r8
         endif

But this is just serendipitous. shalcu or ```SH_FRAC``` has no such limiter so it's is being passed around the code and all sorts of unhinged behavior may be happening under our nose.

I'll point out that this reinventing of cloud_fraction in clubb_intr has already caused some issues (the namelist vars ```cldfrac_dp1``` and ```cldfrac_dp2``` became uncoupled because they were hard coded into clubb_intr ... this was fixed by Kate's PPE PR). And it is really ugly that the ```DP_FRAC``` and ```SH_FRAC``` are registered into the pbuf using the cloud_fraction.F90 register subroutine, even though we no longer use the cloud_fraction.F90 runtime subroutines.

@mikaelwitte

swrneale:
I also think we may want to consider that CAM6 PRECL is not really large-scale cloud in the sense it has been in the past, because ShCU is now part of CLUBB. I would bet that PRECL is being compared somewhere between CAM5 and CAM6 with some head scratching. Maybe change to PREC_NODEEP or something?

PR #483, opened 2021-12-09
Initialize wet dep variables

### LINKED ISSUE #482: Initialize a couple of wet dep variables
(1) ```dgnumwet``` was initialized to 0, and I'm proposing initializing it with a non-zero value instead. In the first time-step w/ the new re-ordering https://github.com/ESCOMP/CAM/pull/387, wet deposition is not called before radiation, and so radiation (via modal_aero_opt.F90) takes the ```log(dgnumwet)``` = NAN. This is not a problem in the current code because the initialized values of dgnumwgt are overwritten by wet deposition before the first radiation call.

(2) ```qaerwat``` was never initialized, and so I'm proposing to initialized it to zero. I don't have as a good a handle on this issue, but for some reason I exposed this problem in my debugging runs; modal_aero_opt is deriving all sorts of optical properties from the initialized value of ```qaerwat```, and it was never initialized. It looks to me like wet deposition changes some values of ```qaerwat```, but not all of them. So I think the non-initialized ```qaerwat``` values are retained for some parts of the arrays in the current code, and I think that's why it changes answers now that I am initializing it to zero.

In sum, my theory is that (1) shouldn't change answers in the current code and (2) should. And that the answer changes in (2) are really just changing some weird uninitialized small number like 2E-349 to zero. And so (2) is a bug fix in the current code. w.r.t (1) it is still open for discussion on whether this change should only be for ```cam_dev``` physics https://github.com/ESCOMP/CAM/pull/387, or for all cases.

cacraigucar:
The pbuf field qaerwat was never initialized.  Initialize it.

Closes #482

### DIFF
--- a/src/chemistry/utils/modal_aero_wateruptake.F90
+++ b/src/chemistry/utils/modal_aero_wateruptake.F90
@@ -22,16 +22,19 @@ private
 save
 
 public :: &
-   modal_aero_wateruptake_init, &
-   modal_aero_wateruptake_dr,   &
-   modal_aero_wateruptake_sub,  &
-   modal_aero_kohler
+   modal_aero_wateruptake_init,  &
+   modal_aero_wateruptake_dr,    &
+   modal_aero_wateruptake_sub,   &
+   modal_aero_kohler,            &
+   modal_aero_wateruptake_rmmin, &
+   modal_aero_wateruptake_rmmax
 
 public :: modal_aero_wateruptake_reg
 
+real(r8), parameter :: modal_aero_wateruptake_rmmin = 0.01e-6_r8
+real(r8), parameter :: modal_aero_wateruptake_rmmax = 25.e-6_r8
 real(r8), parameter :: third = 1._r8/3._r8
 real(r8), parameter :: pi43  = pi*4.0_r8/3.0_r8
-
 
 ! Physics buffer indices
 integer :: cld_idx        = 0
@@ -133,7 +136,8 @@ subroutine modal_aero_wateruptake_init(pbuf2d)
 
    if (is_first_step()) then
       ! initialize fields in physics buffer
-      call pbuf_set_field(pbuf2d, dgnumwet_idx, 0.0_r8)
+      call pbuf_set_field(pbuf2d, dgnumwet_idx, modal_aero_wateruptake_rmmin)
+      call pbuf_set_field(pbuf2d, qaerwat_idx, 0._r8)
       if (modal_strat_sulfate) then
       ! initialize fields in physics buffer to NaN (not a number) 
       ! so model will crash if used before initialization
--- a/src/chemistry/utils/modal_aero_wateruptake.F90
+++ b/src/chemistry/utils/modal_aero_wateruptake.F90
@@ -22,19 +22,16 @@ private
 save
 
 public :: &
-   modal_aero_wateruptake_init,  &
-   modal_aero_wateruptake_dr,    &
-   modal_aero_wateruptake_sub,   &
-   modal_aero_kohler,            &
-   modal_aero_wateruptake_rmmin, &
-   modal_aero_wateruptake_rmmax
+   modal_aero_wateruptake_init, &
+   modal_aero_wateruptake_dr,   &
+   modal_aero_wateruptake_sub,  &
+   modal_aero_kohler
 
 public :: modal_aero_wateruptake_reg
 
-real(r8), parameter :: modal_aero_wateruptake_rmmin = 0.01e-6_r8
-real(r8), parameter :: modal_aero_wateruptake_rmmax = 25.e-6_r8
 real(r8), parameter :: third = 1._r8/3._r8
 real(r8), parameter :: pi43  = pi*4.0_r8/3.0_r8
+
 
 ! Physics buffer indices
 integer :: cld_idx        = 0
@@ -136,7 +133,7 @@ subroutine modal_aero_wateruptake_init(pbuf2d)
 
    if (is_first_step()) then
       ! initialize fields in physics buffer
-      call pbuf_set_field(pbuf2d, dgnumwet_idx, modal_aero_wateruptake_rmmin)
+      call pbuf_set_field(pbuf2d, dgnumwet_idx, 0.0_r8)
       call pbuf_set_field(pbuf2d, qaerwat_idx, 0._r8)
       if (modal_strat_sulfate) then
       ! initialize fields in physics buffer to NaN (not a number) 
--- a/src/physics/cam/modal_aer_opt.F90
+++ b/src/physics/cam/modal_aer_opt.F90
@@ -30,7 +30,7 @@ use cam_logfile,       only: iulog
 use perf_mod,          only: t_startf, t_stopf
 use cam_abortutils,    only: endrun
 
-use modal_aero_wateruptake, only: modal_aero_wateruptake_dr
+use modal_aero_wateruptake, only: modal_aero_wateruptake_dr, modal_aero_wateruptake_rmmin, modal_aero_wateruptake_rmmax
 use modal_aero_calcsize,    only: modal_aero_calcsize_diag
 
 implicit none
@@ -113,9 +113,9 @@ subroutine modal_aer_opt_init()
    ! Local variables
 
    integer  :: i, m
-   real(r8) :: rmmin, rmmax       ! min, max aerosol surface mode radius treated (m)
+
    character(len=256) :: locfile
-   
+
    logical           :: history_amwg            ! output the variables used by the AMWG diag package
    logical           :: history_aero_optics     ! output aerosol optics diagnostics
    logical           :: history_dust            ! output dust diagnostics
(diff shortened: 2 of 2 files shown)


fvitt:
@cacraigucar , I don't see how these mods change answers

src/chemistry/utils/modal_aero_wateruptake.F90
```diff
@@ -133,7 +136,8 @@ subroutine modal_aero_wateruptake_init(pbuf2d)
 
    if (is_first_step()) then
       ! initialize fields in physics buffer
-      call pbuf_set_field(pbuf2d, dgnumwet_idx, 0.0_r8)
+      call pbuf_set_field(pbuf2d, dgnumwet_idx, modal_aero_wateruptake_rmmin)
+      call pbuf_set_field(pbuf2d, qaerwat_idx, 0._r8)
```
fvitt, line 140: These mods must be what changes answers -- dgnumwet initialized differently.

cacraigucar:
In an email exchange from @adamrher where we were discussing this he said:

The aerosol changes are necessary because the new reordering does not allow wet deposition to run before radiation in the first time-step. and so I need to set these two wet dep vars to reasonable values at init. The current init values for these vars breaks cam_dev (one is not initialized at all; the other is initialized with a value that breaks MAM).

src/physics/cam/modal_aer_opt.F90
```diff
@@ -30,7 +30,7 @@ module modal_aer_opt
 use perf_mod,          only: t_startf, t_stopf
 use cam_abortutils,    only: endrun
 
-use modal_aero_wateruptake, only: modal_aero_wateruptake_dr
+use modal_aero_wateruptake, only: modal_aero_wateruptake_dr, modal_aero_wateruptake_rmmin, modal_aero_wateruptake_rmmax
```
fvitt, line 33: This line seems long.  Is it more than 128 characters?

fvitt (approved):

src/physics/cam/modal_aer_opt.F90
cacraigucar: It is 119 characters

src/chemistry/utils/modal_aero_wateruptake.F90
adamrher: Yep. And this is one area I wanted feedback on. This sets the radius of the aerosols for the first radiation time-step. I'd like the radiation in the first time-step to be as close as possible to default cam ordering runs. And so I set it to the minimum radius parameter used in modal_aero_opt. But perhaps the average of the min and max radius would be more realistic? I don't want the ESP folks to find that our initialized prediction runs are depreciated by a bad radiation field at the first time-step.

adamrher:
I'm wondering if we should use this PR to resolve this issue: https://github.com/ESCOMP/CAM/issues/461 as well?

cacraigucar:
> I'm wondering if we should use this PR to resolve this issue: #461 as well?

This hasn't been coded up, has it?  It is your call, but if we start adding things into this PR, it will slow down the process.  We can always open new PRs after this and #387 go in.  

Since you wanted this PR to go in before #387, I thought we were requesting it to be reviewed ASAP.  Reviewers don't start reviewing until the PR is stable (i.e. no more changes other than bug fixes).  We can tell folks to hold off if you want #461 to go into this PR.

adamrher:
fair point @cacraigucar. lets stay focused on the issue at hand, getting the re-ordering PR finished. sometimes my eyes are bigger than my stomach.

src/chemistry/utils/modal_aero_wateruptake.F90
fvitt: It might be better to initialize to the geometric dry mean diameter of each mode. If I interpret the code correctly, these are read from the phys prop files. See `dgnum_amode` in `modal_aero_data`.

adamrher: That seems like an improvement over my method. @andrewgettelman asks

> 'wet' might mean bigger, but we can start with the mean....if that doesn't work, maybe multiply it by 2 to assume some water uptake?

fvitt: Could the wet diameters be calculated during initialization?

cacraigucar: This will be addressed in #387

adamrher: well, I tried initializing the dgnumwet with the mean diameter of each mode. It might make more sense, but it doesn't improve the radiation field in the first time-step. Here are plots of FLDS at time-step 4 minus time-step 1. This shows how different the radiation changes from the first radiation call to the second radiation call:

Screen Shot 2021-12-10 at 5 22 39 PM

(Left) F2000climo, (MIddle) F2000dev, initializing dgnumwet=rmmin, and (Right) F2000dev initializing dgnumwet with the dry mode radius. The middle plot shows that for the reordering (F2000dev), the first radiation call doesn't have enough downward longwave radiation in the mid-to-high latitudes. The right plot shows that initializing with the dry mode radius doesn't relieve this issue. 

I did a run where I increase the dry mode radius by a factor of 10, and the changes to radiation in the first time step were negligible. The takeaway is not surprising, not calling wet dep before radiation in the first time-step can not be ameliorated by simply providing realistic values of dgnumwet. There are a plethora of optical properties other than dgnumwet that are set by wet dep, and using the initialized values instead of calling wet dep degrades the radiation field.

A solution to this problem is to call clubb->mg2->wetdep before radiation for nstep=0. But this would make the new tphysbc really messy. @andrewgettelman just to double check, do I have to call clubb+mg2 before wet dep, for the first time-step?

andrewgettelman: Hi Adam,

I'm not sure if you need to call MG2 before wet dep. You could try calling wet dep without MG2 and see if it works. I think we do some calculations for a conversion rate of cloud to precip, but if this were zero it should still run. I think that is it. It's worth a try.

adamrher: I was able to call wet dep w/o calling mg2 first, in the first time-step. tl;dr, it doesn't make a difference. I suspect ```qaerwat``` is still zero everywhere in that case, but I'll need to look into this some more. Do get a handle on how bad the radiation is in the first time-step, here are four runs showing FLDS for the first time-step.

Screen Shot 2021-12-12 at 11 45 01 AM

From left to right: (1) F2000climo, (2) F2000dev: method1, (3) F2000dev, method2, (4) F2000dev, method3.

method1 = initialize dgnumwet=rmmin, qaerwat=0
method2 = initialize dgnumwet=dry mode radius, qaerwat=0
method3 = don't initialize either, call wet dep before radiation in the first time-step.

For method3, wet dep basically intializes dgnumwet to the dry mode radius as in method2:
https://github.com/ESCOMP/CAM/blob/8497543088c385a7f44a394467c9ddd0a2819696/src/chemistry/utils/modal_aero_wateruptake.F90#L303-L305

And then method3 also sets aerowat=0 at the top of the qaerwat loop, for every time-step:
https://github.com/ESCOMP/CAM/blob/8497543088c385a7f44a394467c9ddd0a2819696/src/chemistry/utils/modal_aero_wateruptake.F90#L418

(this means that I was wrong for part (2) of issue: https://github.com/ESCOMP/CAM/issues/482. qaerwat is being initialized and there is no bug. The answer changes discussed in that issue are actually from dgnumwet, which @fvitt  notes at the top of this thread).

@cacraigucar since all three methods give the same result, I am leaning towards method3 since then we wouldn't need to touch modal_aero_wateruptake.F90 or modal_aer_opt.F90 at all. That would render this PR unnecessary. Method3 won't change answers for non cam_dev configurations since it only modifies cam_dev/physpkg.F90.

All three methods give the same solution, and this soln for FLDS at nstep=0 is obviously wrong when compared to the F2000climo solution. In fact, when I compared the clear sky radiation variables to the full radiation variables they are identical (as far as my eyes can tell) ... and indeed when I look at FSDS, it is clear that there are no clouds for all three methods:

Screen Shot 2021-12-12 at 11 54 43 AM

(there are clouds ... just that radiation doesn't see 'em). So I know you're probably thinking "that's not good..." And I would agree. But I want to make clear that for cam_dev, radiation is called *twice* for nstep=0, and the second call is a realistic radiation field. So while all of CESM will see a cloud free radiation for nstep=0, it will receive a realistic radiation for all other time-steps (this is not obvious since radiation isn't normally called again until nstep=3, but since we called it twice in nstep=0, CESM, incl all its surface components will see this second nstep=0 radiation call for nsteps=1, 2 ... I end up at this conclusion based on lots of diagnostics, and am certain I'm correct about this).

So where do we go from here? Do we hold off the PR until I can figure out how to massage the code to see clouds for the first radiation call in nstep=0? Or do we proceed with method3, and open an issue to remind us that this still needs to be resolved? I'm leaning towards the latter for practical reasons ...

@gold2718 @mvertens @JulioTBacmeister @andrewgettelman @swrneale

cacraigucar: @adamrher - It sounds like closing this PR without implementing is the correct way to go

cacraigucar:
After further investigation, @adamrher determined that the changes to the aerosol packages were not currently needed.  Close without implementing

PR #489, opened 2021-12-22
cam6_3_043: Read ncdata file with either ncol or ncol_d dimension (SE dycore)

### LINKED ISSUE #339: SE dycore not reading default ncdata for ne30np4 F2000
Running the ne30_ne30_mg17 grid with a F2000climo compset errors out:

` ERROR: check_file_layout: ERROR: either ncol or ncol_d dimension not found in ncdata file`

This same default ncdata is used for ne30pg2, ne30pg3 & ne30np4. This file has a dimension size ncol_d = 48602:
/glade/p/cesmdata/cseg/inputdata/atm/cam/inic/se/F2000climo_ne30pg3_mg17_01-01-00000_c200424.nc

This file works fine for ne30pg2 and ne30pg3, as ncol_d = 48602 is consistent with their definition of ncol_d. ne30np4 seems to not want ncol_d, but rather ncol = 48602, which is not present in this file.

ne30np4 runs fine if I instead point to this CAM-Chem inic, which has dimensions ncol = 48602:
/glade/p/cesmdata/cseg/inputdata/atm/cam/inic/se/f.e22.FC2010climo.ne30_ne30_mg17.cam6_2_022.001.cam.i.0016-01-01-00000_c200610.nc

@PeterHjortLauritzen

peverwhee:
Make ini_grid_hdim_name in dyn_grid public and use in dyn_comp so that GLL grids can use initial data files with ncol_d.

Closes #339

### DIFF
--- a/src/dynamics/se/dyn_comp.F90
+++ b/src/dynamics/se/dyn_comp.F90
@@ -2041,8 +2041,8 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok)
 
    ! Check that number of columns in IC file matches grid definition.
    if (trim(ini_grid_hdim_name) == 'none') then
-      call endrun(sub//': ERROR: no horizontal dimension in initial data file. ' &
-         '&Cannot read data from file')
+      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
+         &Cannot read data from file')
    end if
 
    ierr = pio_inq_dimid(file, trim(ini_grid_hdim_name), ncol_did)
--- a/src/dynamics/se/dyn_comp.F90
+++ b/src/dynamics/se/dyn_comp.F90
@@ -2041,8 +2041,8 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok)
 
    ! Check that number of columns in IC file matches grid definition.
    if (trim(ini_grid_hdim_name) == 'none') then
-      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
-         Cannot read data from file')
+      call endrun(sub//': ERROR: no horizontal dimension in initial data file. ' &
+         '&Cannot read data from file')
    end if
 
    ierr = pio_inq_dimid(file, trim(ini_grid_hdim_name), ncol_did)
--- a/src/dynamics/se/dyn_comp.F90
+++ b/src/dynamics/se/dyn_comp.F90
@@ -2040,10 +2040,14 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok)
    !----------------------------------------------------------------------------
 
    ! Check that number of columns in IC file matches grid definition.
+   if (trim(ini_grid_hdim_name) == 'none') then
+      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
+         Cannot read data from file')
+   end if
 
    ierr = pio_inq_dimid(file, trim(ini_grid_hdim_name), ncol_did)
    if (ierr /= PIO_NOERR) then
-      call endrun(sub//': ERROR: either ncol or ncol_d dimension not found in ' &
+      call endrun(sub//': ERROR: '//trim(ini_grid_hdim_name)//' dimension not found in ' &
          //trim(file_desc)//' file')
    end if
 
@@ -2053,7 +2057,7 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok)
          write(iulog, '(a,2(a,i0))') trim(sub), ': ncol_size=', ncol_size, &
              ' : dyn_cols=', dyn_cols
       end if
-      call endrun(sub//': ERROR: dimension ncol size not same as in ncdata file')
+      call endrun(sub//': ERROR: dimension '//trim(ini_grid_hdim_name)//' size not same as in ncdata file')
    end if
 
    ! Set coordinate name associated with ini_grid_hdim_name.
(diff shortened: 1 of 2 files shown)


gold2718 (approved):
Looks good!
Have you tested this with both an `ncol_d` / `area_d` file and an `ncol` / `area` file?

cacraigucar (approved):

peverwhee:
> Looks good! Have you tested this with both an `ncol_d` / `area_d` file and an `ncol` / `area` file?

yes - tested with both F2000climo_ne30pg3_mg17_01-01-00000_c200424.nc (ncol_d/area_d) and initial_data.cam.ne30.L32.RCEMIP_c20190507.nc (ncol/area) on izumi.

src/dynamics/se/dyn_comp.F90
```diff
@@ -2034,18 +2034,20 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok, dimname)
...
-   character(len=max_fieldname_len) :: dimname2, coordname
+   character(len=max_fieldname_len) :: coordname
 
    character(len=*), parameter      :: sub = 'check_file_layout'
    !----------------------------------------------------------------------------
 
    ! Check that number of columns in IC file matches grid definition.
+   if (trim(ini_grid_hdim_name) == 'none') then
+      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
+         Cannot read data from file')
```
fvitt, line 2045: Is this the correct syntax for continuing a string to the next line?

src/dynamics/se/dyn_grid.F90
```diff
@@ -62,6 +62,9 @@ module dyn_grid
 
 character(len=3), protected :: ini_grid_name
 
+! Name of horizontal grid dimension in initial file.
+character(len=6), protected :: ini_grid_hdim_name = ''
```
fvitt, line 66: I don't see where `ini_grid_hdim_name` gets set.

gold2718: It is set in the subroutine `get_hdim_name`.
Hmm, maybe we should use a different variable name instead of shadowing the module variable?

src/dynamics/se/dyn_comp.F90
peverwhee: good call - fixed.

gold2718: It is not illegal per se, however, the output is probably not what we want to say. How about:
```suggestion
      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
         &Cannot read data from file')
```

src/dynamics/se/dyn_grid.F90
peverwhee: ini_grid_hdim_name was already being set (unrelated to my changes) on line 224. I just made that variable public.

fvitt: OK thanks.  Maybe I needed more coffee to see it.

peverwhee: @gold2718 we can do that. Why would we want to avoid shadowing the module variable? Because it could be changed in some future update without realizing how it affects the other module?

gold2718: No technical reason, it is just by using `ini` in the variable names (`fh_ini`, `ini_grid_hdim_name`), it can give the impression that this routine is only for initial data files when it could, in principle, be used to examine any SE data file. If you do make the variable name changes, please update the comment ('the initial file') to express this more general usage possibility.

peverwhee: ok. Is it ok that the grid name is ini_grid_name? It seems like changing ini_grid_hdim_name creates a weird mismatch since we use ini_grid_name in dyn_comp as well as ini_grid_hdim_name.

gold2718: I was thinking `grid_hdim_name`.
Also, maybe `fh_ptr` instead of `fh_ini`.

peverwhee: I agree. I guess my question is whether the grid name in dyn_grid should be ini_grid_name or just grid_name ?

gold2718: I think the module variable names are all great, I am just looking at the dummy arguments in `get_hdim_name`. Does that address the question?

peverwhee: it sure does!

peverwhee: variable names changed.

fvitt (approved):

nusbaume (approved):
Looks good!  I just had one Fortran syntax question (which is probably better directed to @gold2718 ).

src/dynamics/se/dyn_comp.F90
```diff
@@ -2034,18 +2034,20 @@ subroutine check_file_layout(file, elem, dyn_cols, file_desc, dyn_ok, dimname)
...
-   character(len=max_fieldname_len) :: dimname2, coordname
+   character(len=max_fieldname_len) :: coordname
 
    character(len=*), parameter      :: sub = 'check_file_layout'
    !----------------------------------------------------------------------------
 
    ! Check that number of columns in IC file matches grid definition.
+   if (trim(ini_grid_hdim_name) == 'none') then
+      call endrun(sub//': ERROR: no horizontal dimension in initial data file. ' &
+         '&Cannot read data from file')
```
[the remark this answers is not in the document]
gold2718: It would be interesting to try but it is not what I recommended. My version is the Fortran 'break a line anywhere' get out of jail free card:
```suggestion
      call endrun(sub//': ERROR: no horizontal dimension in initial data file. &
         &Cannot read data from file')
```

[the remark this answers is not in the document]
peverwhee: fixed.

Issue #491, opened 2021-12-23
cam_dev is failing ERP test


cacraigucar:
Restarts are not BFB in cam_dev (compare_base_rest failing when running on izumi w/ nag). Run phase is failing for cam_dev (on cheyenne w/ intel and debug=T). Further exploration needs to be done to identify the cause.

swrneale:
Is this pre or post the ZM bug fixes?

cacraigucar:
It was in the code we committed with the ZM bugs.   I was hoping the ZM bug fixes would solve it, but it did not.

swrneale:
Still quite sure what you mean. Is it (other) code that went in 'with' the the zm bug fixes, or the zm code itself?

cacraigucar:
The bug is there before the ZM bug fixes (in PR #387) and remains after we use them (using the upcoming PR #490).  At this point we don't know where the restart errors are originating.  There is no reason to suspect the ZM code over any other parts of the code.  @adamrher said he has some ideas on where the issue might be.  I've not heard where he is going to be looking.

gold2718:
It seems there are at least two issue at play (looking at the PR). Can someone please write those down here (as comment or by editing the issue text)?

PR #493, opened 2022-01-03
part of cam6_3_041: mods to pass cam_dev ERP tests, misc cam_dev bug fixes

adamrher:
Fixes https://github.com/ESCOMP/CAM/issues/491

### DIFF
--- a/src/cpl/mct/atm_comp_mct.F90
+++ b/src/cpl/mct/atm_comp_mct.F90
@@ -358,7 +358,7 @@ CONTAINS
        end if
 
        ! Compute time of next radiation computation
-       nextsw_cday = radiation_nextsw_cday()
+       nextsw_cday = radiation_nextsw_cday(init=.true.)
        call seq_infodata_PutData( infodata, nextsw_cday=nextsw_cday )
 
        ! End redirection of share output to cam log
--- a/src/cpl/mct/atm_comp_mct.F90
+++ b/src/cpl/mct/atm_comp_mct.F90
@@ -36,6 +36,7 @@ module atm_comp_mct
   use phys_grid        , only: get_grid_dims
   use ppgrid           , only: pcols, begchunk, endchunk
   use camsrfexch       , only: cam_out_t, cam_in_t
+  use radiation        , only: nextsw_cday
   use cam_initfiles    , only: cam_initfiles_get_caseid, cam_initfiles_get_restdir
   use filenames        , only: interpret_filename_spec
   use spmd_utils       , only: spmdinit, masterproc, iam
@@ -86,7 +87,6 @@ CONTAINS
 
   subroutine atm_init_mct( EClock, cdata_a, x2a_a, a2x_a, NLFilename )
 
-    use radiation        , only: radiation_nextsw_cday
     !-----------------------------------------------------------------------
     !
     ! Arguments
@@ -124,7 +124,6 @@ CONTAINS
     logical           :: perpetual_run         ! If in perpetual mode or not
     integer           :: perpetual_ymd         ! Perpetual date (YYYYMMDD)
 
-    real(r8)          :: nextsw_cday           ! calendar of next atm shortwave
     integer           :: stepno                ! time step
     integer           :: dtime                 ! time step increment (sec)
     integer           :: start_ymd             ! Start date (YYYYMMDD)
@@ -320,8 +319,7 @@ CONTAINS
        ! This will only be used on the first timestep of an initial run
        !
        if (initial_run) then
-          nextsw_cday = get_curr_calday()
-          call seq_infodata_PutData( infodata, nextsw_cday=nextsw_cday )
+          call seq_infodata_PutData( infodata, nextsw_cday=get_curr_calday() )
        end if
 
        ! End redirection of share output to cam log
@@ -358,7 +356,6 @@ CONTAINS
        end if
 
        ! Compute time of next radiation computation
-       nextsw_cday = radiation_nextsw_cday(init=.true.)
        call seq_infodata_PutData( infodata, nextsw_cday=nextsw_cday )
 
        ! End redirection of share output to cam log
@@ -384,7 +381,6 @@ CONTAINS
 
  subroutine atm_run_mct( EClock, cdata_a, x2a_a, a2x_a)
 
-    use radiation        , only: nextsw_cday
     !-----------------------------------------------------------------------
     !
     ! Arguments
--- a/src/cpl/mct/atm_comp_mct.F90
+++ b/src/cpl/mct/atm_comp_mct.F90
@@ -319,7 +319,7 @@ CONTAINS
        ! This will only be used on the first timestep of an initial run
        !
        if (initial_run) then
-          call seq_infodata_PutData( infodata, nextsw_cday=get_curr_calday() )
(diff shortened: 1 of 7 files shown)


cacraigucar (approved):
Preliminary testing indicates this solves the restart issues.

nusbaume (approved):

src/physics/rrtmg/radiation.F90
```diff
@@ -326,11 +330,24 @@ real(r8) function radiation_nextsw_cday()
...
+   end if
+
    radiation_nextsw_cday = -1._r8
    dosw   = .false.
    nstep  = get_nstep()
    dtime  = get_step_size()
    offset = 0
+
+   ! wind back the clock to characterize radiation in restarts (for suites w/ radiation in tphysac)
+   if (doinit .and. is_first_restart_step() .and. cam_physpkg_is("cam_dev")) then
```
gold2718, line 346: This new logic seems fragile. Is it true that a better check would be testing _when_ radiation is called relative to the CAM time step? Sure, at the moment, `cam_dev` is the only package that is different but I would like the logic and documentation to be clear on this point.

gold2718 (commented):
Question about the logic.

fvitt (approved):

src/physics/rrtmg/radiation.F90
adamrher: as we discussed, let's put ```nextwsw_cday``` in the restarts. Should we then take it from the restarts for plain ol' cam physics as well, to make this logic independent of physics package?

w.r.t to the plain ol' cam physics package ... the restarts probably contains the state from dp_coupling or some state after CAM_run4, but before the final CAM_run1 call, right? But what about the radiation in the restarts, is it from the prior CAM_run1 call, or that final CAM_run1 call? I'm trying to understand if the radiation on restarts is the same as that produced during the CAM_run1 call during restart initialization.

gold2718: Yes, this should happen all the time.
Do you need any coding help?

adamrher: Yes. It would be helpful to know (1) what module has the appropriate list of vars where ```nextsw_cday``` should go to be written to restart, and (2) an example of reading a variable from restarts in the atm_comp_xxx module.

gold2718: I think all of this can (and should) go into radiation.F90 (rrtmg for sure and maybe also camrt).
At the module level (just above the `contains` statement):
```
type(var_desc_t) :: nextsw_cday_desc
```
In `radiation_define_restart`:
```
ierr = pio_def_var(File, 'nextsw_cday', pio_int, nextsw_cday_desc)
ierr = pio_put_att(File, nextsw_cday_desc, 'long_name', 'future radiation calday for surface models')
```
In `radiation_write_restart`:
```
ierr = pio_put_var(File, nextsw_cday_desc, (/ nextsw_cday /))
```
In `radiation_read_restart`:
```
ierr = pio_inq_varid(File, 'nextsw_cday', vardesc)
ierr = pio_get_var(File, vardesc, nextsw_cday)
```
Note: Add the `radiation_read_restart` code at the end of the routine. This will force a crash if `nextsw_cday` is not on the restart file.

peverwhee (approved):

src/physics/rrtmg/radiation.F90
adamrher: Or do I need to add ```nextsw_cday``` to ```read_restart_physics``` in read_restart_physics.F90?

adamrher: part 3 of Adam talks to himself. Steve's mods seem to have worked; I've past ERP tests for both F2000dev and F2000climo. I suspect Steve's mods work b/c for ```is_first_restart_step()``` the ```read_restart_physics``` subroutine is called, and this sets the module level value of ```nextsw_cday``` in radiation.F90.

Since I've not done my usual diligence of printing out nextsw_cday, I will wait to push these mods until I get verification from @gold2718 that my understanding is what he intended these mods to do.

gold2718: Great self-talk! Yes, the `pio_get_var` call in `radiation_read_restart` sets the module variable to the value it had when the restart file was written. I think that should set you back on track so that you do not need the other mods. That's the magic of restarts!

adamrher: having slept on this ... I think I can actually get rid of calling ```radiation_nextsw_cday``` in the init phase of mct/nuopc caps, and replace it with a ```use radiation        , only: nextsw_cday```. This will work because for startups, radiation_tend is called in initialization for all phys packages, which sets ```nextsw_cday``` correctly, and then for restart runs ```nextsw_cday``` will be made available by the ```radiation_read_restart``` call. 

I'm not so sure about simple models. I think it will work as long as simple models always calls its simplified radiation_tend in CAM_run1, which will set it to -1.

adamrher: This seems to have worked as it's passing a bunch of ERP tests (nuopc, mct, F2000climo, F2000dev, QPC4). There is an inefficiency in this approach for standard cam physics; the ```nextsw_cday``` read in from restarts will be overwritten by the value computed in CAM_run1 in the initialization phase. But since they are identical, it doesn't break anything.

src/cpl/mct/atm_comp_mct.F90
```diff
@@ -320,8 +319,7 @@ subroutine atm_init_mct( EClock, cdata_a, x2a_a, a2x_a, NLFilename )
        ! This will only be used on the first timestep of an initial run
        !
        if (initial_run) then
-          nextsw_cday = get_curr_calday()
-          call seq_infodata_PutData( infodata, nextsw_cday=nextsw_cday )
+          call seq_infodata_PutData( infodata, nextsw_cday=get_curr_calday() )
```
gold2718, line 322: How about initializing `nextsw_cday` in `radiation_init` so that you can use that here? That keeps radiation in charge of this quantity. In `radiation_init`, you would have something like:
```
if (.not. is_first_restart_step()) then
   nextsw_cday = get_curr_calday()
end if
```
Thoughts?

gold2718 (changes requested):
Not strictly necessary but I think a bit more cleanup and we have radiation always in charge of `nextsw_cday`.

src/physics/rrtmg/radiation.F90
```diff
@@ -314,6 +313,8 @@ end function radiation_do
 !================================================================================================
 
 real(r8) function radiation_nextsw_cday()
+
+   use phys_control,   only: cam_physpkg_is
```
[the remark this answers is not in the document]
adamrher: done

src/cpl/mct/atm_comp_mct.F90
adamrher: I'm on board. As you say, it's the last mod needed to make radiation entirely in charge of ```nextsw_cday```. I did this in camrt as well. I don't think it will be necessary after we fully transition to nuopc, but it nonetheless makes this variable explicitly defined for all time after init.

gold2718 (approved):
By George*, I think we've got it!

*No, I don't know who George is. George V maybe?

peverwhee (approved):

nusbaume (approved):

cacraigucar (approved):

PR #501, opened 2022-01-11
cam6_3_044: match cesm2_3_beta07 externals and bug fix for SE WACCM-X

### LINKED ISSUE #498: add value of zmconv_parcel_pbl in the namelist
I am setting a run with cam6_3_041. 

By default the value of zmconv_parcel_pbl is not output in the namelist. It is possible to go in the code and figure out the value that the model will be using. But if it was in the namelist, it would be less prone to be "forgotten". 

Would it be possible to add this to the namelist?

### LINKED ISSUE #499: Update externals to cesm2_3_beta07

### LINKED ISSUE #500: default pecount for ne120xx needs to be increased to run "out of the box"
The default pecount for ```ne120pg3_ne120pg3_mt13``` fails at run time (with unintelligible error messages in the logs). Increasing ```--pecount 1800``` does run (50 cheyenne nodes), and so I'd suggest making this the default for running on cheyenne. The current default on cheyenne for this grid is 576 tasks.

This presumably applies to all grids containing ne120xx.

cacraigucar:
Update externals to match those in cesm2_3_beta07  - The only exception is FMS which is newer than the version in CESM

bug fix for SE WACCM-X bug

From @gold2718 regarding WACCM-X bug:

- In the old days (pre physgrid),  the SE dycore always used the 'ncol' dimension.
- With physgrid, data on the GLL grid used 'ncol_d' and data on the physgrid used 'ncol'.
- When the switch was made to always read initial data on the GLL grid, that broke the logic in dyn_grid so Brian put in a fix (in late 2020).
- However, that fix prevented reading initial data on a physgrid run with the 'wrong' dimension name for the GLL data.
- cam6_3_043 included a fix for that problem, however, due to the change made in 2020, that caused the SE dycore to always write out new initial data files using the same dimension name as was used on input. For initial data runs not using physgrid (GLL-only), that meant 'ncol'.
- However, WACCM-X needs to also write data to the initial data file and always needs to use 'ncol' (as all physics code does).
- So, on a physgrid WACCM-X run that uses a GLL run input data file, the dycore and WACCM-X were both trying to write using 'ncol' but with two different sizes.
- The fix (we hope) is to have the dycore always write initial data files using 'ncol_d'.

More info from @gold2718:
A CAM history file has grid information (latitude, longitude, area, other grid parameters) for each grid represented in the set of 2D or 3D output variables.
For most dycores, this is the same for variables written from the dynamics or written from the physics.
However, for the SE dycore, there are potentially two grids, the native GLL grid and the finite volume physgrid.
We used to try and share grid attributes if they were the same (e.g., ne) and one of those is area when not using physgrid. In these cases, the dycore used 'ncol' and 'area' and the physics just used that as well.
However, now that we are always writing out the dynamics variables with ncol_d and area_d, we cannot share that with physics which always uses 'ncol'. 

So, the bug was, for the case of non-physgrid (e.g., ne5_ne5_mg37 or any refined mesh run), we were trying to write out area_d as an attribute but 'ncol_d' was not defined if there were no dynamics variables on the file (very common).

My fix (see PR) is to always have the physics write out 'area' with dimension 'ncol', even if the dynamics is also writing variables with the same grid. This will result in duplicated coordinates and area attributes but it is the best I can think of right now.

Closes #498
Closes #499 
Closes #500

### DIFF
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -990,7 +990,7 @@ if ($aer_model eq 'mam' ) {
     %modal_groups = ( 'accum'  => [qw(so4 pom soa bc dst ncl)],
                       'aitken' => [qw(so4 soa ncl dst)],
                       'coarse' => [qw(dst ncl so4)],
-                      'primary_carbon' => [qw(pom bc)], 
+                      'primary_carbon' => [qw(pom bc)],
                       'coarse_strat' => [qw(so4)] );
 
     @mode_num_src  = qw(A A A A A);
@@ -1853,7 +1853,7 @@ if (($chem =~ /trop_mozart/ or $chem =~ /trop_strat/ or $chem =~ /waccm_tsmlt/) 
                    'num_a1   -> ' => 'mam4_num_a1_emis_file',
                    'num_a2   -> ' => 'num_a2_emis_file',
                    'num_a4   -> ' => 'mam4_num_a4_emis_file',
-                   );								
+                   );
     } elsif ($chem =~ /mam4/) {
         %species = (%species,
                    'SOAG     -> ' => 'soag_emis_file',
@@ -2907,6 +2907,8 @@ if (($phys =~ /cam6/ or $phys =~ /cam_dev/) and $nl->get_value('deep_scheme') =~
 # cam_dev specific namelists
 if ($phys =~ /cam_dev/ and $nl->get_value('deep_scheme') =~ /ZM/) {
     add_default($nl, 'zmconv_parcel_pbl', 'val'=>'.true.');
+} else {
+    add_default($nl, 'zmconv_parcel_pbl', 'val'=>'.false.');
 }
 
 # Radiation scheme
@@ -4645,7 +4647,7 @@ sub check_snapshot_settings {
         if ($use_subcol_microp =~ /$TRUE/io) {
             push (@validList_bc, ("'microp_driver_tend_subcol'"));
         }
-        push (@validList_ac, ("'aero_model_wetdep'", 
+        push (@validList_ac, ("'aero_model_wetdep'",
                               "'radiation_tend'",
                               "'aoa_tracers_timestep_tend'",
                               "'co2_cycle_set_ptend'"));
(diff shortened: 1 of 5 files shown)


gold2718 (approved):
Looks good!

peverwhee (approved):

bld/build-namelist
```diff
@@ -2907,6 +2907,8 @@ if (($phys =~ /cam6/ or $phys =~ /cam_dev/) and $nl->get_value('deep_scheme') =~
 # cam_dev specific namelists
 if ($phys =~ /cam_dev/ and $nl->get_value('deep_scheme') =~ /ZM/) {
     add_default($nl, 'zmconv_parcel_pbl', 'val'=>'.true.');
+} else {
```
nusbaume, line 2910: Should there be a `deep_scheme` check here?  I imagine we wouldn't want to write this namelist value if ZM wasn't being used at all.

nusbaume (approved):
Looks good, but I did have one `build-namelist` question.

bld/build-namelist
gold2718: That makes sense but would also require some cleanup as that namelist is a mess at the moment. For instance, at least parts of that namelist are set in all cases except simple physics. Still, it is probably safe to change this to:
```suggestion
} elsif ($nl->get_value('deep_scheme') =~ /ZM/) {
```
or maybe:
```suggestion
} elsif (!$simple_phys) {
```
Note that the ZM namelist is _always_ read, even when running simpler models. Just more infrastructure crud we have to outgrow.

cacraigucar: If we are always reading the ZM namelist, having the value written out with a value of .false. is probably fine (as it says we are not doing the zmconv_parcel_pbl).  i.e. I am proposing we leave it as Steve initially made the mod.  Thoughts?

gold2718: Sure, it will not cause any trouble

nusbaume: If The ZM namelist is always read regardless of the model configuration then I am fine with leaving this as-is.  I was more concerned with trying to write a namelist variable that went completely un-used, but it doesn't sounds like that is the case here.

PR #510, opened 2022-01-25
part of cam_cesm2_1_rel55: add smyle L83 compset

jedwards4b:
Creates a compset BWsc1850smyle for the SMYLE forecast project.


ekluzek:
The main thing I see here is to coordinate this with the CTSM tag that does the L83 compset for CLM.  So that's https://github.com/ESCOMP/CTSM/pull/1614

ekluzek:
One question I have for the CAM folks is that the use-case for this sets a list of history settings. Overall we've avoided doing that in CTSM, and put history settings either in build-namelist or in user-mod directories rather than use-cases. I'm not sure that putting them in a use-case is a bad thing to do -- just that we've avoided doing that. So I wanted to hear if anyone knows if there are reasons why not to put them in use-cases? It does potentially have problems if certain configurations turn off some of the history fields that are in the list. But, that would be OK, if the SMYLE use-case isn't going to be used in that situation.

cime_config/config_component.xml
```diff
@@ -150,6 +150,10 @@
       <value compset="_CAM\d0%WC.*%SDYN">-nlev 88</value>
       <value compset="_CAM\d0%WX.*%SDYN">-nlev 145</value>
 
+      <!-- SMYLE -->
+      <value compset="_CAM60%WCSC_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD">-nlev 83</value>
```
cacraigucar, line 154: @fvitt - Is the way this is going in going to break other %WCSC runs (I'm not sure if the long name has an identifying feature that "normal" %WCSC runs do not have).  Since this is only slated to go into CESM2.1, it may not be as big a concern.

jedwards4b: This is the reason that I fully qualified the compset longname here, none of the other aliased WCSC compsets include the BGC modifier.

jedwards4b:
I have removed the history file changes to user_mods in jedwards4b/CESM2-Realtime-Forecast
which is used to generate and run ensembles for the smyle experiment.

gold2718:
> @jedwards4b: remove history file changes to another mechanism 

Where did it go?

cime_config/config_component.xml
cacraigucar: At this point, there are no compsets which use BGC, but there is nothing preventing a user from specifying a long name and getting this setup "by accident".  @fvitt - I do not know if this is a legitimate concern or not and will let you decide.  If it is, I would suggest implementing a %SMYLE qualifier.  @gold2718 - do you want to chime in?

cacraigucar (approved):

PR #514, opened 2022-01-27
part of cam_cesm2_1_rel55:  adding compset for CAM extending 2100-2500 for LE2

nanr:
LE2-2100extension:  This compset cycles most CAM forcing over year 2100. GHG evolves to 2500. Solar forcing has been extended by MikeMills from 2300 to 2500.  co2flux_fuel_file and aircraft_co2_file files were extended by KLindsay.

### DIFF
--- a/bld/namelist_files/use_cases/ssp370BBext_cam6.xml
+++ b/bld/namelist_files/use_cases/ssp370BBext_cam6.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0"?>
+
+<namelist_defaults>
+
+
+<!-- Emissions  -->
+<ext_frc_specifier>
+	 'H2O    -> $INPUTDATA_ROOT/atm/cam/chem/emis/elev/H2OemissionCH4oxidationx2_3D_L70_1849-2101_CMIP6_HistEnsAvg_SSP3-7.0.002_c20220120.nc',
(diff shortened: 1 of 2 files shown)


bld/namelist_files/use_cases/ssp370BBext_cam6.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0"?>
+
+<namelist_defaults>
+
+
+<!-- Emissions  -->
+<ext_frc_specifier>
+	 'H2O    -> $INPUTDATA_ROOT/atm/cam/chem/emis/elev/H2OemissionCH4oxidationx2_3D_L70_1849-2101_CMIP6_HistEnsAvg_SSP3-7.0.002_c20220120.nc',
```
cacraigucar, line 8: This data file does not reside in the svn repo, so users will be unable to be run on any machine other than where it currently resides (presumably cheyenne).  I don't know whether this file should be imported into the svn repo or not, but am pointing this out in case it is an issue.  I assume the other data files will also have this same issue

cacraigucar (changes requested):

bld/namelist_files/use_cases/ssp370BBext_cam6.xml
cacraigucar: @nanr - Can you please respond back on whether these data files should reside in the svn inputdata repository or not

nanr: Hi Cheryl - I have added these data to the SVN repo.  Thank you for pointing this out.  Nan

PR #518, opened 2022-01-28
Bugfix double call of outfld HNO3_GAS

### LINKED ISSUE #517: Double call of outfld for HNO3_GAS in mo_gas_phase_chemdr.F90
As soon as `has_strato_chem` is true, the following line is called twice in `src/chemistry/mozart/mo_gas_phase_chemdr.F90`:

`call outfld( 'HNO3_GAS',   vmr(:ncol,:,hno3_ndx), ncol, lchnk )`

This means that the sum of `HNO3_GAS`, `HNO3_STS` and `HNO3_NAT` is no more equal to `HNO3_TOTAL` in the output. Suggestion is to remove the second call of the above subroutine.

@fvitt

MiWeim:
These changes remove the double call of outfld for HNO3_GAS. In addition, right parantheses were added to some chenistry output field long names, that were missing.

Please note that the changes have not been tested yet.

Closes #517 

@fvitt

### DIFF
--- a/src/chemistry/mozart/mo_chm_diags.F90
+++ b/src/chemistry/mozart/mo_chm_diags.F90
@@ -332,7 +332,7 @@ contains
 
     call addfld( 'NOX',     (/ 'lev' /), 'A', 'mol/mol', 'nox (N+NO+NO2)' )
     call addfld( 'NOY',     (/ 'lev' /), 'A', 'mol/mol', &
-                 'noy = total nitrogen (N+NO+NO2+NO3+2N2O5+HNO3+HO2NO2+ORGNOY+NH4NO3' )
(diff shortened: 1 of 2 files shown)


fvitt:
Thanks for the corrections @MiWeim .  They look good to me.  I can include these changes in one of my future tags.  

@gold2718 @cacraigucar Should I include these in cam6_3_047 tag?  Or some later tag?

cacraigucar (approved):

cacraigucar:
> Thanks for the corrections @MiWeim . They look good to me. I can include these changes in one of my future tags.
> 
> @gold2718 @cacraigucar Should I include these in cam6_3_047 tag? Or some later tag?

I've asked for a quick review from the other SE's.  I would suggest you include it in your cam6_3_047 tag.

nusbaume (approved):

gold2718 (approved):

peverwhee (approved):

PR #527, opened 2022-02-04
add a hist compset for smyle L83

cacraigucar:
Opening PR based on email from @jedwards4b


bld/namelist_files/use_cases/waccm_sc_hist_smyle_L83.xml
```diff
@@ -0,0 +1,97 @@
...
+<!-- Reduce output size by removing some variables. -->
+<fexcl1>
+ 'BTAUN', 'BTAUS', 'BTAUE', 'BTAUW', 'BTAUNET', 'BUTEND1', 'BUTEND2', 'BUTEND3', 'BUTEND4', 'BUTEND5', 'BVTGWSPEC',
+ 'MAXQ0', 'HDEPTH', 'NETDT', 'TAUN', 'TAUS', 'TAUE', 'TAUW', 'TAUGWX', 'TAUGWY', 'UTEND1', 'UTEND2', 'UTEND3', 
+ 'UTEND4', 'UTEND5', 'FRONTGF', 'FRONTGFA', 'EKGW', 'QNO', 'QRLNLTE', 'QRL_TOT', 'DUV', 'DVV', 'TTPXMLC'
+</fexcl1>
+<!-- smyle specific changes -->
+<effgw_beres_dp>0.90D0</effgw_beres_dp>
+<gw_qbo_hdepth_scaling>0.25D0</gw_qbo_hdepth_scaling>
+<qbo_use_forcing>.false.</qbo_use_forcing>
```
fvitt, line 85: QBO options are specified above.  I think this line can be removed.

fvitt (approved):

bld/namelist_files/use_cases/waccm_sc_hist_smyle_L83.xml
jedwards4b: Actually I think that this line is intended to override the defaults, I can remove the QBO options above.

fvitt: You are turning off the QBO for all resolutions.  I am guessing this is only used on the f09 grid so this might be ok.  The GW settings you have are probably resolution dependent as well...

gold2718:
Do we really need to process a PR every time someone wants to do a simulation? This is getting expensive.

jedwards4b:
@gold2718 This not a single simulation - it's a project which will do thousands of runs both hindcasting from 1970 to present and forecasting with an ensemble of model members.   The project has already produced some 500 TB of data with the lowtop model and will eventually produce more than that with the L83 model.   There is the potential for a number of publications based on these runs and having it in a compset and tag is important.

gold2718:
Lots of people write papers without needing new CESM tags. We were told that the fire drill we did earlier in the week was what was needed for this project. Now, suddenly, without any open issue or discussion, there is another emergency PR. What happened? What sort of confidence should we have that this will not keep happening?

I have no way of evaluating this PR as there is no issue explaining what it is supposed to do, no explanation in this PR, and no ChangeLog entry to read.

swrneale:
I agree @gold2718. It would be nice if there was a plan for the proposed PRs and we can then judge them against what our individual priorities might be.

cacraigucar:
Closing without implementing -- postponed until a later date

Issue #495, opened 2022-01-07
Merging CARMA aerosol model for troposphere and stratosphere (trop_strat) into the CESM development version


tilmes:
CARMA is the Community Aerosol and Radiation Model for the Atmosphere, which is a framework for modeling both aerosols and clouds using a size resolving description. While there are already some CARMA models linked to CESM2, there is currently no version that completely replaces the modal aerosols in CAM with the sectional aerosol model in both the troposphere and the stratosphere. Currently, existing aerosol models in CESM2 run in parallel with the MAM scheme and do not interact fully with clouds, radiation, and chemistry.

The newly developed sectional aerosol model has the advantage to describe aerosol microphysics in much more detail than a modal aerosol model and is more physical because it fully integrates the sectional aerosols in physics, radiation, and chemistry. The implementation of a model is in particular important for exploring the effects of stratospheric aerosols injection (geoengineering), but also to understand the effects of volcanoes. The coupling to the entire climate system is therefore needed. In addition, for the troposphere, this model has been shown to be able to resolve tropospheric aerosols well. A presentation to the CGD process group on the topic was given in May 2021. 
[CGD_process_meeting_May2021.pptx](https://github.com/ESCOMP/CAM/files/7831331/CGD_process_meeting_May2021.pptx)

The development has been performed in an earlier model version (close to CESM2.2.0) and needs to be merged into the development code.  The implementation includes changes and additions of several routines that are listed in this google doc: https://docs.google.com/document/d/15if8UXjDcyhduCjOCVOgU5mHsoXIIepHI0CPzuj5k2g/edit
Details of the different changes in the different routines can be provided.

fvitt:
See branch:
https://github.com/fvitt/CAM/tree/carma_model_dev

PR #546, opened 2022-03-21
cam6_3_077: Refactor ndrop to use generalized aerosol interfaces

fvitt:
This modifies ndrop (cloud-aerosol interface) code to use generalized (or abstract) cloud interfaces.  Introduced are abstract aerosol_properties and aerosol_state classes which are extended for concrete modal aerosol classes.  This is the 1st step (of many) to address issue #495 (integrating CARMA trop_strat model into CAM physics).


src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,217 @@
+module aerosol_state_mod
+  use shr_kind_mod, only: r8 => shr_kind_r8
+  use cam_abortutils, only: endrun
+  use cam_logfile, only: iulog
```
gold2718, line 4: I don't think that an abstract interface should be using model-specific interfaces. Perhaps take in an input for a log unit and return an error code instead of calling `endrun`?

gold2718 (changes requested):
I have only looked at the abstract interfaces so far. However, I think it is important to clarify this new interface before diving further into the code.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+  ! ojbect initializer
+  !------------------------------------------------------------------------------
+  subroutine aero_props_init(self, nbin, ncnst, nspec, nmasses, amcubecoefs, alogsig, f1,f2 )
+    class(aerosol_properties), intent(inout) :: self
+    integer, intent(in) :: nbin               ! number of bins
+    integer, intent(in) :: ncnst              ! total number of constituents
+    integer, intent(in) :: nspec(nbin)        ! number of species in each bin
+    integer, intent(in) :: nmasses(nbin)      ! number of masses in each bin
+    real(r8),intent(in) :: amcubecoefs(nbin)  ! coefficient used to calc the radius^3 of each bin
+    real(r8),intent(in) :: alogsig(nbin)      ! ??? some log factor of each bin ???
```
tilmes, line 107: log of the standard deviation (sigma) of the aerosol mode

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+     !  density
+     !  hygroscopicity
+     !------------------------------------------------------------------------
+     subroutine aero_props_get(self, bin_ndx, species_ndx, density,hygro)
+       import
+       class(aerosol_properties), intent(in) :: self
+       integer, intent(in) :: bin_ndx             ! bin index
+       integer, intent(in) :: species_ndx         ! species index
+       real(r8), optional, intent(out) :: density ! density (kg/m3)
+       real(r8), optional, intent(out) :: hygro   ! hygroscopicity
```
fvitt, line 55: @bardeenc @drmikemills 
Does hygroscopicity have units?

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+     private
+     integer :: nbins_ = 0
+     integer :: ncnst_tot_ = 0
+     integer, allocatable :: nmasses_(:)
+     integer, allocatable :: nspecies_(:)
+     integer, allocatable :: indexer_(:,:)
+     real(r8), allocatable :: amcubecoefs_(:)
+     real(r8), allocatable :: alogsig_(:)
+     real(r8), allocatable :: f1_(:)
+     real(r8), allocatable :: f2_(:)
```
fvitt, line 20: Should maybe document the data members.

src/physics/cam/ndrop.F90
```diff
@@ -1823,148 +1574,34 @@ subroutine ccncalc(state, pbuf, cs, ccn)
...
-            state, pbuf, 1, ncol, k, &
+         call aero_state%loadaer( aero_props, &
+            1, ncol, k, &
             m, cs, phase, naerosol, vaerosol, &
             hygro)
 
-         where(naerosol(:ncol)>1.e-3_r8)
-            amcube(:ncol)=amcubecoef(m)*vaerosol(:ncol)/naerosol(:ncol)
+         where(naerosol(:ncol)>1.e-3_r8 .and. hygro(:ncol).gt.1.e-10_r8)
+            amcube(:ncol)=aero_props%amcubecoef(m)*vaerosol(:ncol)/naerosol(:ncol)
```
fvitt, line 1587: maybe just use aero_props%amcube function directly

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: Please note some of these variables have already been defined further down in the code.
!! The abstract aerosol interface allows aerosol representations that define a specific number of bins/modes (nbins). Each of the bins/modes consists of a specific number of species (nspecies) that are assumed to be internally mixed. Therefore each bin/number has a total aerosol number and total mass. Specific aerosol descriptions may not require to define/present total mass. The code allows either option.

nbin !! number of bins or modes   
ncnst_tot !! total number of bin/mode number, total mass, and number of species
ncnst_extd !! twice the total number of bin/mode number, bin total mass, and concentration species (not yet included)
nmasses !!  number of species per bin/mode plus 1 to consider the total mass (if available)
indexer !!  table for local indexing of bin/mode aero number and mmr. This table uses species index 0 for the number conc.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+
+    allocate(self%nspecies_(nbin))
+    allocate(self%nmasses_(nbin))
+    allocate(self%amcubecoefs_(nbin))
+    allocate(self%alogsig_(nbin))
+    allocate(self%f1_(nbin))
+    allocate(self%f2_(nbin))
+
+    allocate( self%indexer_(nbin,0:maxval(nmasses)) )
+
```
tilmes, line 121: Add comment:
! Local indexing compresses the mode and number/mass indices into one index.
   ! This indexing is used by the pointer arrays used to reference state and pbuf
   ! fields. We add number = 0, total mass = 1 (if available), and mass from each constituency into mm.

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: amcubecoefs may be removed here and just used in the rcubed (amcube) routine.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,306 @@
...
+     generic :: nmasses => nmassesa,nmassesv
+     procedure :: indexer
+     procedure :: maxsat
+     procedure :: amcubecoef
+     procedure :: amcube
+     procedure :: alogsig
+     procedure(aero_props_get), deferred :: get
+     procedure(aero_actfracs), deferred :: actfracs
+     procedure(aero_num_names), deferred :: get_num_names
+     procedure(aero_mmr_names), deferred :: get_mmr_names
```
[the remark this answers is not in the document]
tilmes: I added some summary above to address this.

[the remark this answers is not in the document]
tilmes: nmassesa and nmassesv are not intuitive names to me. In the code it says:
nmassesa. ! number of species masses for all bins.    -> However, I do not see it being used anywhere?
nmassesv. ! number of species masses in a given bin number. -> again I don't see it used in the code at the moment. Will it be used in a different code (that is not added yet)?

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: Definitions here or on the top of the program or both? Should be consistent.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+  ! returns number of constituents (or elements) totaled across all bins
+  !------------------------------------------------------------------------------
+  pure integer function ncnst_tot(self)
+    class(aerosol_properties), intent(in) :: self
+
+    ncnst_tot = self%ncnst_tot_
+  end function ncnst_tot
+
+  !------------------------------------------------------------------------------
+  ! returns coefficient factor for radius^3 calc
```
tilmes, line 237: Instead of this function, I think, we can just have 1 function that returns rcubed and include the coefficient there.

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,213 @@
...
+    integer,allocatable :: nspecies(:)
+    integer,allocatable :: nmasses(:)
+    real(r8),allocatable :: amcubecoefs(:)
+    real(r8),allocatable :: alogsig(:)
+    real(r8),allocatable :: f1(:)
+    real(r8),allocatable :: f2(:)
+
+    allocate(newobj)
+
+    call rad_cnst_get_info(0, nmodes=nmodes)
```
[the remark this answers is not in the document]
tilmes: @gold2718  Can we assume for now that it works with rad_cnst?

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,215 @@
...
+  ! amcube is overridden to keep MAM b4b
+  !------------------------------------------------------------------------------
+  pure real(r8) function amcube(self, bin_ndx, volconc, numconc)
+
+    class(modal_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx
+    real(r8), intent(in) :: volconc
+    real(r8), intent(in) :: numconc
+
+    amcube = (3._r8*volconc/(4._r8*pi*self%exp45logsig_(bin_ndx)*numconc))
```
tilmes, line 168: Are we calculating amcube here or in aerosol_properties_mod.F90?  I am thinking, if you define the coefficient for the calculation, you can do it in aerosol_properties_mod, if you do it here, you want to also do the same for CARMA (not in here) and you could remove the coefficient.

src/physics/cam/microp_aero.F90
```diff
@@ -477,19 +498,14 @@ subroutine microp_aero_run ( &
...
+
       call pbuf_get_field(pbuf, ast_idx,  cldn, start=(/1,1,itim_old/), kount=(/pcols,pver,1/) )
       call pbuf_get_field(pbuf, cldo_idx, cldo, start=(/1,1,itim_old/), kount=(/pcols,pver,1/) )
-
-      call rad_cnst_get_info(0, nmodes=nmodes)
-      call pbuf_get_field(pbuf, dgnumwet_idx, dgnumwet, start=(/1,1,1/), kount=(/pcols,pver,nmodes/) )
-
-      allocate(factnum(pcols,pver,nmodes))
-
+      call pbuf_get_field(pbuf, dgnumwet_idx, dgnumwet)
```
tilmes, line 504: This is aerosol model specific (MAM) and it does not seem to be needed?

tilmes: line 686ff: this is also aerosol specific

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+  !------------------------------------------------------------------------------
+  pure real(r8) function amcubecoef(self, bin_ndx)
+    class(aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+
+    amcubecoef = self%amcubecoefs_(bin_ndx)
+  end function amcubecoef
```
mattldawson, line 244: Is `amcubecoef()` ever used other than for calculating r^3 for a given number and volume? If not, consider merging this function with `amcube()` and possibly making it `deferred` so that each aerosol model can calculate r^3 based on its representation

tilmes: I was thinking the same, but the r^3 calculation needs volume and number and that would have to be passed to the r^3 calculation, maybe doing it like Francis suggesting is actually not a bad idea?

src/chemistry/aerosol/aerosol_properties_mod.F90
mattldawson: Would it make sense to have this be two functions `aero_props_get_density` and `aero_props_get_hygroscopicity`? Both outputs are already optional, and the separate functions' names would be more clear.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+     generic :: nmasses => nmassesa,nmassesv
+     procedure :: indexer
+     procedure :: maxsat
+     procedure :: amcubecoef
+     procedure :: amcube
+     procedure :: alogsig
+     procedure(aero_props_get), deferred :: get
+     procedure(aero_actfracs), deferred :: actfracs
+     procedure(aero_num_names), deferred :: get_num_names
+     procedure(aero_mmr_names), deferred :: get_mmr_names
```
mattldawson, line 37: I would suggest consistently using (or not) the `get_` prefix here:
```
     (...)
     procedure :: get_amcube
     procedure :: get_alogsig
     procedure(aero_props), deferred :: get_props
     procedure(aero_act_fracs), deferred :: get_act_fracs
     procedure(aero_num_names), deferred :: get_num_names
     procedure(aero_mmr_names), deferred :: get_mmr_names
```
or
```
     (...)
     procedure :: amcube
     procedure :: alogsig
     procedure(aero_props), deferred :: props
     procedure(aero_act_fracs), deferred :: act_fracs
     procedure(aero_num_names), deferred :: num_names
     procedure(aero_mmr_names), deferred :: mmr_names
```
One argument for the latter is that it's shorter and you're already using a trailing underscore for data member names, which will prevent naming conflicts with accessor functions.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+       integer, intent(in) :: bin_ndx             ! bin index
+       integer, intent(in) :: species_ndx         ! species index
+       real(r8), optional, intent(out) :: density ! density (kg/m3)
+       real(r8), optional, intent(out) :: hygro   ! hygroscopicity
+     end subroutine aero_props_get
+
+     !------------------------------------------------------------------------
+     ! returns mass and number activation fractions
+     !------------------------------------------------------------------------
+     subroutine aero_actfracs(self, bin_ndx, smc, smax, fn, fm )
```
mattldawson, line 61: I think we discussed leaving the original names of functions and their arguments for now to help people familiar with the old names navigate the refactoring, but proposing new names in the comments. Here is one:
```
! proposed new name: aero_activation_fractions
```

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+     !------------------------------------------------------------------------
+     !------------------------------------------------------------------------
+     subroutine aero_mmr_names(self, bin_ndx, species_ndx, name_a, name_c)
+       import
+       class(aerosol_properties), intent(in) :: self
+       integer, intent(in) :: bin_ndx           ! bin number
+       integer, intent(in) :: species_ndx       ! species number
+       character(len=32), intent(out) :: name_a ! constituent name of ambient aerosol MMR
+       character(len=32), intent(out) :: name_c ! constituent name of cloud-borne aerosol MMR
+     end subroutine aero_mmr_names
```
mattldawson, line 91: Is it likely that these will be unneeded once the CAM dependencies are removed from the aerosol code? If not, some description of what these names are used for could be included function descriptions

src/chemistry/aerosol/aerosol_state_mod.F90
mattldawson: @gold2718 - I don't disagree with your proposal, but I thought we decided that in the first wave of refactoring the CAM dependencies would be left in place, and that these would be addressed in the second wave of refactoring. Is this correct?

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
gold2718: The whole point of the new interfaces is to provide a stable interface to CAM which allows the underlying implementation to evolve and change without affecting the CAM implementation. This code is grabbing data from inside of CAM rather than through its interface which makes it non-modular (i.e., it does not meet the purpose of the new interfaces).

src/chemistry/aerosol/aerosol_state_mod.F90
gold2718: I'm not sure. My understanding is that we would continue to use elements of the CAM implementation for now but present interfaces that would not need to change when the implementation was made independent of CAM. In particular, this one seems like an easy add and one that is necessary for modular code.

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,218 @@
...
+  use aerosol_properties_mod, only: aerosol_properties
+
+  implicit none
+
+  private
+
+  public :: aerosol_state
+  public :: ptr2d_t
+
+  type, abstract :: aerosol_state
```
mattldawson, line 14: I think it would be helpful to include descriptions of `aerosol_state` and `aerosol_propoerties` that describe the conceptual differences between the two so that someone adding a new aerosol model can make the right choices about where to include specific data members

src/chemistry/aerosol/aerosol_properties_mod.F90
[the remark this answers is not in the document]
fvitt: @tilmes 
Search for `aero_props%nmasses` in `ndrop.F90`.  The `nmasses` procedure is overloaded (see the `generic` keyword) so that it can return either a single value for a given bin number or an array of values for all bins.

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
fvitt: I think @gold2718 has his sights set on the end goal of having aerosol object code that is completely independent of CAM which is where we all want to end up.  Another goal is to develop an interface that works for both MAM and CARMA so that we can easily incorporate CARMA into the CAM physics parameterizations.  For the time being, we are using the rad_cnst interfaces to get information and state of the aerosols.  It may take a while to develop aerosol classes that are completely independent of CAM.

src/chemistry/aerosol/aerosol_state_mod.F90
fvitt: I agree that `endrun` and `iulog` could be easily replaced at this point.  The `rad_cnst_get` interfaces are more difficult.

src/physics/cam/ndrop.F90
fvitt: This change causes the MAM baseline tests for fail due the roundoff differences in the CCN diagnostics.  This makes it a little tricky to verify that I have not changed MAM results.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: I agree with these suggestions.  Note, these changes to amcube do cause roundoff differences in the CCN diagnostics of MAM.

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
gold2718: I think I have misunderstood some of this. @fvitt, is it true that the `rad_cnst_get_info` uses only constituent information and nothing from the CAM state? If so, then this is a valid use. My real concern here is management of state and tendency information between the host model and the aerosol code.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
+module aerosol_properties_mod
+  use shr_kind_mod, only: r8 => shr_kind_r8
+
+  implicit none
+
+  private
+
+  public :: aerosol_properties
+
+  type, abstract :: aerosol_properties
```
mattldawson, line 10: please add comment:
```
!> aerosol_properties defines the configuration of any aerosol package (using any aerosol representation)
!! based on user specification. These values are set during initialization and do not vary during the simulation.
!!
!! Each aerosol package (e.g., MAM, CARMA, etc) must extend the aerosol_properties class to define the
!! details of their configuration. Any package must implement each of the deferred procedures of the abstract
!! aerosol_properties class, may include additional private data members and type-bound procedures, and may
!! override functions of the abstract class.
!!
!! Please see the modal_aerosol_properties module for an example of how the aerosol_properties class can
!! be extended for a specific aerosol package.
```

src/chemistry/aerosol/aerosol_state_mod.F90
mattldawson: please add comment:
```
!> aerosol_state defines the interface to the time-varying aerosol state variables (e.g., mixing ratios, number
!! concentrations). This includes the aerosol portion of the overall model state.
!!
!! Each aerosol package (e.g., MAM, CARMA, etc) must extend the aerosol_state class to allow access to
!! the state information (transported and not transported) of the aerosol package.
!! Any package must implement each of the deferred procedures of the abstract
!! aerosol_state class, may include additional private data members and type-bound procedures, and may
!! override functions of the abstract class.
!!
!! Please see the modal_aerosol_state module for an example of how the aerosol_state class can
!! be extended for a specific aerosol package.
```

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+    class(aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+
+    nspecies = self%nspecies_(bin_ndx)
+  end function nspecies
+
+  !------------------------------------------------------------------------------
+  ! returns number of species masses in a given bin number
+  !------------------------------------------------------------------------------
+  pure function nmassesv(self,bin_ndx) result(val)
```
mattldawson, line 189: suggest naming function: `n_masses_per_bin`

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+    integer, intent(in) :: bin_ndx           ! bin number
+    integer :: val
+
+    val = self%nmasses_(bin_ndx)
+  end function nmassesv
+
+  !------------------------------------------------------------------------------
+  ! returns an array of number of species masses for all bins
+  !------------------------------------------------------------------------------
+  pure function nmassesa(self) result(arr)
```
mattldawson, line 200: suggest naming function: `n_masses_all_bins`

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,328 @@
...
+    class(aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+
+    alogsig = self%alogsig_(bin_ndx)
+  end function alogsig
+
+  !------------------------------------------------------------------------------
+  ! returns radius^3 (m3)
+  !------------------------------------------------------------------------------
+  pure real(r8) function amcube(self, bin_ndx, volconc, numconc)
```
mattldawson, line 259: Consider moving this to `aerosol_state`

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: I have made the suggested change here

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Added these comments.

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
fvitt: I removed the coefficients

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: done

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: I wen with the 2nd option here

src/chemistry/aerosol/aerosol_state_mod.F90
fvitt: Added this

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Added

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: done

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: done

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: hygro does not have units

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: Good with me

src/chemistry/aerosol/aerosol_properties_mod.F90
tilmes: looks good to me

cacraigucar (commented):
My first round of questions/comments

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+  type, abstract :: aerosol_properties
+     private
+     integer :: nbins_ = 0
+     integer :: ncnst_tot_ = 0
+     integer, allocatable :: nmasses_(:)
+     integer, allocatable :: nspecies_(:)
+     integer, allocatable :: indexer_(:,:)
+     real(r8), allocatable :: alogsig_(:)
+     real(r8), allocatable :: f1_(:)
+     real(r8), allocatable :: f2_(:)
```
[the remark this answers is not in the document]
mattldawson: Hi @cacraigucar - I would say yes on both counts. I'm not sure about Fortran, but trailing underscores I think are fairly commonly used in other languages for data members (e.g., [here](https://google.github.io/styleguide/cppguide.html#Variable_Names)) and to prevent naming conflicts (e.g., [here](https://pep8.org/#naming-conventions)). I've found trailing underscores for private data members useful in Fortran for the reason you mentioned - you can have accessor functions with the data member name w/o the underscore (but I haven't seen anyone else specifically recommend it for Fortran).

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -90,35 +100,51 @@ subroutine aero_mmr_names(self, bin_ndx, species_ndx, name_a, name_c)
...
+     ! returns radius^3 (m3) of a given bin number
+     !------------------------------------------------------------------------------
+     pure elemental real(r8) function aero_amcube(self, bin_ndx, volconc, numconc)
+       import
+
+       class(aerosol_properties), intent(in) :: self
+       integer, intent(in) :: bin_ndx  ! bin number
+       real(r8), intent(in) :: volconc ! volume conc (m3/m3)
+       real(r8), intent(in) :: numconc ! number conc (1/m3)
+
```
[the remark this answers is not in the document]
fvitt: This just defines the interface.  The extending concrete classes must implement the method.  See the implementation of `amcube` in `modal_aerosol_properties_mod.F90`.

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,261 @@
...
+
+    select type(aero_props)
+    type is (modal_aerosol_properties)
+
+       ! adjust number so that dgnumlo < dgnum < dgnumhi not done for bins
+       do i = istart, istop
+          naerosol(i) = max(naerosol(i), vaerosol(i)*aero_props%voltonumbhi(m))
+          naerosol(i) = min(naerosol(i), vaerosol(i)*aero_props%voltonumblo(m))
+       end do
+    end select
```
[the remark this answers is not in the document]
fvitt: This is a common routine used by both MAM and CARMA.  This particular block of code is executed only for MAM.

src/physics/cam/physpkg.F90
```diff
@@ -1312,6 +1314,7 @@ subroutine phys_final( phys_state, phys_tend, pbuf2d )
     call chem_final
     call carma_final
     call wv_sat_final
+    call microp_aero_final()
```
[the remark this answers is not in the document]
fvitt: @cacraigucar : I can add the call to `cam_dev/physpkg`.  What are the regression tests that exercise `cam_dev/physpkg`?

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,261 @@
...
+          do i = istart, istop
+             vol(i) = max(qqcw(i,k), 0._r8)/specdens
+          end do
+       else if (phase == 1) then
+          do i = istart, istop
+             vol(i) = max(raer(i,k), 0._r8)/specdens
+          end do
+       else
+          write(iulog,*)'phase = ',phase,' in loadaer not recognized'
+          call endrun('phase error in loadaer')
```
cacraigucar, line 172: Would it be useful to have this say loadaer1 instead of the generic loadaer?

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+       integer, intent(in) :: species_ndx       ! species number
+       character(len=32), intent(out) :: name_a ! constituent name of ambient aerosol MMR
+       character(len=32), intent(out) :: name_c ! constituent name of cloud-borne aerosol MMR
+     end subroutine aero_mmr_names
+
+     !------------------------------------------------------------------------------
+     ! returns radius^3 (m3) of a given bin number
+     !------------------------------------------------------------------------------
+     pure elemental real(r8) function aero_amcube(self, bin_ndx, volconc, numconc)
+       import
```
cacraigucar, line 107: I would suggest that `import` follow the same conventions as `use` and the fields being imported be explicitly listed

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+     procedure :: ncnst_tot
+     procedure :: nspecies
+     procedure,private :: n_masses_all_bins
+     procedure,private :: n_masses_per_bin
+     generic :: nmasses => n_masses_all_bins,n_masses_per_bin
+     procedure :: indexer
+     procedure :: maxsat
+     procedure(aero_amcube), deferred :: amcube
+     procedure :: alogsig
+     procedure(aero_props_get), deferred :: get
```
cacraigucar, line 44: `get` seems a little simplified to me for a name - it made me ask "get what"?  Should it be `get_props` at the minimum?

cacraigucar (commented):
Next round of comments/questions

src/physics/cam/microp_aero.F90
```diff
@@ -635,19 +651,28 @@ subroutine microp_aero_run ( &
 
       call outfld('LCLOUD', lcldn, pcols, lchnk)
 
+      ! create the aerosol state object
+      aero_state_obj => modal_aerosol_state( state1, pbuf, aero_props_obj )
+
+      allocate(factnum(pcols,pver,aero_props_obj%nbins()))
```
peverwhee, line 657: add stat check?

peverwhee (approved):
a few suggestions/comments

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Do you prefer `import :: aerosol_properties, r8`?  I would like to see the opinions of others on this.

src/chemistry/aerosol/aerosol_state_mod.F90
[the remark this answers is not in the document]
fvitt: Perhaps this code block could be replaced with a call to a limiter function that can be overridden in the modal_aerosol_properties class.

src/chemistry/aerosol/aerosol_properties_mod.F90
gold2718: Yes, it is always valuable to know what is being imported.

src/physics/cam/ndrop.F90
```diff
@@ -1823,148 +1574,34 @@ subroutine ccncalc(state, pbuf, cs, ccn)
...
-         call loadaer( &
-            state, pbuf, 1, ncol, k, &
+         call aero_state%loadaer( aero_props, &
+            1, ncol, k, &
             m, cs, phase, naerosol, vaerosol, &
             hygro)
 
-         where(naerosol(:ncol)>1.e-3_r8)
-            amcube(:ncol)=amcubecoef(m)*vaerosol(:ncol)/naerosol(:ncol)
+         where(naerosol(:ncol)>1.e-3_r8 .and. hygro(:ncol).gt.1.e-10_r8)
```
gold2718, line 1586: Weird mix of old style and new (with the new code using the old syntax). Can you use `>` for both terms?

src/physics/cam/ndrop.F90
```diff
@@ -1823,148 +1574,34 @@ subroutine ccncalc(state, pbuf, cs, ccn)
...
             sm(:ncol)=smcoef(:ncol)/sqrt(hygro(:ncol)*amcube(:ncol)) ! critical supersaturation
          elsewhere
             sm(:ncol)=1._r8 ! value shouldn't matter much since naerosol is small
          endwhere
          do l=1,psat
             do i=1,ncol
-               arg(i)=argfactor(m)*log(sm(i)/super(l))
-               ccn(i,k,l)=ccn(i,k,l)+naerosol(i)*0.5_r8*(1._r8-erf(arg(i)))
+               arg(i)=argfactor(m)*log(sm(i)/super(l))                      ! staturation factors are a little diff         <--| ???
+               ccn(i,k,l)=ccn(i,k,l)+naerosol(i)*0.5_r8*(1._r8-erf(arg(i))) ! looks like activation frac fn (arg is a little diff) ??
```
gold2718, line 1595: These comments do not give me the warm fuzzies (especially the comment on the comment). Can you clarify a bit? What is "a little diff"?

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,261 @@
...
+  !! data members and type-bound procedures, and may override functions of the
+  !! abstract class.
+  !!
+  !! Please see the modal_aerosol_state module for an example of how the aerosol_state
+  !! class can be extended for a specific aerosol package.
+  type, abstract :: aerosol_state
+   contains
+     procedure :: number_transported
+     procedure :: get_transported
+     procedure :: set_transported
```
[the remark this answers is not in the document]
gold2718: These calls need to be put into the code so that it would run correctly if the objects really had a copy of the transported data (instead of a pointer to the CAM state).

gold2718 (changes requested):
A few things to change and a few older questions that have not yet been answered.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: done

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,201 @@
...
+  end function constructor
+
+  !------------------------------------------------------------------------------
+  !------------------------------------------------------------------------------
+  subroutine destructor(self)
+    type(modal_aerosol_properties), intent(inout) :: self
+
+    deallocate(self%exp45logsig_)
+    deallocate(self%voltonumblo_)
+    deallocate(self%voltonumbhi_)
```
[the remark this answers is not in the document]
fvitt: Added checks

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,201 @@
...
+    allocate(newobj%voltonumblo_(nmodes))
+    allocate(newobj%voltonumbhi_(nmodes))
+
+    ncnst_tot = 0
+
+    do m = 1, nmodes
+       call rad_cnst_get_info(0, m, nspec=nspecies(m))
+
+       ncnst_tot =  ncnst_tot + nspecies(m) + 1
+       nmasses(m) = nspecies(m)
```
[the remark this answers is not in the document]
fvitt: Did away with the redundant nmasses here.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+    self%nmasses_(:) = nmasses(:)
+    self%nspecies_(:) = nspec(:)
+    self%alogsig_(:) = alogsig(:)
+    self%f1_(:) = f1(:)
+    self%f2_(:) = f2(:)
+
+  end subroutine aero_props_init
+
+  !------------------------------------------------------------------------------
+  ! Ojbect clean
```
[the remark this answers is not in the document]
fvitt: fixed

nusbaume (approved):
I have some small requests, but nothing that should really hold up this PR (assuming other people's requests are resolved).  Thus I'm happy to approve it now.

src/physics/cam/physpkg.F90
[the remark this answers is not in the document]
fvitt: Done

src/chemistry/aerosol/aerosol_state_mod.F90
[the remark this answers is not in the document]
fvitt: This block is replaced with a generic limiter routine (aerosol model independent).

src/physics/cam/ndrop.F90
```diff
@@ -22,30 +21,31 @@ module ndrop
...
+! mathematical constants
+real(r8), parameter :: zero     = 0._r8
+real(r8), parameter :: third    = 1._r8/3._r8
+real(r8), parameter :: twothird = 2._r8*third
+real(r8), parameter :: sixth    = 1._r8/6._r8
+real(r8), parameter :: sq2      = sqrt(2._r8)
+real(r8), parameter :: sq2pi    = sqrt(2._r8*pi)
+real(r8), parameter :: sqpi     = sqrt(pi)
+real(r8), parameter :: surften  = 0.076_r8
+real(r8), parameter :: t0       = 273._r8
```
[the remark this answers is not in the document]
fvitt: This is slightly different (SHR_CONST_TKFRZ = 273.15).

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,201 @@
...
+
+       newobj%exp45logsig_(m) = exp(4.5_r8*alogsig(m)*alogsig(m))
+
+       f1(m) = 0.5_r8*exp(2.5_r8*alogsig(m)*alogsig(m))
+       f2(m) = 1._r8 + 0.25_r8*alogsig(m)
+
+       newobj%voltonumblo_(m) = 1._r8 / ( (pi/6._r8)*                          &
+            (dgnumlo**3._r8)*exp(4.5_r8*alogsig(m)**2._r8) )
+       newobj%voltonumbhi_(m) = 1._r8 / ( (pi/6._r8)*                          &
+            (dgnumhi**3._r8)*exp(4.5_r8*alogsig(m)**2._r8) )
```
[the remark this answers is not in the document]
fvitt: I could make this change. However, this would introduce roundoff differences.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+  !------------------------------------------------------------------------------
+  ! returns number of species masses in a given bin number
+  !------------------------------------------------------------------------------
+  pure function n_masses_per_bin(self,bin_ndx) result(val)
+    class(aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+    integer :: val
+
+    val = self%nmasses_(bin_ndx)
+  end function n_masses_per_bin
```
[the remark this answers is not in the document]
fvitt: This matches `n_masses_all_bins` which returns an array.

src/chemistry/aerosol/aerosol_properties_mod.F90
mattldawson: type-bound procedures are always called on an object, so as long as it's clear from the type what is being "got", I would argue that this:
```
  type(properties_t) :: my_properties
  real :: foo  
  foo = my_properties%get( "foo" )
```
is preferable to this:
```
  type(properties_t) :: my_properties
  real :: foo
  foo = my_properties%get_property( "foo" )
```

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,261 @@
...
+  end subroutine get_transported
+
+  !------------------------------------------------------------------------------
+  ! returns aerosol number, volume concentrations, and bulk hygroscopicity
+  !------------------------------------------------------------------------------
+  subroutine loadaer1( self, aero_props, istart, istop, k,  m, cs, phase, &
+                       naerosol, vaerosol, hygro)
+
+    use aerosol_properties_mod, only: aerosol_properties
+    use modal_aerosol_properties_mod, only: modal_aerosol_properties
```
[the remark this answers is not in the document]
fvitt: removed the use statement here

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
nusbaume: Right, good point! If some other change produces round-off differences then it would be great to bring this in as well, but otherwise I am happy if you want to maintain bit-for-bit output for this PR.

src/chemistry/aerosol/aerosol_properties_mod.F90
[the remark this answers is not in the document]
nusbaume: Fair enough.  At first I was thinking it should match the previous functions given that it returns a single integer value like they do, but I can also see how it would be better to have it match `n_masses_all_bins` given that they return related physical quantities.  I'll go ahead and mark this particular request as resolved.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -130,28 +148,48 @@ subroutine aero_props_init(self, nbin, ncnst, nspec, nmasses, alogsig, f1,f2 )
...
+    allocate(self%f1_(nbin),stat=ierr)
+    if( ierr /= 0 ) then
+       call endrun(prefix//'error allocating self%f1_')
+    end if
+    allocate(self%f2_(nbin),stat=ierr)
+    if( ierr /= 0 ) then
+       call endrun(prefix//'error allocating self%f2_')
+    end if
 
     allocate( self%indexer_(nbin,0:maxval(nmasses)) )
```
nusbaume, line 176: It looks like this `allocate` statement is missing the actual `stat` output (but does have the `if ( ierr /= 0)` check).

src/physics/cam/ndrop.F90
fvitt: fixed

src/chemistry/aerosol/modal_aerosol_state_mod.F90
```diff
@@ -0,0 +1,128 @@
...
+
+  !------------------------------------------------------------------------------
+  !------------------------------------------------------------------------------
+  subroutine get_states( self, aero_props, raer, qqcw )
+    class(modal_aerosol_state), intent(in) :: self
+    class(aerosol_properties), intent(in) :: aero_props
+    type(ptr2d_t), intent(out) :: raer(:)
+    type(ptr2d_t), intent(out) :: qqcw(:)
+
+    integer :: m,mm,l
```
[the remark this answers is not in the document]
fvitt: fixed

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,201 @@
...
+    allocate(nspecies(nmodes))
+    allocate(nmasses(nmodes))
+    allocate(alogsig(nmodes))
+    allocate( f1(nmodes) )
+    allocate( f2(nmodes) )
+
+    allocate(sigmag(nmodes))
+    allocate(newobj%exp45logsig_(nmodes))
+    allocate(newobj%voltonumblo_(nmodes))
+    allocate(newobj%voltonumbhi_(nmodes))
```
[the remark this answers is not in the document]
fvitt: added status checks

src/physics/cam/microp_aero.F90
fvitt: done

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+  !! aerosol_properties class, may include additional private data members and
+  !! type-bound procedures, and may override functions of the abstract class.
+  !!
+  !! Please see the modal_aerosol_properties module for an example of how the
+  !! aerosol_properties class can be extended for a specific aerosol package.
+  type, abstract :: aerosol_properties
+     private
+     integer :: nbins_ = 0
+     integer :: ncnst_tot_ = 0
+     integer, allocatable :: nmasses_(:)
```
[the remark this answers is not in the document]
fvitt: added comments

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,306 @@
...
+
+     !------------------------------------------------------------------------
+     !------------------------------------------------------------------------
+     subroutine aero_props_get(self, m,l, density,hygro)
+       import
+       class(aerosol_properties), intent(in) :: self
+       integer, intent(in) :: m,l
+       real(r8), optional, intent(out) :: density
+       real(r8), optional, intent(out) :: hygro
+     end subroutine aero_props_get
```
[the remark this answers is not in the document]
fvitt: Comments have been added to document the interfaces.   More meaningful names are used for the arguments.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: comments have been added

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: These are used to provide names for cam history output fields.

src/physics/cam/ndrop.F90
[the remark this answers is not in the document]
gold2718: Right. The question is whether that is the correct value. If it is, then what scientific constant is it? It would be great to have some idea of what these quantities are. `t0` seems to be used in equations as the difference between Kelvin and Celsius and 273 is not the correct conversion factor.
What is `surften`? Is it surface tension of water? Newtons / meter? Is it at a particular temperature?

src/chemistry/aerosol/aerosol_state_mod.F90
gold2718: I agree with @fvitt's point. The `rad_cnst_get` interfaces are the main target of what I expected to be dealt with in the next phase.

src/chemistry/aerosol/aerosol_state_mod.F90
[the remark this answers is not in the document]
gold2718: 1. `get` and `set` are pretty vague words. While the OO community uses these terms fairly consistently, how about some help for the scientists that have to use and read this code later on? Please add some documentation.
2. Why aren't these abstract interfaces? Wouldn't each aerosol model have their own implementation of each of these?
3. I just want to note that while these interfaces provide the data for transport, they provide no  information (e.g., wet / dry, MW, name) that are used for transport and/or diagnostics. This interface will need to be added later.
4. Here are my notes about adding the calls:

`number_transported`: We will add the interface for this later when after the CAM thermo rewrite is done.

`get_transported`: A call needs to be added _just before_ each call to `p_d_coupling` 

`set_transported`: A call needs to be added _just after_ each call to `d_p_coupling`

src/physics/cam/microp_aero.F90
fvitt: @tilmes dgnumwet is still used in microp_aero and yes this is MAM specific

tilmes: @fvitt dgnumwet seems to be used to calculate rndst(i,k,3) = 0.5_r8*dgnumwet(i,k,mode_coarse_dst_idx)
What is rndst used for?  I may have thought, we only need it for BAM, but that may not be true.

fvitt: @tilmes RNDST is used in pumas.

tilmes: @fvitt If only used in pumas, why isn't it provided from the aerosol model (as dust radius) directly?

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+  !------------------------------------------------------------------------------
+  subroutine aero_props_init(self, nbin, ncnst, nspec, nmasses, alogsig, f1,f2 )
+    class(aerosol_properties), intent(inout) :: self
+    integer, intent(in) :: nbin               ! number of bins
+    integer, intent(in) :: ncnst              ! total number of constituents
+    integer, intent(in) :: nspec(nbin)        ! number of species in each bin
+    integer, intent(in) :: nmasses(nbin)      ! number of masses in each bin
+    real(r8),intent(in) :: alogsig(nbin)      ! natural log of the standard deviation (sigma) of the aerosol bins
+    real(r8),intent(in) :: f1(nbin)           ! abdul-razzak functions of width
+    real(r8),intent(in) :: f2(nbin)           ! abdul-razzak functions of width
```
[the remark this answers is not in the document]
fvitt: Tried to improve the comments on f1 and f2

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,326 @@
...
+    integer  :: nbins
+    real(r8) :: sum, g1, g2, g1sqrt, g2sqrt
+
+    smax=0.0_r8
+    nbins = self%nbins_
+
+    check_loop: do m=1,nbins
+       if((zeta(m) > 1.e5_r8*eta(m)) .or. (smc(m)*smc(m) > 1.e5_r8*eta(m))) then
+          ! weak forcing -- essentially none activated
+          smax=1.e-20_r8
```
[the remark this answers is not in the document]
fvitt: These are now parameters and tried to document the best I can.

src/physics/cam/microp_aero.F90
```diff
@@ -457,6 +475,9 @@ subroutine microp_aero_run ( &
    integer :: lchnk, ncol
 
    real(r8), allocatable :: factnum(:,:,:) ! activation fraction for aerosol number
+
+   class(aerosol_state), pointer :: aero_state_obj
```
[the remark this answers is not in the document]
fvitt: Initializing the aero_state to null() at the subroutine caused threading issues when compiled with gnu.

src/physics/cam/zm_microphysics.F90
```diff
@@ -187,24 +189,26 @@ module  zm_microphysics
 real(r8), parameter :: mucon = 5.3_r8
 real(r8), parameter :: lambdadpcu = (mucon + 1._r8)/dcon
 
+type(modal_aerosol_properties), pointer :: aero_props_obj
```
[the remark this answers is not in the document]
fvitt: I have initialized the aero_props objects at the module level to null().

src/physics/cam/microp_aero.F90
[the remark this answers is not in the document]
gold2718: It is true that one should (almost) never add initial values (including `null()`) to declarations in threaded code (it causes the `save` attribute to be set which is almost never what you want, even in non-threaded code).
The correct way to implement this behavior (IMHO) is to use the `nullify` statement(s) before other executable statements. In this case, that would be on line 480:
```
nullify(aero_state_obj)
```

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: The modal aerosol model implementation of this function depends on `exp45logsig` property which is unique to the modal representation.  It is more convenient to leave `amcube` in `aerosol_properties` since `exp45logsig` is incapsulated in `modal_aerosol_properties`.

src/physics/cam/microp_aero.F90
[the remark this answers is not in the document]
fvitt: @gold2718 Thanks for the tip.  I added the `nullify` statement to initialize `aero_state_obj`.

src/chemistry/aerosol/aerosol_state_mod.F90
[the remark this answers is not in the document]
fvitt: @gold2718 
I am not familiar with the CAM thermo rewrite and don't know why that would have any relevance with this.   I made an attempt to add calls to these interfaces in fv/stepon and added some commits.  Are the latest code mods moving in the right direction?

src/physics/cam/microp_aero.F90
fvitt: @andrewgettelman Please weigh in on this discussion.  How is RNDST used in pumas?  Do we need to set this pbuf field here?

tilmes: Andrew to clarify, currently, rndst, which seems to be a parameter of 4 dimensions (probably based on the old BAMS scheme), is set to a constant and then only bin 3 (coarse mode) is set to 0.5* diameter of coarse mode dust. To me we should define rndst, whatever that should be) in the aerosol interface, and not in mircop_aero.F90, so it can be defined specifically for each aerosols scheme. The question for you would be then, can rndst be one dimension and what exactly is it? So for CARMA, would we need to define it as a coarse mode dust radius, if so, where would be the cut-off size?

andrewgettelman: rndust can be 1 value, and you can define it wherever you want. It is whatever you want the coarse mode dust size to be. 

It is used in the old contact freezing parameterization, which is CAM5 and not CAM6 code (probably not CAM7 either):

https://github.com/ESCOMP/PUMAS/blob/pumas_cam-release_v1.26/micro_pumas_utils.F90#L1643

Which is from Young, 1974:
Young, K. C. “The Role of Contact Nucleation in Ice Phase Initiation of Clouds.” J. Atmos. Sci. 31 (1974): 768–76.

Which I think just needs whatever sizes of dust are in the model (multiple modes, or one mode). 

Again, we don't use this anymore in CAM6, but not sure I want to break it yet (the CAM6 contact nucleation as part of the classical nucleation theory might need some work).

Basically you do want the aerosols to provide this, but it should be queried in micro_aero_ts probably and loaded into pbuf. Unless you want to do it in the microphysics interface.

Hope that helps.

Regards,

Andrew

gold2718 (changes requested):
I still do not see how this could work if the CAM state was not a pointer.

cacraigucar (commented):
I was surprised to see I have a pending comment, but am unable to find it.  I'm submitting this review so we can see what my comment is.   (Now I see is was just a confirmation reply)

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
fvitt: @gold2718 :   `rad_cnst_get_info` does not use the CAM state.

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,248 @@
...
+       class(aerosol_state), intent(in) :: self
+       class(aerosol_properties), intent(in) :: aero_props ! properties of the aerosol model
+       type(ptr2d_t), intent(out) :: raer(:) ! state of interstitial aerosols
+       type(ptr2d_t), intent(out) :: qqcw(:) ! state of cloud-borne aerosols
+
+     end subroutine aero_get_states
+
+     !------------------------------------------------------------------------------
+     ! sets transported components
+     ! This aerosol model with the state of the transported aerosol constituents
```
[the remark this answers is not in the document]
fvitt: made the change as suggested

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,248 @@
...
+     !------------------------------------------------------------------------------
+     subroutine aero_set_transported( self, transported_array )
+       import :: aerosol_state, r8
+       class(aerosol_state), intent(inout) :: self
+       real(r8), intent(in) :: transported_array(:,:,:)
+     end subroutine aero_set_transported
+
+     !------------------------------------------------------------------------------
+     ! returns transported components
+     ! This returns to current state of the transported aerosol constituents
```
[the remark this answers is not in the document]
fvitt: made the change as suggested

src/dynamics/fv/stepon.F90
```diff
@@ -251,7 +266,16 @@ subroutine stepon_init(dyn_in, dyn_out)
...
       deallocate (delpdryxy)
-      
+
+   end if
+
+   call rad_cnst_get_info(0, nmodes=nmodes)
+   clim_modal_aero = (nmodes > 0)
+
+   if (clim_modal_aero) then
+      aero_props_obj => modal_aerosol_properties()
```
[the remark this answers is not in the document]
fvitt: Now get a pointer to an instantiated  aerosol state object from microp_aero

src/dynamics/fv/stepon.F90
```diff
@@ -333,28 +370,66 @@ subroutine stepon_run1( dtime_out, phys_state, phys_tend, pbuf2d,        &
    call d_p_coupling(dyn_state%grid, phys_state, phys_tend,  pbuf2d, dyn_out)
    call t_stopf('d_p_coupling')
 
+   if (num_trans_aerosols>1) then
```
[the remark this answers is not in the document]
fvitt: Should be `>0` -- made the change

src/dynamics/fv/stepon.F90
```diff
@@ -333,28 +370,66 @@ subroutine stepon_run1( dtime_out, phys_state, phys_tend, pbuf2d,        &
...
+
+      do c = begchunk,endchunk
+         pbuf => pbuf_get_chunk(pbuf2d, c)
+         aero_state_obj => modal_aerosol_state( phys_state(c), pbuf )
+         ! pass number mass or number mixing ratios of aerosol constituents
+         ! to aerosol state object
+         call aero_state_obj%set_transported(aero_constituents(:,:,:,c))
```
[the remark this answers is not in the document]
fvitt: Now pass in `phys_state%q`

src/dynamics/fv/stepon.F90
```diff
@@ -333,28 +370,66 @@ subroutine stepon_run1( dtime_out, phys_state, phys_tend, pbuf2d,        &
    call d_p_coupling(dyn_state%grid, phys_state, phys_tend,  pbuf2d, dyn_out)
    call t_stopf('d_p_coupling')
 
+   if (num_trans_aerosols>1) then
+
+      do c = begchunk,endchunk
+         pbuf => pbuf_get_chunk(pbuf2d, c)
+         aero_state_obj => modal_aerosol_state( phys_state(c), pbuf )
```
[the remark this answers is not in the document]
fvitt: now use aero state created in microp_aero

src/physics/cam/zm_microphysics.F90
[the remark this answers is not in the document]
gold2718: I'm not seeing that change.

gold2718 (commented):
I think the structural issues have been resolved!
There are still some unresolved conversations. Some are open requests for change, others may be fixed. @nusbaume, could you check your open conversations?
Getting really close.

src/physics/cam/zm_microphysics.F90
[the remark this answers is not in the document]
nusbaume: I'm not seeing this change either.  Once this is done all of my concerns will be resolved.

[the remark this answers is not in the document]
nusbaume: Looks good now, thanks!

nusbaume:
@fvitt @gold2718  all of my issues have been resolved, so this PR is good to go (at least for me).  Thanks!

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,258 @@
+module modal_aerosol_properties_mod
+  use shr_kind_mod, only: r8 => shr_kind_r8
+  use physconst, only: pi
+  use aerosol_properties_mod, only: aerosol_properties
+  use rad_constituents, only: rad_cnst_get_info, rad_cnst_get_mode_props, rad_cnst_get_aer_props
+  use cam_abortutils, only: endrun
```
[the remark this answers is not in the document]
gold2718: I would at least like a plan on how we remove CAM dependence here. `pi` is easy, just `SHR_CONST_PI`. 
I think the `endrun` calls should be replaced with a call to a passed-in error handling function that takes a string. Where is a good place to open an issue on this? CAM? NCAR/ACOM-SE-tasks?

src/chemistry/aerosol/aerosol_properties_mod.F90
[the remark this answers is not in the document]
gold2718: @tilmes, Thanks for the summary. Are the issues you raised above resolved?

src/dynamics/fv/stepon.F90
```diff
@@ -251,7 +266,16 @@ subroutine stepon_init(dyn_in, dyn_out)
          end if
       end do
       deallocate (delpdryxy)
-      
+
+   end if
+
+   call rad_cnst_get_info(0, nmodes=nmodes)
```
gold2718, line 272: Is this code only active for modal aerosols?

gold2718 (commented):
Still trying to figure out this code. What is the timing like? It seems like the way `loadaer` is implemented, what used to be a loop over `pcols` is now a loop over `pcols*pcols` (`dropmixnuc` calls `loadaer` for a single bin (`loadaer2`) which calls `loadaer` for an entire array (`pcols` ==> `loadaer1`) and then returns the the desired column). Am I missing something?

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
fvitt: ESCOMP/CAM is a good place for the issue.

src/dynamics/fv/stepon.F90
fvitt: Yes.  For now, only for modal.

gold2718: Is there a plan to create aerosol-model-agnostic interface (e.g., when the rad_cnst code gets absorbed into the new aerosol objects)? If so, it would be good to have an open issue on this as well as the current implementation introduces a new area that is model specific. If not, we need a plan!

fvitt: Yes we need to plan on how to deal with the `rad_cnst` interfaces.  I have not thought that far ahead yet.

fvitt:
> Still trying to figure out this code. What is the timing like? It seems like the way `loadaer` is implemented, what used to be a loop over `pcols` is now a loop over `pcols*pcols` (`dropmixnuc` calls `loadaer` for a single bin (`loadaer2`) which calls `loadaer` for an entire array (`pcols` ==> `loadaer1`) and then returns the the desired column). Am I missing something?

@gold2718 
The `loadaer2` method is only used in carma specific code and is not needed for this PR.  I propose that we leave loadaer2 out of this PR.  I might need to rethink this if it is really needed for carma.

src/chemistry/aerosol/aerosol_state_mod.F90
fvitt: @gold2718 Is this resolved?

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
fvitt: @gold2718 Is this resolved?  Do I need to open the issue?

src/dynamics/fv/stepon.F90
fvitt: This will be dealt with in a later PR.

src/chemistry/aerosol/aerosol_properties_mod.F90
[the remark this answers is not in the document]
tilmes: Yes, that  makes sense. @fvitt This is resolved.

src/chemistry/aerosol/aerosol_state_mod.F90
gold2718: > @gold2718 Is this resolved?

I don't think so. On April 12, you said, "I agree that `endrun` and `iulog` could be easily replaced at this point." but you did not go ahead and replace them. Since it represents an interface change, it would be good to get it into this version.

src/dynamics/fv/stepon.F90
gold2718: Okay. I am leaving this conversation unresolved for this PR.

src/dynamics/fv/stepon.F90
```diff
@@ -251,7 +266,16 @@ subroutine stepon_init(dyn_in, dyn_out)
...
+   clim_modal_aero = (nmodes > 0)
```
gold2718, line 273: It occurs to me that an easy way to correct this might be for microp_aero to make `aero_props_obj` be `public, protected`. Then, this becomes:
```suggestion
   clim_modal_aero = associated(aero_props_mod)
```
and you do not need the call to `rad_cnst_get_info` above. You could also have a logical function interface that returns `associated(aero_props_mod)` so that it works similarly to `aerosol_state_object`.

gold2718 (changes requested):
Just a couple of easy to fix (IMHO) items.

cacraigucar (approved):

src/dynamics/fv/stepon.F90
fvitt: @gold2718 I like your thinking.  I implemented `aerosol_properties_object` which returns the pointer to the `aero_props_obj` in `microp_aero`.  Let me know if you okay with what I did here.

gold2718: @fvitt, yes, this is great, thanks!

src/chemistry/aerosol/aerosol_state_mod.F90
fvitt: I removed the dependencies on `endrun` and `iulog` from the aerosol class code.

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,248 @@
...
+          do i = istart, istop
+             vol(i) = max(qqcw(i,k), 0._r8)/specdens
+          end do
+       else if (phase == 1) then
+          do i = istart, istop
+             vol(i) = max(raer(i,k), 0._r8)/specdens
+          end do
+       else
+          write(iulog,*)'phase = ',phase,' in aerosol_state::loadaer1 not recognized'
+          call endrun('phase error in aerosol_state::loadaer1')
```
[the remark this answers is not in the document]
fvitt: An error string is now returned to caller.

src/dynamics/fv/stepon.F90
gold2718: Turns out this is now taken care of.

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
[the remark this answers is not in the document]
gold2718: This is up to you and ACOM. At some point, you will want to remove CAM dependence but that is not relevant for CAM.

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -0,0 +1,248 @@
...
+
+  !------------------------------------------------------------------------------
+  ! returns aerosol number, volume concentrations, and bulk hygroscopicity
+  ! for a single grid box
+  !------------------------------------------------------------------------------
+  subroutine loadaer2( self, aero_props, i, k, m, cs, phase, &
+                       naerosol, vaerosol, hygro )
+
+    use aerosol_properties_mod, only: aerosol_properties
+    use ppgrid, only: pcols, pver
```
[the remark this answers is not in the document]
gold2718: This has been fixed.

gold2718 (approved):
Looks good now, thanks for all the work!

cacraigucar (changes requested):
At today's meeting, the issue with NAG needing the --mismatch flag to compile was identified as needing to be looked at further.  As a result of this, I am requesting that Steve re-review the code and I am also rescinding my approval until we confirm this isn't an issue with the new class structure.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -0,0 +1,380 @@
...
+     procedure(aero_props_get), deferred :: get
+     procedure(aero_actfracs), deferred :: actfracs
+     procedure(aero_num_names), deferred :: num_names
+     procedure(aero_mmr_names), deferred :: mmr_names
+     procedure(aero_apply_num_limits), deferred :: apply_number_limits
+
+     procedure :: final=>aero_props_final
+  end type aerosol_properties
+
+  interface
```
gold2718, line 56: Since you have deferred methods, I believe this needs to be `abstract`:
```suggestion
  abstract interface
```

gold2718 (changes requested):
I think you are missing an `abstract` keyword

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Do we need to add the `abstract` key word in `aerosol_state_mod.F90` as well?

gold2718: I believe so but doesn't that module already use the `abstract` keyword?

gold2718: Oops, I was looking at the wrong code base. Yes, you should have one but you only run into this nag error in certain (explicitly polymorphic) situations.

fvitt: Thanks. The `abstract` key word fixes the nag issue.

gold2718: Sorry for not noticing that earlier. I think the general rule is:

> An abstract interface gives a name to a set of characteristics and argument keyword names that would constitute an explicit interface to a procedure, without declaring any actual procedure to have those characteristics.

cacraigucar (approved):

tilmes (approved):

src/physics/cam/zm_microphysics.F90
```diff
@@ -25,6 +25,8 @@ module  zm_microphysics
                          secondary_ice_production, accrete_rain_snow, heterogeneous_rain_freezing, &
                          accrete_cloud_water_rain, self_collection_rain, accrete_cloud_ice_snow
 
+use modal_aerosol_properties_mod, only: modal_aerosol_properties
```
gold2718, line 28: This seems like a step backwards.

src/physics/cam/zm_microphysics.F90
```diff
@@ -187,24 +189,26 @@ module  zm_microphysics
 real(r8), parameter :: mucon = 5.3_r8
 real(r8), parameter :: lambdadpcu = (mucon + 1._r8)/dcon
 
+type(modal_aerosol_properties), pointer :: aero_props_obj => null()
```
gold2718, line 192: Why can't this be `class(aerosol_properties)`?

src/physics/cam/zm_microphysics.F90
```diff
@@ -287,6 +291,9 @@ subroutine zm_mphyi
 
         rin = 0.1e-6_r8
 
+! MAM aerosol properties
+   aero_props_obj => modal_aerosol_properties()
```
gold2718, line 295: This seems like a step backwards. Why can't you use `aerosol_properties_object`?

gold2718 (changes requested):
I don't understand why you are using the modal aerosol interface in ZM microphysics instead of the generic routine from microp_aero (`aerosol_properties_object`).

fvitt:
> I don't understand why you are using the modal aerosol interface in ZM microphysics instead of the generic routine from microp_aero (`aerosol_properties_object`).

Sorry for the backwards step.  It was a solution to a failing regression test (outfrq3s_convmic).  I have implemented another solution which I don't think is a step backwards.

src/physics/cam/zm_microphysics.F90
fvitt: There was an issue of timing.  This was attempting to get a pointer to aerosol properties object was instantiated.   I now get the pointer later.

Issue #568, opened 2022-04-14
Add Material Surface Boundary Conditions and Add Hydrodynamic Balance Option


patcal:
The current formulation of the governing equations does not accurately represent the Earths surface at the lower boundary. To assert it's status as a material boundary the Earth applies a force to the atmosphere such that there is no air flow in the direction normal to the surface and such that there is no net momentum tendency in the direction normal to the surface.

adamrher:
I got a little background on this issue from reading the intro in this paper: 

https://journals.ametsoc.org/view/journals/mwre/131/7/1520-0493_2003_131_1229_ncomti_2.0.co_2.xml?tab_body=pdf

PR #569, opened 2022-04-14
Cam hydrodynamic bc

patcal:
Issue  #568 

### Hydrodynamic Balance

- The formulation of hydrostatic balance in the terrain following coordinates assumes that momentum tendencies in the vertical vanish, resulting in an analytic expression of balance. This results in the inherent assumption that the model surfaces instantaneously adjust to satisfy this analytic expression.
- A more rigorous examination of the governing equations reveals a more appropriate balance condition in terrain following coordinates which includes non-zero values of vertical momentum tendencies associated with horizontal momentum tendencies that project onto variations in the model surface heights.
- The resulting differential relation to maintain this hydrodynamic balance is solved via a fixed point iteration which converges after only a few iterations.
- The deviation from hydrostatic balance is conveyed in the expression

        Hdev = (1 + grad_Zm*grad_Zm)/(1 + (dV'/dt)*grad_Zm)
- Rather than the hydrostatic requirement that all vertical momentum tendencies vanish, the resulting contributions in the governing equations associated with Hdev express the more appropriate balance condition that all tendencies normal to the model surfaces must vanish.

### Material Boundary Conditions at the Surface

- The formulation of the governing equations in terrain following coordinates does not restrict the flow of the atmosphere relative to model surfaces. However at the bottom of the atmosphere, the Earth is a material boundary and there is a physical constraint that restricts motions normal to the surface. This constraint is currently not represented in the formulation the equations.
- To maintain it's status as a material boundary, the Earth's surface exerts a force on the atmosphere requiring that large scale motions in the lowest layer must be tangent to the surface and momentum tendencies normal to the surface must vanish.
- The second condition is satisfied by the implementation of hydrodynamic balance above. The first condition requires that an appropriate value must be added to the diagnostic equation for omega and that a corresponding tendency must be added to the horizontal momentum equations.
- Physically it is expected that, since the Earth is only in contact with the lowest layer, the forcing should only be applied to that layer and the effects should be propagated to the layers above via physical processes.
- Currently these processes are not adequately represented in the model. To address this shortcoming, since B(n) represents the effect of the surface terrain on model levels, an influence function based on these values is introduced to convey the surface constraints. With Gamma regulating the extent to which the lower model levels are affected, the influence function implemented has the form:

             S(n) = (B(n)/B(nlev))**Gamma
- First order (visual) testing at 50km resolution over South America leads to a value of about Gamma=8 for the exponent. More rigorous testing is needed to explore the optimal value for this variable and unfortunately the value is most likely going to be resolution dependent. Ideally, a resolution independent physically based function is desired. In the meantime, this should provide an adequate vehicle for exploring the impact of lower boundary forcing on model results.

While there are still some lingering anomalies (e.g. a dry halo) that appear to be associated with the proximity of the SALLJ too close to the face of the Andes ridge, these corrections provide a means of regulating the long standing problem of runaway precipitation over steep topography.

The figure illustrates the impact the changes have on November 2010 mean precipitation at 50km over South America.
PRECT_50km

This figure shows the corresponding impact on Diurnal variations for a profile across the Andes ridge.
Prf_PRECT_PR

November 2010 precipitation at 100km for the standard ne30_mg17 grid.
PRECT_100km

### DIFF
--- a/src/dynamics/se/dycore/element_mod.F90
+++ b/src/dynamics/se/dycore/element_mod.F90
@@ -35,7 +35,6 @@ module element_mod
      !
      ! storage for subcycling tracers/dynamics
      !
-    real (kind=r8) :: dp0  (np,np,nlev)                        ! Hydrodynamic balance correction term
     real (kind=r8) :: vn0  (np,np,2,nlev)                      ! velocity for SE tracer advection
     real (kind=r8) :: dpdiss_biharmonic(np,np,nlev)            ! mean dp dissipation tendency, if nu_p>0
     real (kind=r8) :: dpdiss_ave(np,np,nlev)                   ! mean dp used to compute psdiss_tens
--- a/src/dynamics/se/dycore/element_mod.F90
+++ b/src/dynamics/se/dycore/element_mod.F90
@@ -35,6 +35,7 @@ module element_mod
      !
      ! storage for subcycling tracers/dynamics
      !
+    real (kind=r8) :: dp0  (np,np,nlev)                        ! Hydrodynamic balance correction term
     real (kind=r8) :: vn0  (np,np,2,nlev)                      ! velocity for SE tracer advection
     real (kind=r8) :: dpdiss_biharmonic(np,np,nlev)            ! mean dp dissipation tendency, if nu_p>0
     real (kind=r8) :: dpdiss_ave(np,np,nlev)                   ! mean dp used to compute psdiss_tens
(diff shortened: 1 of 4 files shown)


PeterHjortLauritzen:
Thanks Patrick! Could you give a presentation with the full mathematical details etc.  for those who are interested in AMP? (Rich, Julio, Adam, me, ...)

patcal:
Closing this Pull request in preparation for a new pull request to replace it.

Issue #512, opened 2022-01-25
CIME update needed for running the pumas codes on GPU and introduce a GPU regression test suite


sjsprecious:
In the PR #409 , the pumas codes with tag `v1.22` will be brought in and it includes the file name changes from `micro_mg3_0.F90` to `micro_pumas_v1.F90` (https://github.com/ESCOMP/PUMAS/commit/9c562418de23fbbf3ac6aa507a30b9b187deb60b).

Thus the CIME codes (or https://github.com/ESMCI/ccs_config_cesm for CIME tag later than `cime6.0.13`) need to be updated so that they applies the GPU flags to the correct source codes.

In addition, the CICE5/CICE6/ccs_config modules need to be updated correspondingly before we could introduce a GPU regression test suite

cacraigucar:
@sjsprecious - It sounds like changes need to be made in cime and/or ccs_config_cesm?  If so, those changes need to be made and tested in CESM and tagged.  Once a CESM alpha tag has been made with these changes, please open a PR that contains these updated externals.  If this update requires code changes in CAM as well, then we'll need to coordinate the changes.  Right now the latest cime tag I see is cime6.0.13 and the ccs_config tag is ccs_config_cem0.0.4.

sjsprecious:
Thanks @cacraigucar and yes, changes are needed in cime and ccs_config_cesm. I will do what you suggested on the cime side first and bring in those tags in my CAM PR later.

cacraigucar:
@sjsprecious I just wanted to let you know that #513 has been tagged.  This brings in cime6.0.13 and introduces the ccs_config external ccs_config_cesm0.0.4 .  I believe you indicated you will need mods beyond these which is why I left this issue open.

sjsprecious:
@cacraigucar , thanks a lot for your help. I have addressed this issue in this PR (https://github.com/ESMCI/ccs_config_cesm/pull/8) but there is a runtime error of using NUOPC driver with the NVHPC compiler and OpenMPI lib. Jim and I are working on this issue right now and I will update it here once the issue is resolved.

On the other hand, this tag (https://github.com/ESMCI/ccs_config_cesm/tree/ccs_config_cesm0.0.6) works fine with MCT driver and nvhpc/openmpi on Casper.

sjsprecious:
[...]

PR #577, opened 2022-04-20
part of cam6_3_056: PUMAS GPU regression test suite

sjsprecious:
This PR introduces a new regression test suite for the GPU-enabled PUMAS codes, which works on Casper.

There is no source code change here but just a few xml files. It also requires the updated `CICE5/CICE6/CIME/ccs_config` modules to work properly.

This regression test suite includes five different ERP tests:
- CAM run with MG2, default namelist
- CAM run with MG3, default namelist
- CAM run with MG3, change the MG logical variables from their default values to the opposite values
- CAM run with MG3, default namelist, different PCOLS values
- CAM run with MG3, default namelist, different number of GPUs per node

Previously I made some GPU changes and somehow the CAM run could finish but return wrong results. The latter two new tests are able to detect those unexpected NBFB changes quickly before doing an ECT test and therefore I add them to the test suite.

To generate a baseline for this GPU regression test suite on Casper, use the following commands:

- `cd /path_to_CAM_main_dir/test/system`
- `module load python/3.7.9`
- `env BL_TESTDIR='' CAM_ACCOUNT=YOUR_PROJECT_ID CAM_FC=nvhpc-gpu CIME_MACHINE=casper ./test_driver.sh --cesm casper_gpu --baseline-dir /path_to_save_baseline -f
`

To perform a GPU regression test against the baseline generated above, use the following commands:
- `env BL_TESTDIR=/path_to_save_baseline CAM_ACCOUNT=YOUR_PROJECT_ID CAM_FC=nvhpc-gpu CIME_MACHINE=casper ./test_driver.sh --cesm casper_gpu --no-baseline -f`

The status of the regression test could be viewed through the following commands:
- `cd /glade/scratch/$user/casper_gpu`
- `./cs.status.casper_gpu_nvhpc-gpu_xxxx`, where `xxxx` is the time stamp for this test run

Ideally we should observe `PASS` for all the output from the command above. However, I saw `FAIL` for the `COMPARE_base_rest` and there were 4 different fields in the cpl restart file. I later ran the ERP test for the PUMAS CPU codes on Casper and got the same error. However, if I switched the `ERP` test to `ERS` test and re-ran it with PUMAS GPU codes, there was no `FAIL` in the output. Thus I think the error message here is not caused by the PUMAS GPU codes but I do not know how to fix this issue.

Fix https://github.com/ESCOMP/CAM/issues/512.


cime_config/testdefs/testmods_dirs/cam/outfrq9s_mg2_default/shell_commands
```diff
@@ -0,0 +1,28 @@
...
+./xmlchange ROOTPE_OCN='0'
+./xmlchange NTASKS_GLC=36
+./xmlchange NTHRDS_GLC=1
+./xmlchange ROOTPE_GLC='0'
+./xmlchange NTASKS_WAV=36
+./xmlchange NTHRDS_WAV=1
+./xmlchange ROOTPE_WAV='0'
+./xmlchange NTASKS_CPL=36
+./xmlchange NTHRDS_CPL=1
+./xmlchange ROOTPE_CPL='0'
```
gold2718, line 24: For future reference, `xmlchange` recognizes these variables as categories so you can replicate this behavior with:
```suggestion
./xmlchange NTASKS=36
./xmlchange NTHRDS=1
./xmlchange ROOTPE='0'
```

test/system/test_driver.sh
```diff
@@ -251,7 +251,7 @@ case $hostname in
     mach_workspace="/glade/scratch"
 
     # Check for CESM baseline directory
-    if [ -n "{$BL_TESTDIR}" ] && [ ! -d "${BL_TESTDIR}" ]; then
+    if [ -n "${BL_TESTDIR}" ] && [ ! -d "${BL_TESTDIR}" ]; then
```
gold2718, line 254: Nice catch! I guess folks were not running this with a blank value for `BL_TESTDIR`.

gold2718 (approved):
Couple of comments but otherwise, looks good!

gold2718 (changes requested):
On second thought, I would like a chance to see your ChangeLog entry (can be before you have final test data).

cime_config/testdefs/testmods_dirs/cam/outfrq9s_mg2_default/shell_commands
sjsprecious: Thanks @gold2718. I just updated all the `shell_commands` files with your suggestions.

test/system/test_driver.sh
sjsprecious: Thanks! That is exactly how I find out this issue and I hope I am not the only one using an empty `BL_TESTDIR`.

cacraigucar (approved):

test/system/test_driver.sh
```diff
@@ -465,7 +520,12 @@ if [ "${cesm_test_suite}" != "none" -a -n "${cesm_test_mach}" ]; then
     fi
 
     ## Setup CESM work directory
-    cesm_testdir=$mach_workspace/$LOGNAME/$test_id
+    if [ "${hostname:0:6}" == "casper" ] || [ "${hostname:0:5}" == "crhtc" ]; then
+       ## Would fail to compile on Casper with long foler name
```
nusbaume, line 524: Possibly a typo here (`foler`)?

nusbaume (approved):
Looks good to me!  Just one possible typo.

peverwhee (approved):

test/system/test_driver.sh
sjsprecious: Thanks @nusbaume for catching it. Just fix the typo!

peverwhee:
this was brought in with #581

PR #595, opened 2022-05-20
Add capability for SCAM forcing based on gridded met fields.

### LINKED ISSUE #594: Want to add capability for SCAM to be forced by gridded reanalyses/model fields.

I'd like to add some new files, modify some existing files in src/physics/cam, src/dynamics/eul, bld/namelist_files ...

@jtruesdal @adamrher @andrewgettelman @

I Don't Know

JulioTBacmeister:
This capability currently requires pre-computed gridded fields on FV latlon 1 degree grid

Have run a number of cases using this capability - can provide user_nl_cam as well path to gridded ERA reanalysis.

Have also run with this capability is off and standard SCAM IOPs. Answers *look* fine but have not done rigorous testing.

closes #594


jtruesdal:
@JulioTBacmeister could you add @jtruesdal as a developer on your repo if it's OK for me to make changes to this branch for the PR.

JulioTBacmeister:
@jtruesdal I've made you a collaborator and github implies you should have received an invite of some sort. Let me know if I should do something else.

peverwhee:
@JulioTBacmeister @jtruesdal 

Jim Edwards came across the following when debugging our CICE/SCAM issues:

Because the namelist variable **scm_ana_frc_file_template** is a template rather than an actual file, this message appears when you build:

> Model cam missing file scm_ana_frc_file_template = '%y/ERAI_fv09_L58.cam2.i.%y-%m-%d-%[s.nc](http://s.nc/)'

To avoid this, we may need a change to CAM's buildnml so that this isn't written to the
Buildconf/cam.input_data_list

JulioTBacmeister:
Yes. I've noticed this message.  The model nevertheless builds and runs without issues.  This should have also come up when the nudging code was incorporated - it uses a similar template.

peverwhee:
> Yes. I've noticed this message. The model nevertheless builds and runs without issues. This should have also come up when the nudging code was incorporated - it uses a similar template.

Gotcha. I'll open a low-priority issue for the nudging message.

peverwhee:
@JulioTBacmeister @jtruesdal 

OK, bringing this up again because it looks like the logic can handle the nudging template file because it does not see it as an absolute path (and thus we do not get the warning). Because of the "%y/" at the start of the scm_ana_frc_file_template file, we get the warning for this new variable.

If you replace line 4405 in bld/build_namelist with the following (to not write the variables whose pathnames start with '%' to the cam.input_data_list file), it should remove the warning:

```
if (!($pathname =~ '%.*') {
    print $fh "$var = $pathname\n";
}
```

I'm also happy to make a PR to your branch with this change.

Thanks!

JulioTBacmeister:
I made the change in build_namelist on my branch and pushed it to origin.  Would you check it?

peverwhee:
@JulioTBacmeister Looks good to me. Sorry my code snippet was missing a parenthesis! Thanks!

bld/namelist_files/namelist_definition.xml
```diff
@@ -5336,6 +5336,79 @@ Force scam to use the lat lon fields specified in the scam namelist not what is
...
+Interpolate ana fields to constant pressure surfaces
+Default: FALSE
+</entry>
+
+
+<entry id="scm_ana_frc_file_template" type="char*128" input_pathname="abs" category="scam"
+       group="scam_nl" valid_values="" >
+template for analysis forcing dataset.
+Default: set by build-namelist.
+</entry>
```
gold2718, line 5398: This variable is a string, not a filename (path), therefore, it should not have the `input_pathname` attribute. I believe this is the source of the build-namelist problems.

bld/build-namelist
```diff
@@ -4404,9 +4404,8 @@ sub check_input_files {
                         print $fh "$var = $pathname\n";
                     }
                     else {
-                        if (-e $pathname) {  # use -e rather than -f since the absolute pathname
-                                             # might be a directory
-                            print "OK -- found $var = $pathname\n";
+                        if (!($pathname =~ '%.*')) {
+                            print $fh "$var = $pathname\n";
```
gold2718, line 4408: I do not think this should be needed (see below) and could have unintended effects.

gold2718 (changes requested):
Suggested approach to problem with pathnames.

peverwhee:
@nusbaume @cacraigucar and I recommended (via email) that Julio remove the shell scripts and ncdata file from the scam_STUB usermods directory. Documenting that here.

XML variables set in the shell script(s) can be set in scam_STUB/shell_commands and variables written to user_nl_cam can be added to scam_STUB/user_nl_cam

cacraigucar:
@JulioTBacmeister and @adamrher - Is this Stub IOP another IOP which we want to support?  If so, I would suggest that the user_nl_cam and shell_commands be put in cime_config/usermods_dirs/scam_Stub_IOP or something like that.  You can see what was done for cime_config/usermods_dirs/scam_arm97 as an example.  @jtruesdal may be able to give some addtional guidance as he's done this more recently than I have.

Externals_CAM.cfg
```diff
@@ -21,16 +21,16 @@ required = True
...
 required = True
 
 [silhs]
 local_path = src/physics/silhs
-protocol = git
-repo_url = https://github.com/ESCOMP/SILHS_CESM
-tag = silhs_clubb_release_b76a124_20200220_c20200320
+protocol = svn
+repo_url = https://github.com/larson-group/clubb_release/tags/
+tag = clubb_4ncar_20220311_f51de38/src/SILHS
```
cacraigucar, line 33: @JulioTBacmeister - Are these changes really needed for your SCAM PR?  Hopefully you can use the tags that are currently being used in CAM instead

JulioTBacmeister: I want to use the "new" CLUBB external brought in with cam6_3_059.  This is so that we can assess prognostic momentum transport using SCAM.

JulioTBacmeister:
Superseded by PR #628

PR #597, opened 2022-05-20
PUMAS updates and fixes

### LINKED ISSUE #593: PUMAS Microphysics Updates and Fixes
### Issue Type

Code Clean-up

### Issue Description

This corresponds to PUMAS issue: https://github.com/ESCOMP/PUMAS/issues/38 
Associated with CAM PR https://github.com/ESCOMP/CAM/pull/597 and PUMAS PR https://github.com/ESCOMP/PUMAS/pull/41

I am readying code for PUMAS that does several things:

1. Adds number tendencies to PUMAS for output through CAM history
2. Cleans up numerics for the vapor deposition onto snow limiters (this was probably a bug in the previous version)
3. Changes Immersion Freezing from the CNT version in CAM6 BACK to the old Bigg formulation in CAM5

Item 3 is necessary because immersion freezing from CNT is not bounded and does not work well at long timesteps (>60s or les). So it was having a massive influence on ice number. Instead we will go back to the Bigg formulation from CAM5, at least for now.

Code will be ready soon: cam and PUMAS code. It's pretty straightforward from an SE perspective. It will change answers. I will try to tune back the climate to make it look like standard CAM6. 

### Will this change answers?

Yes

cacraigucar:
Adds number tendencies to PUMAS for output through CAM history
Cleans up numerics for the vapor deposition onto snow limiters (this was probably a bug in the previous version)
Changes Immersion Freezing from the CNT version in CAM6 BACK to the old Bigg formulation in CAM5

Closes #593

### DIFF
--- a/src/physics/cam/micro_pumas_cam.F90
+++ b/src/physics/cam/micro_pumas_cam.F90
@@ -111,7 +111,11 @@ logical  ::  micro_mg_evap_scl_ifs = .false.      ! Scale evaporation as IFS doe
 logical  ::  micro_mg_evap_rhthrsh_ifs = .false.  ! Evap RH threhold following IFS
 logical  ::  micro_mg_rainfreeze_ifs = .false.    ! Rain freezing at 0C following IFS
 logical  ::  micro_mg_ifs_sed = .false.           ! Snow sedimentation = 1 m/s following IFS
-logical  ::  micro_mg_precip_fall_corr = .false.    ! Precip fall speed following IFS
+logical  ::  micro_mg_precip_fall_corr = .false.    ! Precip fall speed following IFS (does not go to zero)
+
+logical  ::  micro_mg_implicit_fall = .false. !Implicit fall speed (sedimentation) for hydrometeors
+
+logical  ::  micro_mg_accre_sees_auto = .false.    !Accretion sees autoconverted rain
 
 character(len=10), parameter :: &      ! Constituent names
    cnst_names(10) = (/'CLDLIQ', 'CLDICE','NUMLIQ','NUMICE', &
@@ -269,8 +273,8 @@ subroutine micro_pumas_cam_readnl(nlfile)
        micro_do_massless_droplet_destroyer, &
        micro_mg_evap_sed_off, micro_mg_icenuc_rh_off, micro_mg_icenuc_use_meyers, &
        micro_mg_evap_scl_ifs, micro_mg_evap_rhthrsh_ifs, &
-       micro_mg_rainfreeze_ifs, micro_mg_ifs_sed, micro_mg_precip_fall_corr
-
+       micro_mg_rainfreeze_ifs, micro_mg_ifs_sed, micro_mg_precip_fall_corr, &
+       micro_mg_accre_sees_auto, micro_mg_implicit_fall
 
   !-----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


andrewgettelman:
Note that current clubb_detice_rad = 1.D-6 (one micron!) which is too small and not really physical (results in too high ice number in convective regions).

cacraigucar:
> Note that current clubb_detice_rad = 1.D-6 (one micron!) which is too small and not really physical (results in too high ice number in convective regions).

@andrewgettelman - Should this be changed for all runs, not just cam_dev?

adamrher:
> @adamrher - We are proposing having the namelist settings for cam_dev be as described in the description. Do you see any problems with this?

Not really. Two of these are already set w/ cam_dev:
micro_mg_accre_sees_auto = .true.
micro_mg_implicit_fall = .true.

@andrewgettelman this one is currently set to 25D-6, not 1D-6. But propose doubling it to:
clubb_detice_rad = 61.0D-6

This is the only one that is not currently a namelist default
micro_mg_precip_fall_corr = .true.

andrewgettelman:
Yes, that is correct. So it's just a minor change. Increasing the detrained ice size is a bit more physical, but we might tune it further later.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1346,15 +1335,35 @@
...
+    </machines>
+    <options>
+      <option name="wallclock">01:00:00</option>
+    </options>
+  </test>
+  <test compset="F2000dev" grid="f09_f09_mg17" name="ERP_Ln9_Vnuopc" testmods="cam/outfrq9s_mg3" supported="false">
     <machines>
+      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+      <machine name="cheyenne" compiler="intel" category="prealpha"/>
       <machine name="cheyenne" compiler="intel" category="prebeta"/>
```
cacraigucar, line 1351: @Katetc - Having a test be prealpha and prebeta is redundant (there are no code changes made between prealpha and prebeta - it is just additional tests which are run).  I propose removing the prebeta test.  

I will also change the time limit to be 20 minutes as my sample run for the MG2 took a little over 5 minutes.  MG3 shouldn't be dramatically more expensive, should it?

Katetc: I was wondering why so many tests are one or the other. I thought I would cover my bases here, but if it's redundant than no worries. Just pre-alpha is fine. 20 minutes should be fine too.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1346,15 +1335,35 @@
...
+    </machines>
+    <options>
+      <option name="wallclock">01:00:00</option>
+    </options>
+  </test>
+  <test compset="F2000dev" grid="f19_f19" name="SMS_Ld1_Vnuopc" testmods="cam/outfrq9s" supported="false">
+    <machines>
       <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+      <machine name="cheyenne" compiler="intel" category="prealpha"/>
```
cacraigucar, line 1360: @Katetc  What was your thought about adding a prealpha test?  I am tempted to remove the aux_cam test for the f19 grid as the functionality is being tested with the ne30pg3 test and the prealpha test will catch any problems with changes with externals that aux_cam would not.

Katetc: I wanted a longer run (1 day instead of just 9 steps) for both aux_cam and pre alpha. I put it on the lower resolution so it would run a little quicker. Also, I have found issues pop up with ncols/pcols issues in higher fv resolution but not in lower fv resolution (and rarely vice versa) so I think it's a good idea to test f09 and f19 all the time.

Externals_CAM.cfg
```diff
@@ -36,8 +36,8 @@ required = True
 [pumas]
 local_path = src/physics/pumas
 protocol = git
-repo_url = https://github.com/ESCOMP/PUMAS
-tag = pumas_cam-release_v1.22
+repo_url = https://github.com/cacraigucar/PUMAS
+hash = 039331f7
```
gold2718, line 40: Why do we need a CAM tag with an external from a personal fork hash?

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1523,6 +1579,31 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
...
+   real(r8)  :: nprcio(state%psetcols,pver)     
+   real(r8)  :: ncsedten(state%psetcols,pver)     
+   real(r8)  :: nisedten(state%psetcols,pver)     
+   real(r8)  :: nrsedten(state%psetcols,pver)     
+   real(r8)  :: nssedten(state%psetcols,pver)    
+   real(r8)  :: ngsedten(state%psetcols,pver)    
+   real(r8)  :: nmelto(state%psetcols,pver)    
+   real(r8)  :: nmeltso(state%psetcols,pver)   
+   real(r8)  :: nmeltgo(state%psetcols,pver)   
+!--   ag
```
gold2718, line 1605: Is this needed?

gold2718 (changes requested):
Why the fork hash?

nusbaume (approved):

peverwhee (approved):

cime_config/testdefs/testlist_cam.xml
cacraigucar: @Katetc - I just realized I didn't finalize the discussion on these tests.  We will leave these three tests being tested at both aux_cam and prealpha for now since cam_dev testing is limited.   It should be  understood that as more tests are moved to cam_dev physics, that these tests will be limited to either aux_cam or prealpha dependent on other tests which are being added.

Issue #624, opened 2022-07-21
WACCMX issue with pio2_5_7


fvitt:
### What happened?

WACCM-X compiled with NAG on izumi aborts with error message:

 Abort with message Multiple sources in map for a single destination in file /home/fvitt/camdev/cam6_3_070/libraries/parallelio/src/clib/pio_darray.c at line 676

### What are the steps to reproduce the bug?

SMS_D_Ln3_Vnuopc.ne5pg3_ne5pg3_mg37.QPX2000.izumi_nag.cam-outfrq3s

### What CAM tag were you using?

cam6_3_068

### What machine were you running CAM on?

CGD machine (e.g. izumi)

### What compiler were you using?

NAG

### Path to a case directory, if applicable

NA

### Will you be addressing this bug yourself?

Yes, but I will need some help

### Extra info

The issue seems to be related to the output of history fields which are on the geomagnetic grid.  If I remove the geomag grid fields from the history output the test will pass.    The test also passes if I use pio2_5_6. 

Jim Edwards commented:
"..the error indicates that more than one source point in the decomposition map is going to the same destination point.   Check the dof argument to pio_init_decomp."

PR #625, opened 2022-07-22
part of cam6_3_071: Corrections to WACCMX coordinate maps for oplus and geomag history grids

fvitt:
Fixes issue #624 

The follow test passes with these changes:
 SMS_D_Ln3_Vnuopc.ne5pg3_ne5pg3_mg37.QPX2000.izumi_nag.cam-outfrq3s

This PR moves this WACCMX test back to izumi_nag

### DIFF
--- a/cime_config/testdefs/testlist_cam.xml
+++ b/cime_config/testdefs/testlist_cam.xml
@@ -604,7 +604,7 @@
   </test>
   <test compset="QPX2000" grid="ne5pg3_ne5pg3_mg37" name="SMS_D_Ln3_Vnuopc" testmods="cam/outfrq3s">
     <machines>
-      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+      <machine name="izumi" compiler="nag" category="aux_cam"/>
       <machine name="izumi" compiler="nag" category="waccmx"/>
     </machines>
     <options>
@@ -2156,7 +2156,7 @@
       <option name="wallclock">00:50:00</option>
     </options>
   </test>
-  <test compset="FXHIST" grid="f09_f09_mg17" name="SMS_D_Ln9_Vnuopc"   testmods="cam/waccmx_weimer">
+  <test compset="FXHIST" grid="f09_f09_mg17" name="SMS_D_Ln9_Vnuopc_P384x3"   testmods="cam/waccmx_weimer">
     <machines>
       <machine name="cheyenne" compiler="intel" category="waccmx"/>
     </machines>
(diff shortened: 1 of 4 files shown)


src/ionosphere/waccmx/edyn_init.F90
```diff
@@ -181,14 +181,22 @@ subroutine register_grids()
...
            map=coord_map)
       nullify(coord_map)
 
       allocate(coord_map(omlon1 - mlon0 + 1))
-      coord_map = (/ (i, i = mlon0, omlon1) /)
+      if (mlat0==1) then
+         coord_map = (/ (i, i = mlon0, omlon1) /)
+      else
+         coord_map = 0
+      end if
```
gold2718, line 199: Yowza! This is similar to the fix that went into FV when PIO2 came along. I still can't figure out how this worked with previous versions of PIO2.

gold2718 (approved):

src/ionosphere/waccmx/edyn_init.F90
fvitt: @gold2718 ,
I think more consistency checks where added to the latest version of PIO2.  

It is mysterious to my how this worked on cheyenne with pio2.5.7 (latest pio2 tag).  

I don't completely understand the fix.  I just mimicked what was done for fv_centers grid.

nusbaume (approved):

peverwhee (approved):

cacraigucar (changes requested):
I never hit submit yesterday on this!

cime_config/testdefs/testlist_cam.xml
```diff
@@ -604,7 +604,7 @@
   </test>
   <test compset="QPX2000" grid="ne5pg3_ne5pg3_mg37" name="SMS_D_Ln3_Vnuopc" testmods="cam/outfrq3s">
     <machines>
-      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+      <machine name="izumi" compiler="nag" category="aux_cam"/>
```
cacraigucar, line 607: I know this is addressing a bug with the NAG compiler with this compset/configuration and as such is correct to do.  Since the NAG regression test suite is already overloaded, is it possible to move a different WACCM NAG test to be GNU, so that the net result is that the number of NAG tests is not increased.

fvitt: Moved FW4madSD to gnu

cacraigucar (approved):

Issue #368, opened 2021-04-23
Add feature for zonal mean output and feature for zonal mean/zonal eddy nudging in physics, for any grid


nadavis-ncar:
**Summary**

There should be a feature to provide zonal mean history output for any dycore. There should also be a feature that supports zonal mean/eddy nudging within physics, as this expands the capability of CESM to support fundamental research into the atmospheric circulation.

**Background**

The capability to output pressure-interpolated zonal mean quantities supports rapid model assessment, reduces overhead on users, and reduces demand on disk space.

Model evaluation of dynamics and chemistry is often grounded in an analysis of the zonal mean climate. This is not just convenient but also physically motivated, as the governing physics of insolation and fluid dynamics vary fundamentally with latitude (and not with longitude). Zonal mean output reduces the disk space required to quantify basic wave-mean flow dynamics by several orders of magnitude by eliminating the longitude dimension and the need to sample instantaneous output at high time resolution. The finite volume dycore currently provides zonal mean output via the ctem.F90 module, so the ability to output zonal mean quantities can be seen as a regression test against existing functionality.

In addition to output capability, there is a scientific need to have access to the zonal mean of prognostic fields within physics modules. For example, nudging either the zonal mean or zonally-anomalous part of the flow can effectively constrain wave-mean flow feedbacks, supporting better research into the circulation response to climate forcings. This ability exists implicitly in the finite volume dycore and can be implemented with simple source modifications.

However, because the spectral element and other dycores employ non-rectilinear grids, these capabilities are absent. 

Adoption of future dycores will be hindered by their current inability to support online zonal mean capability and zonal mean output, especially as the finite volume dycore possesses both.

**Specification**

“Zonal mean” is defined as an average over the longitude dimension with dimensions of pressure and latitude. It should be an average in physical units, and not an average with dimensions of, for example, model level. 

A physics package should be able to call a function to retrieve the zonal mean of a meteorological/advected model field, defined on the local column(s). This implies the zonal mean, in dimensions of pressure and latitude, can be converted back to model level and (arbitrary) latitude appropriate for each column. For this specification, “meteorological” means variables like the wind vector (u, v, w) and temperature, while “advected” means any advected constituents such as ozone or water vapor.

[...]

PR #629, opened 2022-08-02
cam6_3_092:  Cam zonal mean 01

### LINKED ISSUE #744: CONUS prealpha test needs longer time limit
### What happened?

According to @fischer-ncar: 
 SMS_D_Ln9_Vnuopc_P720x1.ne0CONUSne30x8_ne0CONUSne30x8_mt12.FCnudged.cheyenne_intel.cam-outfrq9s

Is constantly taking longer than 20 minutes to run.  30 minutes should work.

Just need to increase time limit

cam6_3_086

CISL machine (e.g. cheyenne)

Intel

_No response_

Any CAM SE can do this

### Extra info

_No response_

patcal:
Initial Implementation of zonal_mean.F90 module for computing Zonal Mean values to answer issue #368.

Implemented usage to optionally apply a zonal mean filter in nudging.F90.

Note the usage description in the comments is for the stand-alone offline 
zonal_mean module. The usage description for the installed MPI version 
can be found at the top of the zonal_mean.F90 module.

closes #744


src/physics/cam/nudging.F90
```diff
@@ -270,16 +271,24 @@ module nudging
...
   real(r8)          :: Nudge_Hwin_max
   real(r8)          :: Nudge_Hwin_min
+  logical           :: Nudge_ZonalFilter =.false.
+  integer           :: Nudge_ZonalNbasis
+
+  ! Nudging Zonal Filter variables
+  !---------------------------------
+  type(ZonalMean_t)   :: ZM
+  real(r8),allocatable:: Zonal_Bamp2d(:)
+  real(r8),allocatable:: Zonal_Bamp3d(:,:)
```
fvitt, line 281: @patcal You made the basis amplitude arrays global (ie, shared by all openMP threads) by declaring these at the module level.  Will this be thread safe?  This might be okay if these are used at timestep_init when the model is not threaded.

src/physics/cam/zonal_mean.F90
```diff
@@ -0,0 +1,2093 @@
+module Zonal_Mean
```
fvitt, line 1: The name of this module clashes with `src/dynamics/fv/zonal_mean.F90`

fvitt:
@patcal The zonal mean utility seems to produce slightly different results when changing the number of MPI tasks.  Do you have any ideas why this might be the case?

patcal:
Hi @fvitt 
For me the shr_reprosum_calc() routine is a black box and I inferred the arguments 
for it from other usage cases. So I expect that there may be some optional arguments
that can be added to provide stable results. Based on the ~available~ documentation, I 
have no idea what they should be. There are differences in the arguments I use for this 
routine in the Zonal_Mean module where I was trying different settings, but that was just 
guesswork so it is possible that those optional arguments may be causing problems.

fvitt:
@patcal,
I have come to the conclusion that roundoff errors are occurring before the sum across the tasks takes place via `shr_reprosum_calc`.   There are differences in the last 1 or 2 of digits of the sums, depending on the number of mpi tasks (or PE layout).  I have yet to come up with clever way of dealing with the differences.

nadavis-ncar:
It would be helpful to have the option to zonally filter the target data in the same way as the model data. There are cases where we may want to use the same target data, but use either the full field or the zonal mean as we see fit. That way we wouldn't need to create multiple target datasets. For flexibility it should probably be a separate namelist option. E.g., Nudge_ZonalFilter_Model and Nudge_ZonalFilter_Target.

nadavis-ncar:
Disregard the above, I see how this propagates through to the target data as well.

patcal:
Patrick,

No, I overlook the changes where this propagates to the target data as
well. You can disregard my previous comment.

-Nick

Nicholas A. Davis <https://staff.ucar.edu/users/nadavis>
Atmospheric Chemistry Observations and Modeling
National Center for Atmospheric Research

fvitt:
@patcal Are you okay with me pushing changes directly to your cam_ZonalMean_01 branch in your github fork?  Or would you prefer a pull request?

patcal:
@Francis Vitt ***@***.***> ,
Yes,  go ahead and push your mods to cam_ZonalMean_01.

src/physics/cam/nudging.F90
```diff
@@ -1230,7 +1279,7 @@ subroutine nudging_timestep_init(phys_state)
          ncol=phys_state(lchnk)%ncol
          Target_S(:ncol,:pver,lchnk)=cpair*Target_T(:ncol,:pver,lchnk)
        end do
-     elseif(Nudge_TSmode == 1) then
+     elseif(Nudge_TSmode == 1) then   !***** FOR ZONAL MEAN VALUES WILL THIS OPTION BE A PROBLEM?? *****
```
fvitt, line 1282: @patcal Should we not allow zonal filtering when Nudge_TSmode == 1 ?

bld/namelist_files/namelist_definition.xml
```diff
@@ -345,6 +345,16 @@
...
+<entry id="Nudge_ZonalFilter" type="logical" category="nudging"
+       group="nudging_nl" valid_values="" >
+       Zonal Mean filtering ...
+       Default: FALSE
+       </entry>
+
+<entry id="Nudge_ZonalNbasis" type="integer" category="nudging"
+       group="nudging_nl" valid_values="" >
+       Number of basis functions used in Zonal Mean filtering ...
+       </entry>
```
fvitt, line 357: @patcal Can you provide descriptions for these namelist options?

patcal:
@Francis Vitt ***@***.***>
TSmode was an option created in the early development of the nudging
module and I don't know if users ever use anything other than the default
setting.
This likely will go away in upcoming revisions to nudging, so I would not
spend time implementing zonal filtering for this option now.

--> Patrick

src/utils/zonal_mean_mod.F90
```diff
@@ -0,0 +1,2084 @@
...
+      ! calc_ZonalProfile_1Damps: Given 1D data values for the nlat zonal
+      !                           profiles gridpoints, compute the zonal
+      !                           profile basis amplitudes.
+      !=====================================================================
+      !
+      ! Passed Variables
+      !------------------
+      class(ZonalProfile_t):: this
+      real(r8),intent(in ):: I_Zdata(:)
+      real(r8),intent(out):: O_Bamp (:)
```
fvitt, line 777: @patcal It seems that the zdata input to ZonalProfile needs to be on the zonal average profile grid of size nlat, rather than on the cam physics grid (pcols,begchunk:endgrid).  Is this correct?  I don't see how this can be used in cam without having the input zdata on the zonal averaged grid.

patcal:
Hi @Nicholas Davis ***@***.***>

That base functionality is already included. There is not a namelist
interface for switching to
zonal nudging yet, currently it is a hardcoded 'Nudge_ZonalFilter' flag
that when toggled to true, filters
3D model and the 3D input target data to zonal mean values, and then
applies nudging to the
differences. So the target data does not need to be pre-processed with a
zonal filter before
reading it in. Were you suggesting options to be able to, for example,
alternatively nudge 3D
model states to zonal filtered target data?

--> Patrick

fvitt (approved):

fvitt:
@cacraigucar , This PR is ready for other SEs to review.

nusbaume (approved):
Looks good!  Just a few small requests/questions.

src/physics/cam/nudging.F90
```diff
@@ -894,6 +911,17 @@ subroutine nudging_init
    endif
 !!DIAG
 
+   ! Initialize the Zonal Mean type if needed
+   !------------------------------------------
+   if(Nudge_ZonalFilter) then
+     call ZM%init(Nudge_ZonalNbasis)
+     allocate(Zonal_Bamp2d(Nudge_ZonalNbasis))
+     allocate(Zonal_Bamp3d(Nudge_ZonalNbasis,pver))
```
nusbaume, line 919: Whoops, forgot one last request.  Can you add `stat` checks to these `allocate` statements.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1551,6 +1551,68 @@
...
+      <machine name="cheyenne" compiler="intel" category="camchem"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:30:00</option>
+      <option name="comment">Zonal Mean Nudged</option>
+    </options>
+  </test>
+ <test compset="FCnudged" grid="f09_f09_mg17" name="SMS_D_Ln9_Vnuopc" testmods="cam/outfrq3s_zmean_nudging">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="prealpha"/>
```
cacraigucar, line 1566: Is it necessary to add a new test, or can one of the existing tests (FCnudged) be expanded to use this setup?

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1551,6 +1551,68 @@
...
+    <options>
+      <option name="wallclock">00:10:00</option>
+      <option name="comment">Zonal Mean Nudged</option>
+    </options>
+  </test>
+  <test compset="QPC4" grid="ne5_ne5_mg37" name="ERC_D_Ln9_Vnuopc" testmods="cam/outfrq3s_nudging_ne5_L26">
+    <machines>
+      <machine name="izumi" compiler="nag" category="zmean_nudging"/>
+      <machine name="izumi" compiler="gnu" category="zmean_nudging"/>
+      <machine name="izumi" compiler="gnu" category="aux_cam"/>
```
cacraigucar, line 1609: Curious why this test is using CAM4 physics instead of CAM6 (QPC6)?  Also, some tests were added with the TEM PR.  Could one of them be modified to handle what this is testing and we don't need to add a new aux_cam test?

src/physics/cam/nudging.F90
```diff
@@ -1586,6 +1653,46 @@ subroutine nudging_set_profile(rlat,rlon,Nudge_prof,Wprof,nlev)
...
+    if (allocated(Nudge_File_Present)) deallocate(Nudge_File_Present)
+    if (allocated(Nobs_U)) deallocate(Nobs_U)
+    if (allocated(Nobs_V)) deallocate(Nobs_V)
+    if (allocated(Nobs_T)) deallocate(Nobs_T)
+    if (allocated(Nobs_Q)) deallocate(Nobs_Q)
+    if (allocated(Nobs_PS)) deallocate(Nobs_PS)
+
+    call ZM%final()
+    if (allocated(Zonal_Bamp2d)) deallocate(Zonal_Bamp2d)
+    if (allocated(Zonal_Bamp3d)) deallocate(Zonal_Bamp3d)
```
cacraigucar, line 1692: I would suggest moving these lines above `call ZM%final()`.  The reason being that if additional variables are deallocated in the future, they most likely will be placed at the end, and the call to `ZM%final` may become lost for someone looking quickly at the routine.

src/physics/cam/nudging.F90
```diff
@@ -1461,6 +1524,10 @@ subroutine nudging_update_analyses(anal_file)
               1,pcols,begchunk,endchunk,Tmp2D,        &
               VARflag,gridname='physgrid',timelevel=1 )
    if(VARflag) then
+     if(Nudge_ZonalFilter) then
+       call ZM%calc_amps(Tmp2D,Zonal_Bamp2d)
+       call ZM%eval_grid(Zonal_Bamp2d,Tmp2D)
+     endif
      Nobs_PS(:,begchunk:endchunk,Nudge_ObsInd(1)) = Tmp2D(:,begchunk:endchunk)
    else
      call endrun('Varibale "PS" is missing in '//trim(anal_file))
```
cacraigucar, line 1533: `Varibale` should be `Variable` here and all the other lines where the typo occurs above

cacraigucar (changes requested):

src/physics/cam/nudging.F90
```diff
@@ -520,6 +533,8 @@ subroutine nudging_readnl(nlfile)
    call MPI_bcast(Nudge_Vwin_Lindex  , 1, mpi_real8 ,  mstrid, mpicom, ierr)
    call MPI_bcast(Nudge_Vwin_Ldelta  , 1, mpi_real8 ,  mstrid, mpicom, ierr)
    call MPI_bcast(Nudge_Vwin_Invert,   1, mpi_logical, mstrid, mpicom, ierr)
+   call MPI_bcast(Nudge_ZonalFilter,   1, mpi_logical, mstrid, mpicom, ierr)
+   call MPI_bcast(Nudge_ZonalNbasis,   1, mpi_integer, mstrid, mpicom, ierr)
```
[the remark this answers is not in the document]
fvitt: @cacraigucar Do I need to add checks after all the mpi_bcast calls?  If the MPI bast fails does it actually return an error code that can be dealt with?

cime_config/testdefs/testlist_cam.xml
fvitt: QPC4 is a cheaper configuration and is sufficient for testing the new functionality.

cime_config/testdefs/testlist_cam.xml
fvitt: This is testing a different type of nudging and thus needs to be tested separately.  I would be OK with removing this from the prealpha list if we think the ne5 aux_cam test added below is sufficient.

src/physics/cam/nudging.F90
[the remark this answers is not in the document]
cacraigucar: According to:  https://www.mpich.org/static/docs/latest/www3/MPI_Bcast.html, there are a number of useful error codes that are returned.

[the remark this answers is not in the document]
fvitt: @cacraigucar  There are 72 calls to mpi_bcast in this module.  Do I need to add checks after all of those mpi_bcast calls?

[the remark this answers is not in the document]
fvitt: I have added all the checks.

src/physics/cam/nudging.F90
fvitt: done

src/physics/cam/nudging.F90
fvitt: fixed

src/physics/cam/nudging.F90
fvitt: done

cime_config/testdefs/testlist_cam.xml
cacraigucar: What about moving it to prebeta?  We don't expect it to fail, but it is a good "last ditch check"

cacraigucar (approved):

PR #632, opened 2022-08-03
Introduce PUMAS DDT

### LINKED ISSUE #589: Introduce process rate DDT for PUMAS #589
Create a DDT which will contain the process rate variables for PUMAS

cacraigucar:
Brings in the PUMAS DDT (proc_rates)

closes #589

### DIFF
--- a/Externals_CAM.cfg
+++ b/Externals_CAM.cfg
@@ -36,8 +36,8 @@ required = True
 [pumas]
 local_path = src/physics/pumas
 protocol = git
-repo_url = https://github.com/ESCOMP/PUMAS
-tag = pumas_cam-release_v1.22
+repo_url = https://github.com/cacraigucar/PUMAS
+hash =  1faa0ba1
 required = True
 
 [pumas-frozen]
--- a/Externals_CAM.cfg
+++ b/Externals_CAM.cfg
@@ -37,7 +37,7 @@ required = True
 local_path = src/physics/pumas
 protocol = git
 repo_url = https://github.com/cacraigucar/PUMAS
-hash =  1faa0ba1
+hash =   2dcd19c8
 required = True
 
 [pumas-frozen]
--- a/src/physics/cam/micro_pumas_cam.F90
+++ b/src/physics/cam/micro_pumas_cam.F90
@@ -816,11 +816,12 @@ end subroutine micro_pumas_cam_init_cnst
 
 !===============================================================================
 
-subroutine micro_pumas_cam_init(pbuf2d)
+subroutine micro_pumas_cam_init(pbuf2d, proc_rates)
    use time_manager,   only: is_first_step
    use micro_pumas_utils, only: micro_pumas_utils_init
    use micro_mg1_0, only: micro_mg_init1_0 => micro_mg_init
    use micro_pumas_v1, only: micro_mg_init3_0 => micro_pumas_init
+   use micro_pumas_diag, only: proc_rates_type
 
    !-----------------------------------------------------------------------
    !
@@ -829,6 +830,7 @@ subroutine micro_pumas_cam_init(pbuf2d)
    !-----------------------------------------------------------------------
 
    type(physics_buffer_desc), pointer :: pbuf2d(:,:)
+   type(proc_rates_type), pointer     :: proc_rates
 
    integer :: m, mm
    logical :: history_amwg         ! output the variables used by the AMWG diag package
@@ -845,6 +847,9 @@ subroutine micro_pumas_cam_init(pbuf2d)
 
    call phys_getopts(use_subcol_microp_out=use_subcol_microp, &
                      do_clubb_sgs_out     =do_clubb_sgs)
+
+   ! Allocate the elements within the proc_rates DDT
+   call proc_rates%allocate(psetcols,nlev)
 
    if (do_clubb_sgs) then
      allow_sed_supersat = .false.
@@ -952,7 +957,7 @@ subroutine micro_pumas_cam_init(pbuf2d)
    call addfld ('CME',        (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of cond-evap within the cloud'                      )
    call addfld ('PRODPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of conversion of condensate to precip'              )
    call addfld ('EVAPPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling precip'                   )
-   call addfld ('EVAPSNOW',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
+   call addfld ('EVAPSNOW',   (/ 'trop_cld_nlev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
    call addfld ('HPROGCLD',   (/ 'lev' /), 'A', 'W/kg'    , 'Heating from prognostic clouds'                          )
    call addfld ('FICE',       (/ 'lev' /), 'A', 'fraction', 'Fractional ice content within cloud'                     )
    call addfld ('CLDFSNOW',   (/ 'lev' /), 'A', '1',        'Cloud fraction adjusted for snow'                        )
@@ -1440,8 +1445,8 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    real(r8)  :: prect(state%psetcols)
    real(r8)  :: preci(state%psetcols)
    real(r8)  :: am_evp_st(state%psetcols,pver)  ! Area over which precip evaporates
-   real(r8)  :: evapsnow(state%psetcols,pver)   ! Local evaporation of snow
-   real(r8)  :: prodsnow(state%psetcols,pver)   ! Local production of snow
+!   real(r8)  :: evapsnow(state%psetcols,pver)   ! Local evaporation of snow
+!   real(r8)  :: prodsnow(state%psetcols,pver)   ! Local production of snow
    real(r8)  :: cmeice(state%psetcols,pver)     ! Rate of cond-evap of ice within the cloud
    real(r8)  :: qsout(state%psetcols,pver)      ! Snow mixing ratio
    real(r8)  :: cflx(state%psetcols,pverp)      ! grid-box avg liq condensate flux (kg m^-2 s^-1)
@@ -2156,10 +2161,10 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    prect(:ncol)=0._r8
    preci(:ncol)=0._r8
    nevapr(:ncol,:top_lev-1)=0._r8
-   evapsnow(:ncol,:top_lev-1)=0._r8
+!  NO LONGER NEEDED AS THE FIELDS ONLY GO TO top_lev?evapsnow(:ncol,:top_lev-1)=0._r8
    am_evp_st(:ncol,:top_lev-1)=0._r8
    prain(:ncol,:top_lev-1)=0._r8
-   prodsnow(:ncol,:top_lev-1)=0._r8
+!  NO LONGER NEEDED AS THE FIELDS ONLY GO TO top_lev?   prodsnow(:ncol,:top_lev-1)=0._r8
    cmeice(:ncol,:top_lev-1)=0._r8
    dei(:ncol,:top_lev-1)=0._r8
    mu(:ncol,:top_lev-1)=0._r8
@@ -2280,6 +2285,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
                  rel_fn_dum(:ncol,top_lev:),      &
                  rei(:ncol,top_lev:), prect(:ncol), preci(:ncol), nevapr(:ncol,top_lev:), evapsnow(:ncol,top_lev:), &
                  am_evp_st(:ncol,top_lev:), &
+! CAC proc_rates will replace the individual elements
                  prain(:ncol,top_lev:), prodsnow(:ncol,top_lev:), cmeice(:ncol,top_lev:), dei(:ncol,top_lev:), mu(:ncol,top_lev:), &
                  lambdac(:ncol,top_lev:), qsout(:ncol,top_lev:), des(:ncol,top_lev:), rflx(:ncol,top_lev:), sflx(:ncol,top_lev:), &
                  qrout(:ncol,top_lev:), reff_rain_dum(:ncol,top_lev:), reff_snow_dum(:ncol,top_lev:), qcsevap(:ncol,top_lev:), &
@@ -2332,6 +2338,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
               prect(:ncol),           preci(:ncol),           &
               nevapr(:ncol,top_lev:),          evapsnow(:ncol,top_lev:),        &
               am_evp_st(:ncol,top_lev:),                               &
+! CAC proc_rates will replace the individual elements
               prain(:ncol,top_lev:),           prodsnow(:ncol,top_lev:),        &
               cmeice(:ncol,top_lev:),          dei(:ncol,top_lev:),             &
               mu(:ncol,top_lev:),              lambdac(:ncol,top_lev:),         &
@@ -2613,6 +2620,8 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       call subcol_field_avg(qme,       ngrdcol, lchnk, qme_grid)
       call subcol_field_avg(nevapr,    ngrdcol, lchnk, nevapr_grid)
       call subcol_field_avg(prain,     ngrdcol, lchnk, prain_grid)
+! CAC - Check whether this will work with fields only going to top_lev - YES it is okay as it finds the size of the second and
+! subsequent dimensions of the field passed in
       call subcol_field_avg(evapsnow,  ngrdcol, lchnk, evpsnow_st_grid)
       call subcol_field_avg(bergso,    ngrdcol, lchnk, bergso_grid)
 
@@ -2715,7 +2724,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
 
       am_evp_st_grid  = am_evp_st
 
-      evpsnow_st_grid = evapsnow
+      evpsnow_st_grid(:,top_lev:) = proc_rates%evapsnow
       qrout_grid      = qrout
       qsout_grid      = qsout
       nsout_grid      = nsout
@@ -3304,7 +3313,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    call outfld('MPDICE',      qiten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('MPDNLIQ',     ncten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('MPDNICE',     niten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
-   call outfld('EVAPSNOW',    evapsnow,    psetcols, lchnk, avg_subcol_field=use_subcol_microp)
+   call outfld('EVAPSNOW',    proc_rates%evapsnow,    psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QCSEVAP',     qcsevap,     psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QISEVAP',     qisevap,     psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QVRES',       qvres,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
--- a/src/physics/cam/micro_pumas_cam.F90
+++ b/src/physics/cam/micro_pumas_cam.F90
@@ -816,12 +816,11 @@ end subroutine micro_pumas_cam_init_cnst
 
 !===============================================================================
 
-subroutine micro_pumas_cam_init(pbuf2d, proc_rates)
+subroutine micro_pumas_cam_init(pbuf2d)
    use time_manager,   only: is_first_step
    use micro_pumas_utils, only: micro_pumas_utils_init
    use micro_mg1_0, only: micro_mg_init1_0 => micro_mg_init
    use micro_pumas_v1, only: micro_mg_init3_0 => micro_pumas_init
-   use micro_pumas_diag, only: proc_rates_type
 
    !-----------------------------------------------------------------------
    !
@@ -830,7 +829,6 @@ subroutine micro_pumas_cam_init(pbuf2d, proc_rates)
    !-----------------------------------------------------------------------
 
    type(physics_buffer_desc), pointer :: pbuf2d(:,:)
-   type(proc_rates_type), pointer     :: proc_rates
 
    integer :: m, mm
    logical :: history_amwg         ! output the variables used by the AMWG diag package
@@ -847,9 +845,6 @@ subroutine micro_pumas_cam_init(pbuf2d, proc_rates)
 
    call phys_getopts(use_subcol_microp_out=use_subcol_microp, &
                      do_clubb_sgs_out     =do_clubb_sgs)
-
-   ! Allocate the elements within the proc_rates DDT
-   call proc_rates%allocate(psetcols,nlev)
 
    if (do_clubb_sgs) then
      allow_sed_supersat = .false.
@@ -957,7 +952,7 @@ subroutine micro_pumas_cam_init(pbuf2d, proc_rates)
    call addfld ('CME',        (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of cond-evap within the cloud'                      )
    call addfld ('PRODPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of conversion of condensate to precip'              )
    call addfld ('EVAPPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling precip'                   )
-   call addfld ('EVAPSNOW',   (/ 'trop_cld_nlev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
+   call addfld ('EVAPSNOW',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
    call addfld ('HPROGCLD',   (/ 'lev' /), 'A', 'W/kg'    , 'Heating from prognostic clouds'                          )
    call addfld ('FICE',       (/ 'lev' /), 'A', 'fraction', 'Fractional ice content within cloud'                     )
    call addfld ('CLDFSNOW',   (/ 'lev' /), 'A', '1',        'Cloud fraction adjusted for snow'                        )
@@ -1445,8 +1440,8 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    real(r8)  :: prect(state%psetcols)
    real(r8)  :: preci(state%psetcols)
    real(r8)  :: am_evp_st(state%psetcols,pver)  ! Area over which precip evaporates
-!   real(r8)  :: evapsnow(state%psetcols,pver)   ! Local evaporation of snow
-!   real(r8)  :: prodsnow(state%psetcols,pver)   ! Local production of snow
+   real(r8)  :: evapsnow(state%psetcols,pver)   ! Local evaporation of snow
+   real(r8)  :: prodsnow(state%psetcols,pver)   ! Local production of snow
    real(r8)  :: cmeice(state%psetcols,pver)     ! Rate of cond-evap of ice within the cloud
    real(r8)  :: qsout(state%psetcols,pver)      ! Snow mixing ratio
    real(r8)  :: cflx(state%psetcols,pverp)      ! grid-box avg liq condensate flux (kg m^-2 s^-1)
@@ -2161,10 +2156,10 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    prect(:ncol)=0._r8
    preci(:ncol)=0._r8
    nevapr(:ncol,:top_lev-1)=0._r8
-!  NO LONGER NEEDED AS THE FIELDS ONLY GO TO top_lev?evapsnow(:ncol,:top_lev-1)=0._r8
+   evapsnow(:ncol,:top_lev-1)=0._r8
    am_evp_st(:ncol,:top_lev-1)=0._r8
    prain(:ncol,:top_lev-1)=0._r8
-!  NO LONGER NEEDED AS THE FIELDS ONLY GO TO top_lev?   prodsnow(:ncol,:top_lev-1)=0._r8
+   prodsnow(:ncol,:top_lev-1)=0._r8
    cmeice(:ncol,:top_lev-1)=0._r8
    dei(:ncol,:top_lev-1)=0._r8
    mu(:ncol,:top_lev-1)=0._r8
@@ -2285,7 +2280,6 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
                  rel_fn_dum(:ncol,top_lev:),      &
                  rei(:ncol,top_lev:), prect(:ncol), preci(:ncol), nevapr(:ncol,top_lev:), evapsnow(:ncol,top_lev:), &
                  am_evp_st(:ncol,top_lev:), &
-! CAC proc_rates will replace the individual elements
                  prain(:ncol,top_lev:), prodsnow(:ncol,top_lev:), cmeice(:ncol,top_lev:), dei(:ncol,top_lev:), mu(:ncol,top_lev:), &
                  lambdac(:ncol,top_lev:), qsout(:ncol,top_lev:), des(:ncol,top_lev:), rflx(:ncol,top_lev:), sflx(:ncol,top_lev:), &
                  qrout(:ncol,top_lev:), reff_rain_dum(:ncol,top_lev:), reff_snow_dum(:ncol,top_lev:), qcsevap(:ncol,top_lev:), &
@@ -2338,7 +2332,6 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
               prect(:ncol),           preci(:ncol),           &
               nevapr(:ncol,top_lev:),          evapsnow(:ncol,top_lev:),        &
               am_evp_st(:ncol,top_lev:),                               &
-! CAC proc_rates will replace the individual elements
               prain(:ncol,top_lev:),           prodsnow(:ncol,top_lev:),        &
               cmeice(:ncol,top_lev:),          dei(:ncol,top_lev:),             &
               mu(:ncol,top_lev:),              lambdac(:ncol,top_lev:),         &
@@ -2620,8 +2613,6 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       call subcol_field_avg(qme,       ngrdcol, lchnk, qme_grid)
       call subcol_field_avg(nevapr,    ngrdcol, lchnk, nevapr_grid)
       call subcol_field_avg(prain,     ngrdcol, lchnk, prain_grid)
-! CAC - Check whether this will work with fields only going to top_lev - YES it is okay as it finds the size of the second and
-! subsequent dimensions of the field passed in
       call subcol_field_avg(evapsnow,  ngrdcol, lchnk, evpsnow_st_grid)
       call subcol_field_avg(bergso,    ngrdcol, lchnk, bergso_grid)
 
@@ -2724,7 +2715,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
 
       am_evp_st_grid  = am_evp_st
 
-      evpsnow_st_grid(:,top_lev:) = proc_rates%evapsnow
+      evpsnow_st_grid = evapsnow
       qrout_grid      = qrout
       qsout_grid      = qsout
       nsout_grid      = nsout
@@ -3313,7 +3304,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    call outfld('MPDICE',      qiten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('MPDNLIQ',     ncten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('MPDNICE',     niten,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
-   call outfld('EVAPSNOW',    proc_rates%evapsnow,    psetcols, lchnk, avg_subcol_field=use_subcol_microp)
+   call outfld('EVAPSNOW',    evapsnow,    psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QCSEVAP',     qcsevap,     psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QISEVAP',     qisevap,     psetcols, lchnk, avg_subcol_field=use_subcol_microp)
    call outfld('QVRES',       qvres,       psetcols, lchnk, avg_subcol_field=use_subcol_microp)
--- a/src/physics/cam_dev/micro_pumas_cam.F90
+++ b/src/physics/cam_dev/micro_pumas_cam.F90
@@ -947,14 +947,14 @@ subroutine micro_pumas_cam_init(pbuf2d)
    call addfld ('ICIMRST',    (/ 'lev' /), 'A', 'kg/kg',    'Prognostic in-stratus ice mixing ratio'                  )
 
    ! MG microphysics diagnostics
-   call addfld ('QCSEVAP',    (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling cloud water'              )
-   call addfld ('QISEVAP',    (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of sublimation of falling cloud ice'                )
-   call addfld ('QVRES',      (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of residual condensation term'                      )
-   call addfld ('CMEIOUT',    (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of deposition/sublimation of cloud ice'             )
-   call addfld ('VTRMC',      (/ 'lev' /), 'A', 'm/s',      'Mass-weighted cloud water fallspeed'                     )
-   call addfld ('VTRMI',      (/ 'lev' /), 'A', 'm/s',      'Mass-weighted cloud ice fallspeed'                       )
-   call addfld ('QCSEDTEN',   (/ 'lev' /), 'A', 'kg/kg/s',  'Cloud water mixing ratio tendency from sedimentation'    )
-   call addfld ('QISEDTEN',   (/ 'lev' /), 'A', 'kg/kg/s',  'Cloud ice mixing ratio tendency from sedimentation'      )
+   call addfld ('QCSEVAP',    (/ 'trop_cld_lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling cloud water'              )
+   call addfld ('QISEVAP',    (/ 'trop_cld_lev' /), 'A', 'kg/kg/s',  'Rate of sublimation of falling cloud ice'                )
+   call addfld ('QVRES',      (/ 'trop_cld_lev' /), 'A', 'kg/kg/s',  'Rate of residual condensation term'                      )
+   call addfld ('CMEIOUT',    (/ 'trop_cld_lev' /), 'A', 'kg/kg/s',  'Rate of deposition/sublimation of cloud ice'             )
+   call addfld ('VTRMC',      (/ 'trop_cld_lev' /), 'A', 'm/s',      'Mass-weighted cloud water fallspeed'                     )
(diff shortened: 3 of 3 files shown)


cacraigucar:
@andrewgettelman and @Katetc - This is the full implementation of the DDT based on the list of variables that @andrewgettelman provided in issue #589.    Please review this and let me know if there are any issues or if additional fields should be added.

Also, the variable "prodsnow" is not used in cam_dev/micro_pumas_cam.F90 once it is returned.  I just wanted to point this out

andrewgettelman (approved):
1. Why are some variables 'trop_cld_lev' and some just 'lev'? It doesn't seem to be logical? E.g. UMR, UMS = lev, but UMG = trop_cld_lev?

2. So now everything is in 'proc_rates'. I think this looks pretty readable to me and don't see an issue. 

Thanks!

Andrew

cacraigucar:
> 1. Why are some variables 'trop_cld_lev' and some just 'lev'? It doesn't seem to be logical? E.g. UMR, UMS = lev, but UMG = trop_cld_lev?
> 2. So now everything is in 'proc_rates'. I think this looks pretty readable to me and don't see an issue.
> 
> Thanks!
> 
> Andrew



> 1. Why are some variables 'trop_cld_lev' and some just 'lev'? It doesn't seem to be logical? E.g. UMR, UMS = lev, but UMG = trop_cld_lev?
> 2. So now everything is in 'proc_rates'. I think this looks pretty readable to me and don't see an issue.
> 
> Thanks!
> 
> Andrew

The answer to #1 is that you found a bug!  I have added trop_cld_lev to these variables and will push it soon.

#2 Yes, the new DDT is called proc_rates.  The name can be anything, so if you have a better name, let me know.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1158,13 +1132,11 @@ subroutine micro_pumas_cam_init(pbuf2d)
    call addfld('VPRAO' ,       horiz_only,   'A', 'kg/kg/s',  'Vertical average of accretion rate'                                )
    call addfld('RACAU' ,       horiz_only,   'A', 'kg/kg/s',  'Accretion/autoconversion ratio from vertical average'              )
 
-   if (micro_mg_version > 1) then
-      call addfld('UMR', (/ 'lev' /), 'A',   'm/s', 'Mass-weighted rain  fallspeed'              )
-      call addfld('UMS', (/ 'lev' /), 'A',   'm/s', 'Mass-weighted snow fallspeed'               )
-   end if
+   call addfld('UMR', (/ 'lev' /), 'A',   'm/s', 'Mass-weighted rain  fallspeed'              )
```
andrewgettelman, line 1135: UMR, UMS and UMG are the same thing for different species (fall speed for Rain, Snow and Grapupel), but UMR and UMS are defined as 'lev'. Shouldn't they be 'trop_cld_lev' since they are part of the DDT?

cacraigucar: You must have missed my response above.  You found a bug which I've fixed and am getting ready to push back to my branch.

sjsprecious:
A naive question: can we pass all the columns and levels to PUMAS in the future so that we do not need to worry about 'ncol' vs. 'pcol' and 'trop_cld_lev' vs. 'lev'? By doing this, we will compute some additional columns and levels, but I guess we won't use them anywhere else and it should hurt nothing. Or will these additional columns/levels crash PUMAS later?

cacraigucar:
> A naive question: can we pass all the columns and levels to PUMAS in the future so that we do not need to worry about 'ncol' vs. 'pcol' and 'trop_cld_lev' vs. 'lev'? By doing this, we will compute some additional columns and levels, but I guess we won't use them anywhere else and it should hurt nothing. Or will these additional columns/levels crash PUMAS later?

@andrewgettelman and/or @Katetc  - @sjsprecious question will need to be answered by one of you.  I don't know the answer.

andrewgettelman:
We decided for various reasons to have a smaller number of levels in the diagnostic output fields since they would just be extra memory to carry around. I still think it is a bit asymmetric, but for efficiency we decided to do this (less memory used). 

pcols v. ncols has to do with potential uneven allocations by chunks I believe.  I'm not sure why it goes all the way into the microphysics, I thought that was all interface level.

Katetc:
@sjsprecious So columns that are >ncol but <pcol are typically not initialized or used anywhere in the model, and that could cause a crash. Levels above trop_cld_lev should, in theory, be ok but I don't think other parts of the model expect the physics to look at those, so they could have bad values. Also, the very high-top model, WACCMX, has many levels of atmosphere above the cloud top that should not have any microphysics working on, and we could slow down that model if we have to look at all of those levels in microphysics.

sjsprecious:
Thanks @andrewgettelman and @Katetc for your clarifications. The reason I asked this question is that I have observed a GPU performance hit with the unpacked interface for PUMAS. This is due to the fact that GPU is accessing partial slices of an array here and GPU may have to generate an underlying temporary array first (@johnmauff could comment more on this issue). It won't be a problem if we could simply pass all the columns and levels to GPU, but the answers above make sense to me.

andrewgettelman:
We need to assess whether the extra memory or the GPU efficiency is more important. How can we do that?

cacraigucar:
I would suggest that this discussion of memory/GPU efficiency should become its own issue.  A few months from now, this discussion may be hard to find as it is buried in the DDT PR.

sjsprecious:
Thanks @andrewgettelman and @cacraigucar for your comments. I agree that my question should be a separate discussion and I apologize for bringing it up here. We can definitely discuss more in another GitHub issue.

One quick question about this PR: Is using `proc_rates` part of the efforts of SIMA or CCPP work? Or is it only to make the argument list shorter?

andrewgettelman:
The derived type for process rates is just to make the argument list shorter. It eliminates some steps in adding diagnostics to PUMAS down the road.

sjsprecious:
Thanks @andrewgettelman for the clarification.

gold2718:
> I apologize for bringing it up here. We can definitely discuss more in another GitHub issue.

No need to apologize. This is just the normal flow to make sure that we do not drop an important idea or request. One possibility (in a future PR) would be to control the allocation of vertical levels via a namelist variable that would allow a run which uses more memory but might run more efficiently on the GPU.

sjsprecious:
Thanks @gold2718. I assume that we will encounter a similar issue when we port more physics components to GPU and glad to know that you already have some ideas in mind.

cacraigucar:
@sjsprecious - I have committed the changes so that PUMAS is now receiving ncol instead of psetscols sized arrays.  My subcolumn tests ran fine a week or two ago, but I have not run it recently.  I assume that my recent changes have not brokens subcolumns, but I will confirm that before I move this PR out of draft status.

@Katetc and @andrewgettelman  - I have committed changes which appear to be _mostly_ running properly for WACCM.  I have one variable PGSACWO which has answer changes between a cam run made without DDTs and one with these changes.  I can not see why this might differ from a number of other variables which are placed back onto the grid before being output and work fine.

@Katetc and @andrewgettelman  - I am not sure how to check the variables that have the trop_cld_lev dimension on output as my usual test using cprnc does not work when the dimensions differ. 

I welcome any suggestions.

sjsprecious:
Thanks @cacraigucar for your efforts and it is great news to know that *ncol* works so far. This should simplify the GPU changes that I have made and I could provide you an updated GPU revision later once you are done with all your tests.

cacraigucar:
I verified that SILHS (subcolumns) still works with this current version and is bit-for-bit with cam6_3_072.  The WACCM run is not, which indicates to me that I might have an error in how I'm handling the mapping from the compressed vertical grid (for the proc_rates DDT) and the nlev arrays in the interface code.  It may be important to note that PGSACWO is bit-for-bit in the SILHS run and not for the WACCM run (in other words, subcolumn handling is okay, but the level handling may have an error for PGSACWO).  The kicker is that the other fields which I map back onto a grid from the proc_rates DDT are bit-for-bit in the WACCM run.

My runs can be seen at:

SILHS
/glade/scratch/cacraig/ddt_silhs_try15-fulloutput/run
/glade/scratch/cacraig/cam6_silhs_try15-fulloutput/run

WACCM
/glade/scratch/cacraig/ddt_FW2000climo_dev_try15/run
 /glade/scratch/cacraig/cam6_FW2000climo_dev_try15/run

andrewgettelman:
Hi @cacraigucar , I took a look.

Differences are only in the upper part of the profile: i.e., the top 27 levels of a 70 level model (i.e. levels 1-27). What I don't understand about your simulations is that the values of say PGSACWO and PSACWSO are zero in the lower atmosphere and have very small values in the upper 27 levels. Why is that? I guess you have trop_cld_lev = 43 so that's why. 

So I think you are looking at noise or something? It's hard to be plotting profiles that are not mappable here with SE code, might be easier to look at FV. 

I guess I am confused as to why a process rate has 70 levels in WACCM when PGSACWO should probably have 43 levels like the other ddt terms?  Maybe we need a call about this.

andrewgettelman:
Also: trop_cld_lev as a dimension needs to have the standard pressure levels it is being referenced too, otherwise there is no way to map from lev to trop_cld_lev. Right now trop_cld_lev is an index.

I'm starting to think we made a bad decision here, and we should just dump all the diagnostics on the lev index, forgetting compacting the ddt in the vertical. Thoughts?

cacraigucar:
@andrewgettelman - I know there are still changes that have been requested like the standard pressure levels to address.  I just hadn't gotten around to addressing them and all the changes that Kate requested yet as I've been concentrating on the runs.  I now realize that you need the field filled in properly for plotting purposes, so I will work on that around meetings today.  

As for writing out all the variables on the same vertical levels, that would mean that all fields which are on the trop_cld_lev would need to have additional fields to which they'd be copied.  Extra memory and extra copy work.  I could probably use the same temp variable over and over right before the outfld calls, if you feel this is important.  There just is no way to subset fields in a DDT, like the other variables are being subset in the vertical dimension at the PUMAS call, so copying to new variables for output would be the only solution.

The WACCM run is one I got from Peter as I was having difficulties getting a WACCM run to work with cam_dev (not fully supported yet). I'm not sure if a FV run will work or not - I had problems getting any WACCM run to work for me at all.  This particular one fails on timestep 19 (for both me and Peter using the unmodified code)

andrewgettelman:
@Katetc and I talked today @cacraigucar, and perhaps we can ignore the WACCM issues for now? If WACCM is broken with cam_dev, that's a separate issue, not for us to fix with this check in.

Instead: could you just run cam setting the top_lev for the physics to something > 1 and run with and without the DDT? That would be a simpler test, and easier I think to debug.

gold2718:
> changes that have been requested like the standard pressure levels to address

Hmm, it occurs to me that this is not trivial. I think you need to change the `add_hist_coord` on line 587 to `add_vert_coord`, however, you do not have a good way of obtaining the values (which are set through the dycore in hycoef).
I can add a new interface, `hist_dimension_values`, that would allow you to grab these values from `lev`. Then, you could mimic the `add_vert_coord` call in hycoef.F90.

andrewgettelman:
This is becoming a bit of a problem. It's not helpful to just get vertical level indicies that don't match the model levels. Perhaps we should just sort out the last variable issues (see my comment from yesterday, do NOT use WACCM for this), and then have to put this aside and wait for new infrastructure.  Otherwise we just to the internal copy as is done now and output on the model levels. But this seems to be a rabbit hole that might not be worth it right now.

cacraigucar:
@andrewgettelman - Before I reverted to using WACCM, I had tried setting trop_cloud_top_press=7.D3.  When I ran the corresponding setup using cam6_3_072, it bombed out on the first time step with a floating point exception during the file writes.  At this point, I figured I might be doing something wrong and heard that Peter was testing WACCM and cam_dev, so I reverted to testing there.  I figured if I could get even a few time steps, I could debug the DDT and it did help me to find and correct a few minor errors.  The failing cam6 using cam_dev and setting trop_cloud_top_press=7.D3 can be seen at:  /scratch/cluster/cacraig/cam6_try084_nag-fulloutput on izumi.  Note that I figured this would be timed off shortly, so I "touched" the files which I think may be important to keep around.

andrewgettelman:
So the control model bombed when resetting trop_cloud_top_press? That seems like a bug. I'm thinking this is a rabbit hole right now that might be worth pulling back from. You seem to be spending a lot of time chasing down other problems in cam and cam_dev. Thoughts?

cacraigucar:
@andrewgettelman - That is exactly why I didn't investigate that bug further and instead concentrated on using a WACCM run that ran successfully for a handful of timesteps.  It did help me find a few places which needed fixing.  My concern is that there is something about PGSACWO which makes it different from all of the other variables which I am using in the DDT and are being written out.  I believe it should be bit-for-bit with the CAM6 run, and I'm trying to figure out why it is not.  The last thing I need to do is add the output for the additional fields which we added along the way and make my runs with them.  Perhaps one or more of them will have an issue as well and that will yield some insight. That is my plan when I have time to get back onto working on this.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1935,6 +1836,16 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
 
    nan_array = nan
 
+   ! Allocate the proc_rates DDT
+   ! IMPORTANT NOTE -- elements in proc_rates are dimensioned to the nlev dimension while
+   !     all the other arrays in this routine are dimensioned pver.  This is required because
+   !     PUMAS only gets the top_lev:pver array subsection, and the proc_rates arrays
+   !     need to be the same levels.
+   call proc_rates%allocate(ncol,nlev, errstring)
```
sjsprecious, line 1844: I am curious whether there is a reference in the CAM code or documentation that describes the difference between `pcols`, `ncol` and `psetcols`. Same question for `pver`, `nlev` and `top_lev` here. Thanks!

sjsprecious (approved):
Thanks @cacraigucar . The PR looks good to me, as long as the PUMAS tag is updated later.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1485,7 +1463,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    real(r8), pointer :: lambdac(:,:)      ! Size distribution slope parameter for radiation
    real(r8), pointer :: des(:,:)          ! Snow effective diameter (m)
    real(r8), pointer :: degrau(:,:)       ! Graupel effective diameter (m)
-   real(r8), pointer :: bergso(:,:)       ! Conversion of cloud water to snow from bergeron
+   real(r8), pointer :: bergstot(:,:)       ! Conversion of cloud water to snow from bergeron
```
nusbaume, line 1466: Re-align the comment here with those above?

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2117,6 +2032,12 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
          call pbuf_get_field(pbuf, rate1_cw2pr_st_idx, rate1ord_cw2pr_st_grid)
       end if
 
+   else
+      allocate (bergso_grid(pcols,pver), stat=ierr)
```
nusbaume, line 2036: I might remove the space between `allocate` and `(` here.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -549,11 +547,23 @@ end subroutine micro_pumas_cam_readnl
...
    logical :: use_subcol_microp  ! If true, then are using subcolumns in microphysics
+   logical :: found
+
+   integer :: i, ierr
+   real(r8) :: all_levs(pver)
+
+   allocate(trop_levs(pver-top_lev+1), stat=ierr)
+
+   if (ierr /= 0) then
+         call endrun( "micro_pumas_cam_register: error allocating trop_levs")
```
nusbaume, line 565: Instead of using `endrun` here, it might be better to use ` handle_allocate_error` from `cam_abortutils`.  The same goes for the other allocate stat checks in this module as well.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -3034,59 +2839,31 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
...
-         drout2_grid(:ngrdcol,top_lev:) = avg_diameter( &
-              qrout_grid(:ngrdcol,top_lev:), &
-              nrout_grid(:ngrdcol,top_lev:) * rho_grid(:ngrdcol,top_lev:), &
-              rho_grid(:ngrdcol,top_lev:), rhow)
-
-         reff_rain_grid(:ngrdcol,top_lev:) = drout2_grid(:ngrdcol,top_lev:) * &
-              1.5_r8 * 1.e6_r8
-      end where
+      reff_rain_grid(:ngrdcol,top_lev:) = drout2_grid(:ngrdcol,top_lev:) * &
+           1.5_r8 * 1.e6_r8
```
nusbaume, line 2851: It looks like the `1.e6_r8` is doing some sort of unit conversion?  If so then it might be worth making it a parameter to help improve readability.

I have the same question/suggestion for `reff_snow_grid` below as well.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -3608,6 +3380,14 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       call outfld("TROPF_RHADJ", cp_rh,       pcols, lchnk)
    end if
 
+   ! deallocate the temporary pbuf grid variable which was allocated if subcolumns are not used
+   if (.not. use_subcol_microp) then
+      deallocate (bergso_grid)
```
nusbaume, line 3385: I might remove the space between `deallocate` and `(` here, as apparently it semi-confuses Github's syntax highlighter.

nusbaume (changes requested):
I realize it might be awhile before this PR is merged, but I wanted to add my questions/comments in now just in case.  Hopefully it all makes sense!

Externals_CAM.cfg
```diff
@@ -37,7 +37,7 @@ required = True
 local_path = src/physics/pumas
 protocol = git
 repo_url = https://github.com/ESCOMP/PUMAS
-tag = pumas_cam-release_v1.27
+tag =   pumas_cam-release_v1.28
```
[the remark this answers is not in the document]
cacraigucar: removed

src/control/cam_history_support.F90
```diff
@@ -1959,4 +1965,114 @@ end function hist_dimension_name
...
+     integer, optional, intent(in)  :: istart
+     integer, optional, intent(in)  :: istop
+     logical, optional, intent(out) :: found
+     ! Local variables
+     integer :: indx, jndx, rndx
+     integer :: ibeg
+     integer :: iend
+     logical :: dim_ok
+
+     dim_ok = .false.
```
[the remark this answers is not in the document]
cacraigucar: added

src/control/cam_history_support.F90
```diff
@@ -1959,4 +1965,114 @@ end function hist_dimension_name
...
+              if (present(istart)) then
+                 ibeg = istart
+              else
+                 ibeg = LBOUND(hist_coords(indx)%real_values, 1)
+              end if
+              if (present(istop)) then
+                 iend = istop
+              else
+                 iend = UBOUND(hist_coords(indx)%real_values, 1)
+              end if
```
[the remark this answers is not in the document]
cacraigucar: added

src/control/cam_history_support.F90
```diff
@@ -1959,4 +1965,114 @@ end function hist_dimension_name
...
+
+  !#######################################################################
+
+  subroutine hist_dimension_values_i(name, ivalues, istart, istop, found)
+     ! Given the name of a dimension, return its (integer) values in <ivalues>
+     ! If <istart> and <istop> are present, they are the beginning and ending
+     ! indices of the dimension values to return in <ivalues>. By default,
+     ! the entire array is copied.
+     ! If <found> is passed, return .true. if <name> is a defined dimension
+     !       with real values.
```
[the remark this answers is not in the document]
cacraigucar: fixed

src/control/cam_history_support.F90
```diff
@@ -1959,4 +1965,114 @@ end function hist_dimension_name
...
+     integer, optional, intent(in)  :: istart
+     integer, optional, intent(in)  :: istop
+     logical, optional, intent(out) :: found
+     ! Local variables
+     integer :: indx, jndx, rndx
+     integer :: ibeg
+     integer :: iend
+     logical :: dim_ok
+
+     dim_ok = .false.
```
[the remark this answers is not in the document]
cacraigucar: added

src/control/cam_history_support.F90
```diff
@@ -1959,4 +1965,114 @@ end function hist_dimension_name
...
+              if (present(istart)) then
+                 ibeg = istart
+              else
+                 ibeg = LBOUND(hist_coords(indx)%integer_values, 1)
+              end if
+              if (present(istop)) then
+                 iend = istop
+              else
+                 iend = UBOUND(hist_coords(indx)%integer_values, 1)
+              end if
```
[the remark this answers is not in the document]
cacraigucar: added

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -298,13 +303,6 @@ subroutine micro_pumas_cam_readnl(nlfile)
 
      ! Verify that version numbers are valid.
      select case (micro_mg_version)
-     case (1)
-        select case (micro_mg_sub_version)
-        case(0)
-           ! MG version 1.0
-        case default
-           call bad_version_endrun()
-        end select
```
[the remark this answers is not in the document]
cacraigucar: Yes, @andrewgettelman confirmed that MG1 would be phased out with this tag moving forward with CAM

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -570,18 +580,29 @@ subroutine micro_pumas_cam_register
...
-           longname='Grid box averaged rain amount', is_convtran1=.true.)
-      call cnst_add(cnst_names(6), mwh2o, cpair, 0._r8, ixsnow, &
-           longname='Grid box averaged snow amount', is_convtran1=.true.)
-      call cnst_add(cnst_names(7), mwh2o, cpair, 0._r8, ixnumrain, &
-           longname='Grid box averaged rain number', is_convtran1=.true.)
-      call cnst_add(cnst_names(8), mwh2o, cpair, 0._r8, ixnumsnow, &
-           longname='Grid box averaged snow number', is_convtran1=.true.)
+   ! Add history coordinate for DDT nlev
+   call hist_dimension_values('lev',all_levs, 1, pver, found)
+   trop_levs(1:pver-top_lev+1) = all_levs(top_lev:pver)
```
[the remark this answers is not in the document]
cacraigucar: moved

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -965,75 +954,72 @@ subroutine micro_pumas_cam_init(pbuf2d)
...
    call addfld ('PRODPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of conversion of condensate to precip'              )
    call addfld ('EVAPPREC',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling precip'                   )
-   call addfld ('EVAPSNOW',   (/ 'lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
+   call addfld ('EVAPSNOW',   (/ 'trop_cld_lev' /), 'A', 'kg/kg/s',  'Rate of evaporation of falling snow'                     )
    call addfld ('HPROGCLD',   (/ 'lev' /), 'A', 'W/kg'    , 'Heating from prognostic clouds'                          )
    call addfld ('FICE',       (/ 'lev' /), 'A', 'fraction', 'Fractional ice content within cloud'                     )
    call addfld ('CLDFSNOW',   (/ 'lev' /), 'A', '1',        'Cloud fraction adjusted for snow'                        )
    call addfld ('ICWMRST',    (/ 'lev' /), 'A', 'kg/kg',    'Prognostic in-stratus water mixing ratio'                )
    call addfld ('ICIMRST',    (/ 'lev' /), 'A', 'kg/kg',    'Prognostic in-stratus ice mixing ratio'                  )
```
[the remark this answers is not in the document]
cacraigucar: The reason for trop_cld_lev is because of the way PUMAS is called.  It was not introduced to save disk space, but was rather to save computation time by not having unnecessary arrays copies happening.  The arrays passed to PUMAS are subsetted on the fly to (:ncol,top_lev:).  As a result all of the arrays inside the new DDT are allocated to these dimensions so that internally in PUMAS all arrays are the same size.  We discussed allocating arrays which covered the entire pver and copying the DDT arrays into the full-sized arrays after PUMAS has been called so they could be output, but this was deemed an unnecessary waste and trop_cld_lev was introduced.  There are various reasons why the other arrays are still dimensioned "lev".  These include that all other CAM output variables are output on "lev" and that some of the fields reside in the pbuf and need to be stored in the "lev" dimension.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1435,8 +1413,7 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    use micro_pumas_utils, only: mg_liq_props, mg_ice_props, avg_diameter
    use micro_pumas_utils, only: rhoi, rhosn, rhow, rhows, rhog, qsmall, mincld
 
-   use micro_mg1_0, only: micro_mg_tend1_0 => micro_mg_tend
-   use micro_pumas_v1, only: micro_pumas_tend => micro_pumas_tend
+   use micro_pumas_v1,    only: micro_pumas_tend => micro_pumas_tend
```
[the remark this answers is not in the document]
cacraigucar: changed

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: aligned

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: done

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: @nusbaume - I have not made this change.  I do not see `handle_allocate_error` being used anywhere inside of CAM yet.  Is this a new CAM "best practice"?  I have mild concerns about using it as it is essentially duplicating what I am doing with my if/endrun constructs, but has the overhead of calling a subroutine.  I'd like to discuss this request further.

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: @Katetc - Since this was in the original code, I don't know if it is indeed a unit conversion and what an appropriate name for it should be.

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: done

src/physics/cam_dev/micro_pumas_cam.F90
nusbaume: @cacraigucar I saw it being used in a PR recently, but sadly I can't remember which one, although it did make me realize that it exists in the first place.

In terms of my reasoning for using `handle_allocate_error`, they are as follows:

1.  It involves a single line of code instead of three for every check.
2.  It gets the CAM code slightly closer to CAMDEN (which has it's own `check_allocate` call).  
3.  The `handle_allocate_error` outputs the error code itself, which can be useful for debugging but is frequently not done in our `endrun` calls.

It is true that the extra subroutine call adds some overhead, but since this is during the model initialization phase I figured we could afford the hit.  Plus `allocate` itself is quite an expensive operation, so I don't think adding one more subroutine call is going to make much of a difference.

Of course if you still feel like the cons outweigh the pros just let me know.  Thanks!

cacraigucar: @fvitt and @peverwhee - If we are updating the CAM standards to use this subroutine, it probably makes sense for you to weigh in on this as well.  Are you both okay with this?

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
nusbaume: Thanks for the explanation @cacraigucar!  Obviously for variables that are in `pbuf`, or that have another code-based reason for them to keep a "lev" dimension then I am happy with them staying as-is.  

However, if that reason doesn't exist for a particular variable in this list then I might still vote to move it to `trop_cld_lev`.  This is because running out of disk space is a constant problem for a lot of scientists, and this could help potentially reduce that burden even if it wasn't the original purpose.  Plus there doesn't seem to be much benefit wasting time writing out what I imagine are just a bunch of zeros above `trop_cld_lev` anyways.

Maybe it would be good to get @peverwhee's opinion as well, as I don' want to make future work for one of us if we collectively don't think it is worth it.

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: Since I've not heard dissent from @fvitt or @peverwhee, I will go ahead and implement the suggestion as @nusbaume has requested

cacraigucar: added to allocate calls in this module

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
peverwhee: I agree with @nusbaume that this could be worth it. Unfortunately, I don't know enough to know what may preclude a variable from being moved to trop_cld_lev (other than being in pbuf). Maybe we need an issue for future work since moving all of the variables we think could be moved could cause unknown problems...

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2840,60 +2642,58 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       qme_grid        => qme
       nevapr_grid     => nevapr
       prain_grid      => prain
-      bergso_grid     => bergso
 
+      bergso_grid(:ncol,top_lev:)    =  proc_rates%bergstot
```
peverwhee, line 2646: should the bergso_grid variable name be changed to bergstot_grid to match these other changes?

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2032,7 +1940,11 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    call pbuf_get_field(pbuf, sadsnow_idx,     sadsnow,     col_type=col_type)
    call pbuf_get_field(pbuf, wsedl_idx,       wsedl,       col_type=col_type)
    call pbuf_get_field(pbuf, qme_idx,         qme,         col_type=col_type)
-   call pbuf_get_field(pbuf, bergso_idx,      bergso,      col_type=col_type)
+   call pbuf_get_field(pbuf, bergso_idx,      bergstot,    col_type=col_type)
```
peverwhee, line 1943: should bergso_idx be bergstot_idx?

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2840,60 +2642,58 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
...
+      mnuccco_grid(:ncol,top_lev:)    = proc_rates%mnuccctot
+      mnuccto_grid(:ncol,top_lev:)    = proc_rates%mnuccttot
+      bergo_grid(:ncol,top_lev:)      = proc_rates%bergtot
+      homoo_grid(:ncol,top_lev:)      = proc_rates%homotot
+      msacwio_grid(:ncol,top_lev:)    = proc_rates%msacwitot
+      psacwso_grid(:ncol,top_lev:)    = proc_rates%psacwstot
+      cmeitot_grid(:ncol,top_lev:)    = proc_rates%cmeitot
+      qireso_grid(:ncol,top_lev:)     = proc_rates%qirestot
+      prcio_grid(:ncol,top_lev:)      = proc_rates%prcitot
+      praio_grid(:ncol,top_lev:)      = proc_rates%praitot
```
peverwhee, line 2666: same with these grid variable names? should they be changed to qcrestot_grid, etc?

peverwhee (changes requested):
a few minor change requests

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -549,11 +547,21 @@ end subroutine micro_pumas_cam_readnl
...
    !-----------------------------------------------------------------------
 
    logical :: prog_modal_aero
    logical :: use_subcol_microp  ! If true, then are using subcolumns in microphysics
+   logical :: found
+
+   integer :: i, ierr
+   real(r8) :: all_levs(pver)
+
+   allocate(trop_levs(pver-top_lev+1), stat=ierr)
```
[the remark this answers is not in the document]
cacraigucar: I experimented with this, and I believe that the deallocate can only happen at the end of the run (which is unnecessary at that point).  I tried deallocating at the end of the register routine and the numbers written out in the file were incorrect.   Upon further investigation, this variable appears to require being declared at the module level with the target attribute.  It appears that the values are not being stored locally when add_vert_coord is called, but rather depend on the values stored in this module.

src/physics/cam_dev/micro_pumas_cam.F90
Katetc: It is a unit conversion, it's going from meters to microns. Maybe a unit name like "mu2m" or "micron2meter"? I don't have a strong opinion on the actual name of the variable.

Katetc: And that only applies to the 1.e6. The 1.5 is, I think, a radius-diameter conversion of some sort... Let's call it a shape parameter.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -570,18 +578,29 @@ subroutine micro_pumas_cam_register
...
-           longname='Grid box averaged rain number', is_convtran1=.true.)
-      call cnst_add(cnst_names(8), mwh2o, cpair, 0._r8, ixnumsnow, &
-           longname='Grid box averaged snow number', is_convtran1=.true.)
+   ! Add history coordinate for DDT nlev
+   call hist_dimension_values('lev',all_levs, 1, pver, found)
+
+   if (found) then
+      trop_levs(1:pver-top_lev+1) = all_levs(top_lev:pver)
+      call add_vert_coord('trop_cld_lev', pver-top_lev+1,                          &
+            'tropopshere hybrid level at midpoints (1000*(A+B))', 'hPa', trop_levs,  &
```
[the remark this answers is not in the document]
Katetc: Hee hee hee. Tro pops here.

src/physics/cam_dev/micro_pumas_cam.F90
Katetc: So the "tot" nomenclature indicated the final value after all sub steps and removing any sub grid variability (aka grid averaged). The _grid nomenclature meant a similar thing but after the sub columns had been officially averaged. Or, a value that was never put on sub columns. So, to me, having "tot_grid" is kind of redundant. I think it's fine to have "tot" in the name of the DDT because that is the internal nomenclature, and it's fine to have "_grid" in the interface because that's the interface signifier. I wouldn't put both together though. "cmeiout_grid" was fine.

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
peverwhee: ok thanks for the clarification!

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
Katetc: So, it looks like the mg1 options are still in configure . I don't think that removing this code is all that is needed to phase out mg1. I think that removing the option to run mg1 is a bigger task and should have its own issue. I think this code should remain to support the configure options that are still out there.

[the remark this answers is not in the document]
cacraigucar: The problem is that MG1 can not be run (or even compile) now that the DDT has been introduced (which is why I asked Andrew if I could remove it).  I would like to not reintroduce this MG1 code until we clean it up completely.  The variables which are in the DDT no longer exist as standalone variables.  In the short-term should we introduce a check/error if MG1 is selected?  @Katetc - what are your thoughts?

src/physics/cam/micro_pumas_cam.F90
```diff
@@ -1324,7 +1324,13 @@ subroutine micro_pumas_cam_init(pbuf2d)
       call pbuf_set_field(pbuf2d, bergso_idx, 0._r8)
       call pbuf_set_field(pbuf2d, icswp_idx, 0._r8)
       call pbuf_set_field(pbuf2d, cldfsnow_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, dei_idx,     0.0_r8)
+      call pbuf_set_field(pbuf2d, des_idx,     0.0_r8)
+      call pbuf_set_field(pbuf2d, mu_idx,     0.0_r8)
+      call pbuf_set_field(pbuf2d, lambdac_idx, 0.0_r8)
 
+      if (degrau_idx > 0) call pbuf_set_field(pbuf2d, degrau_idx, 0.0_r8)
+      if (icgrauwp_idx > 0) call pbuf_set_field(pbuf2d, icgrauwp_idx, 0.0_r8)
```
nusbaume, line 1333: If this is to fix a `cam_dev` specific issue then why is it being added to the CAM6 version?

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2840,60 +2645,58 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       qme_grid        => qme
       nevapr_grid     => nevapr
       prain_grid      => prain
-      bergso_grid     => bergso
 
+      bergso_grid(:ncol,top_lev:)    =  proc_rates%bergstot
```
Katetc, line 2655: I feel dumb, but I just don't understand the special treatment for bergso_grid here. It wasn't treated any differently before...

Katetc (changes requested):
My only issue is that I don't understand what is up with bergso_grid. Other than that, it all looks great!!

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
nusbaume: The relevant issue for this request can now be found in issue #715.

nusbaume (approved):
Looks good to me!  Just a couple final requests to add some test descriptions, and one question related to a cam-dev change.  Thanks!

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1516,7 +1516,7 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-  <test compset="FCnudged" grid="ne0CONUSne30x8_ne0CONUSne30x8_mt12" name="SMS_D_Ln9_Vnuopc" testmods="cam/outfrq9s_refined_camchem">
+  <test compset="FCnudged" grid="ne0CONUSne30x8_ne0CONUSne30x8_mt12" name="SMS_D_Ln9_Vnuopc_P720x1" testmods="cam/outfrq9s">
```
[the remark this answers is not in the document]
cacraigucar: @nusbaume - I just changed the processor count and changed the testmods as @fvitt suggested in #708.  I do not know what this is testing exactly as I did not introduce the test.  I would suggest that test descriptions be added when the testing list is updated?

src/physics/cam/micro_pumas_cam.F90
cacraigucar: @bstephens82 and/or @adamrher  - Can you answer this question since this is coming in from the code you provided in #703

bstephens82: Hmm good question...Adam might have a better answer, but I think I assumed it was just good practice to initialize these variables to zero regardless of the version, even if the problem shows up in cam_dev.

adamrher: @nusbaume the reason I did this for both cam_dev and cam is because it is uncontroversial that pbuf variables should be initialized, and it is required to be initialized in cam_dev. The cam practice of not initializing pbuf variables just because they happen to be set in the time loop before they are used, is a bad habit IMO (and basically why cam_dev was so hard to implement). The only question is then whether 0 is an appropriate initialization, whether these pbuf variables gets divided into anything and therefore creating divide by zeros. We've checked they don't.

adamrher: i don't mean to be so argumentative btw. If you guys think we shouldn't touch cam6 at all than I defer to yall.

nusbaume: @adamrher Thanks for the explanation!  I agree with your analysis, I mostly just wanted to make sure that this modification wasn't there on accident.  I am happy to leave it as-is unless it we find that it substantially changes answers in the regression tests (which would be really weird, but you never know).  I'll go ahead and mark this as resolved.

cime_config/testdefs/testlist_cam.xml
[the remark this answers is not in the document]
nusbaume: @fvitt can you provide a short one sentence description of what model configuration this is testing?  

@cacraigucar Although I imagine we will certainly try to add in test descriptions when we do the test list update, adding descriptions in when we can beforehand will make that (big) job a little easier, and will save us from pinging Francis and other people a bunch of times when that development is ongoing.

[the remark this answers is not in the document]
fvitt: Nudged CAM-Chem on a regionally refined grid

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1622,7 +1622,7 @@
       <machine name="cheyenne" compiler="intel" category="camchem"/>
     </machines>
   </test>
-  <test compset="FCHIST" grid="ne0CONUSne30x8_ne0CONUSne30x8_mt12" name="SMS_D_Ln9_Vnuopc" testmods="cam/outfrq9s_refined_camchem">
+  <test compset="FCHIST" grid="ne0CONUSne30x8_ne0CONUSne30x8_mt12" name="SMS_D_Ln9_Vnuopc_P720x1" testmods="cam/outfrq9s">
```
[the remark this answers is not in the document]
fvitt: Free running CAM-Chem on a regionally refined grid

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2117,6 +2032,10 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
          call pbuf_get_field(pbuf, rate1_cw2pr_st_idx, rate1ord_cw2pr_st_grid)
       end if
 
+   else
+      allocate(bergso_grid(pcols,pver), stat=ierr)
```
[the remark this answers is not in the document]
cacraigucar: See the comment below at line 2655.

src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -2500,13 +2271,13 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
               qgout2(:ncol,top_lev:), ngout2(:ncol,top_lev:), dgout2(:ncol,top_lev:), freqg(:ncol,top_lev:),   &
               freqs(:ncol,top_lev:),           freqr(:ncol,top_lev:),           &
               nfice(:ncol,top_lev:),           qcrat(:ncol,top_lev:),           &
+              proc_rates,                                                       &
```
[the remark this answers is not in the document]
cacraigucar: I aim to please!

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: The problem is that all elements in proc_rates are dimensioned to be (ncol,nlev) so that they can be used directly in PUMAS which can not handle (pcols,nlev).  The pointer logic "assignment" does not work with the elements in proc_rates.  `bergso_grid` is used in subcol and may be acquired directly from the pbuf, so that is why it is conditionally allocated above.  In the old code with the pointer assignment, bergso_grid did not need an allocation.

The reason the other fields do not need to be allocated is because none of them were using the pointer assignments and were "allocated" as follows:
    evpsnow_st_grid is always acquired in a pbuf_get_field
   The other _grid fields were dimensioned to be (pcols,nlev) when they are declared.

This was one of the "gotchas" that made this PR so much fun!

Let me know if you need further explantation

src/physics/cam_dev/micro_pumas_cam.F90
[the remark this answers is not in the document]
cacraigucar: We don't want to remove mg1 from the configure script as it is still supported in cam, just not in cam_dev.  Since cam is using PUMAS-frozen, the DDT work was not brought into it.  I should have been a little more clear, MG1 is not supported in cam_dev from this tag forward.

src/physics/cam_dev/micro_pumas_cam.F90
Katetc: Ok, this is still confusing, but I can let it go for now. I made a note on issue #715 which is not directly related but somewhat adjacent. It at least mentions the need to clarify and simplify the variable array dimensionality. I'd like to take a closer look at this in the future, if possible.

Katetc (approved):

PR #636, opened 2022-08-09
CAM answer changing bug fix for thermodynamics

### LINKED ISSUE #633: part of cam_cesm2_2_rel_04: Bug in spectral-element advection
### Issue Type

Other (please describe below)

### Issue Description

Bug in spectral-element advection. See details here 

https://github.com/NCAR/amwg_dev/issues/151

Source code modifications to fix bug:

In src/dynamics/se/dycore/prim_advance_mod.F90. replace

               elem(ie)%derived%dpdiss_ave(i,j,k)=elem(ie)%derived%dpdiss_ave(i,j,k)+&
                     rhypervis_subcycle*eta_ave_w*(elem(ie)%state%dp3d(i,j,k,nt)

with

                elem(ie)%derived%dpdiss_ave(i,j,k)=elem(ie)%derived%dpdiss_ave(i,j,k)+&
                     rhypervis_subcycle*eta_ave_w*(elem(ie)%state%dp3d(i,j,k,nt)-elem(ie)%derived%dp_ref(i,j,k))

Source code modification here:

/glade/u/home/pel/src/for-cecille/qdp_ref_bufix/prim_advance_mod.F90

### Will this change answers?

All SE tests  with topography should show significant differences  (round-off or no differences for non-topo  tests such as FKESSLER and QPC). Tests with all other dynamical cores should not show differences.

Yes

### LINKED ISSUE #634: SSTICE_GRID_FILENAME and SSTICE_DATA_FILENAME do not belong in config_compsets.xml
### Issue Type

Code Clean-up

### Issue Description

SSTICE_GRID_FILENAME does not belong in nuopc configuraitons.
In addition, I don't understand why these files are here and not in docn. 
I am introducing a new xml variable in cdeps/docn/cime_config which will be SSTICE_MESH_FILENAME that will contain mesh specification for different ocn stream data resolutions. This will replace the SSTICE_GRID_FILENAME and will be used also by cice6 (and cice5).
I am happy to go into more details if that would be helpful.
I will not be implementing the fix in cam - but will be implementing the new capability in docn and cice.

### Will this change answers?

No

No

### LINKED ISSUE #635: CAM Thermo answer-changing bug fixes and code clean-up
### What happened?

There were two change requests made for the CAM Thermo PR (#615) that would fix a bug and help clean-up the code, but which would likely result in answer changes.  Thus it would be great to apply these changes into a PR that is already known to be answer-changing.  The two requested changes are:

1.  In the subroutine `get_Richardson_number` in `physconst.F90`, the loop:

     `do k=nlev-1,2,-1`

      Should be replaced with:

      `do k=nlev,2,-1`

      Otherwise the Richardson number will be undefined at nlev.

2.  In the subroutine `get_molecular_diff_coef` in `physconst.F90`, there are multiple locations where variables are being multiplied by `1.e-7_r8` and `1.e-5_r8`, which are actually part of the variable values themselves.  Thus it would be better if these values were multiplied during the original initialization of these variables during model initialization, as opposed to every time step like is being done here.  Note that this will likely involve modification both of `thermodynamic_active_species_kv` and `thermodynamic_active_species_kc`, as well as of the `kv#` and `kc#` parameters.
 

Not sure, but if you just look at the source code you can hopefully spot it.

cam6_3_071

Other (please explain below)

Other (please specify below)

_No response_

No

### Extra info

I didn't actually run the model to examine these bugs, I just found them during  code review.

### LINKED ISSUE #638: ERP_Ld3 test fails COMPARE_base_rest
### What happened?

This restart test fails
ERP_Ld3_Vmct.f19_f19_mg17.QPC4.cheyenne_intel.cam-outfrq3d

This fails in tags cam6_3_071 and cam6_3_007.  However, it passes in cam6_3_006.  So the changes that went into cam6_3_007 seem to be the culprit which are "Weak scaling infrastructure and bug fixes" (by @gold2718).

qcmd -- create_test --walltime 00:15:00 --test-id myid ERP_Ld3_Vnuopc.f19_f19_mg17.QPC4.cheyenne_intel.cam-outfrq3d

cam6_3_007 - cam6_3_071

CISL machine (e.g. cheyenne)

Intel

/glade/scratch/fvitt/ERP_Ld3_Vmct.f19_f19_mg17.QPC4.cheyenne_intel.cam-outfrq3d.cam6_3_007

No

### Extra info

_No response_

cacraigucar:
Gather the answer changing cleanups from the thermodynamics PR.

Closes #635 
Closes #638


src/utils/physconst.F90
```diff
@@ -168,13 +168,13 @@ module physconst
...
 ! kc1,..,kc4 are coefficients for kmcnd calculation
 real(r8), parameter :: &
-   kv1 = 4.03_r8, &
-   kv2 = 3.42_r8, &
-   kv3 = 3.9_r8,  &
-   kv4 = 0.69_r8, &
-   kc1 = 56._r8,  &
-   kc2 = 56._r8,  &
-   kc3 = 75.9_r8, &
+   kv1 = 4.03_r8 * 1.e-7_r8, &
```
nusbaume, line 171: I think we also need to remove these scaling factors from `get_molecular_diff_coef` and `get_molecular_diff_coef_reference` as well.  Otherwise they'll be double-counted.

nusbaume (changes requested):
Impressively quick turnaround!  However, I still think we need to make additional modifications in `physconst.F90` in order to avoid double-counting.

fvitt (commented):
Can you explain the need for the changes in physconst?  How do those changes affect waccmx?

cacraigucar:
> Can you explain the need for the changes in physconst? How do those changes affect waccmx?

@nusbaume can explain more.  These were answer changing items from  PR #615 that @nusbaume identified as being changed.  There is more explanation in issue #635.  The answers are supposed to be roundoff, unless I implemented something wrong.

src/utils/physconst.F90
```diff
@@ -453,8 +453,8 @@ subroutine composition_init()
           thermodynamic_active_species_cv (icnst) = 0.5_r8*shr_const_rgas*dof2/mw !N2
           thermodynamic_active_species_R  (icnst) = shr_const_rgas/mw
           thermodynamic_active_species_mwi(icnst) = 1.0_r8/mw
-          thermodynamic_active_species_kv(icnst)  = 3.42_r8
-          thermodynamic_active_species_kc(icnst)  = 56._r8
+          thermodynamic_active_species_kv(icnst)  = 3.42_r8 * 1.e-7_r8
+          thermodynamic_active_species_kc(icnst)  = 56._r8 * 1.e-5_r8
```
gold2718, line 457: I thought the point of this change was to move all the math into parameters so this should be:
```suggestion
          thermodynamic_active_species_kv(icnst)  = kv2
          thermodynamic_active_species_kc(icnst)  = kc1
```
Am I missing something? If not, similar changes should be made below (500, 501, 519, 520).

gold2718 (changes requested):
Some cleanup for consistency

src/utils/physconst.F90
nusbaume: Wow, I have no idea how I missed that, good catch!  I agree that we should replace all of these values with their respective parameters.

fvitt:
> > Can you explain the need for the changes in physconst? How do those changes affect waccmx?
> 
> @nusbaume can explain more. These were answer changing items from PR #615 that @nusbaume identified as being changed. There is more explanation in issue #635. The answers are supposed to be roundoff, unless I implemented something wrong.

Will you verify that the changes are no larger than roundoff?

nusbaume:
@fvitt @cacraigucar These changes were driven by confusion on what the `1.e-7_r8` and `1.e-5_r8` numbers were in `get_molecular_diff_coef` while reviewing the CAM thermo review (PR #615).  We were told there that these numbers were in fact part of the `kv` and `kc` parameters themselves, and so it was decided to move them to the parameter definitions instead of using them as an extra factor during the calculation of the molecular diffusivity/conductivity.

Since we are changing the order of operations, but not the equations themselves, it _should_ only have a roundoff effect.  If it doesn't then I agree that we have done something wrong here.

src/utils/physconst.F90
cacraigucar: done

src/utils/physconst.F90
cacraigucar: Done. (though I used kc2 instead of kc1).  Even though they are the same value currently, I thought it should be kc2.

src/utils/physconst.F90
```diff
@@ -1945,8 +1945,8 @@ subroutine get_molecular_diff_coef(i0,i1,j0,j1,k1,nlev,temp,get_at_interfaces,sp
 
      if (dry_air_species_num==0) then
 
-       cnst_vis = (kv1*mmro2*o2_mwi + kv2*mmrn2*n2_mwi)*mbar*1.e-7_r8
-       cnst_cnd = (kc1*mmro2*o2_mwi + kc2*mmrn2*n2_mwi)*mbar*1.e-5_r8
+       cnst_vis = (kv1*mmro2*o2_mwi + kv2*mmrn2*n2_mwi)*mbar
+       cnst_cnd = (kc1*mmro2*o2_mwi + kc2*mmrn2*n2_mwi)*mbar
```
gold2718, line 1949: What do we do about lines below such as 1955, 1956, 2020, and 2021 that used `kc4` and `kv4` without the `1.e-7_r8` factor?

src/utils/physconst.F90
```diff
@@ -2091,11 +2091,11 @@ subroutine get_molecular_diff_coef_reference(k0,k1,tref,press,sponge_factor,kmvi
        kmvis_ref(k) = sponge_factor(k)* &
             (kv1*mmro2*o2_mwi +         &
              kv2*mmrn2*n2_mwi)*mbar*    &
-             tref**kv4 * 1.e-7_r8
+             tref**kv4
```
gold2718, line 2094: You removed the scaling factor from `kv1` and `kv2`, however, `kv4` in this equation have been redefined with no change to the expression. What is the correct usage here?

src/utils/physconst.F90
```diff
@@ -2091,11 +2091,11 @@ subroutine get_molecular_diff_coef_reference(k0,k1,tref,press,sponge_factor,kmvi
...
        kmcnd_ref(k) = sponge_factor(k)* &
             (kc1*mmro2*o2_mwi +         &
              kc2*mmrn2*n2_mwi)*mbar*    &
-             tref**kc4 * 1.e-5_r8
+             tref**kc4
```
gold2718, line 2098: You removed the scaling factor from `kc1` and `kc2`, however, `kc4` in this equation have been redefined with no change to the expression. What is the correct usage here?

gold2718 (changes requested):
After a close look, it seem there were uses of some of these constants without the `1.e-7` or `1.e-5` factors. Some of the uses are even in the exponent. What is the correct usage here and why the discrepancy?

cacraigucar:
After @gold2718 pointed out that these variable are used with and without scaling, I am going to introduce a new set of variables that have scaled in the name (i.e. kc1scaled, etc.)   These will only replace the variables which have the scale factors in the equations.  The variables which do not have the scaling will be unchanged.  I will do a careful edit instead of responding to review comments.

gold2718:
> I am going to introduce a new set of variables that have scaled in the name (i.e. kc1scaled, etc.) These will only replace the variables which have the scale factors in the equations. The variables which do not have the scaling will be unchanged

Since @hanli-liu indicated that the constants are correct _with_ the scale factors (in a post on [#615](https://github.com/ESCOMP/CAM/pull/615#discussion_r934736907)), I do not think this is correct. The question is what is the role of the the "non-scaled" variables currently in physconst.F90?

fvitt:
It seems to me that the 1.e-5 and 1.e-7 scaling factors should stay where they are currently used -- just define them once as parameters.

cacraigucar:
There are currently two proposals for this PR:

1. Leave the code as it is for the most part and replace 1.e-7_r8 with scale_v_factor and 1.e-5_r8 with scale_c_factor in all the appropriate places and make these values parameters.  Note that the variable names need to be something better than what I picked.  (@fvitt's  suggestion)  
2. Set kv1=4.03_r8 * 1.e-7_r8 and kv1_unscaled=4.03_r8 and make appropriate substitutions throughout the code (modification to @nusbaume's suggestion)

@gold2718's question on why these variables are scaled sometimes and not other times, might lend some thought onto which of these two approaches would be the best implementation

nusbaume:
I would personally still vote to have the factors applied directly to all of the parameter except `kv4` and `kc4`, because those are the only parameters that are exponents as opposed to multiplicative factors.  Basically just check out equations 3 and 4 here:

https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2010JA015586

I am happy if we want to rename them from `kv4` to something like `kv_temp_exp` that indicates why they don't contain the scaling factor.  If the scaling factors are being applied inside an exponent, or the non-exponent parameters are being applied somewhere without the parameter, then my guess is that either the equations in the paper are wrong, or that is an actual physics bug.

I also am voting to apply these scaling factors in the parameters simply because it saves the literal number of multiplications that have to be done in the model.  If we leave them as an extra factor then aren't we doing extra multiplications every timestep as opposed to just once during the creation of the parameter itself?  Of course I realize this is likely not a very big cost, and maybe the compiler is smart enough to parameterize it automatically, but it just seems like we are making the code do extra work.   

However, if we find that this method simply doesn't produce round-off level differences then as an alternative I would certainly vote instead for @fvitt's suggestion, which probably can just be applied in the Thermo PR and doesn't need to be done here.

gold2718:
> ... as opposed to just once during the creation of the parameter itself?

As an FYI, the "creation of the parameter" is done at compile time and the machine value sits in code space (not data space).

fvitt:
The equations are a little tricky for me.  If @nusbaume's suggestions are no larger than roundoff than go with that.

src/utils/physconst.F90
cacraigucar: These values do not have the scaling factor when they are saved and subsequently used, so I am going to leave them as the hardcoded numbers.  I could use the scaled numbers and divide by the scaling factor if that is preferred.

nusbaume: Actually the scaling factor is applied to them.  Note that, for example, `thermodynamic_active_species_kv` is only used in one place (in the entire  code base), and that is in the calculation of `kmvis` in `get_molecular_diff_coef`.  Note then too that immediately after the loop over the different constituents the `kmvis` variable is then multiplied by the scale factor `1.e-7_r8` (which is the one we'll remove).

I might end up looking like a fool on this (in which case I owe you multiple preliminary code reviews), but I am pretty sure you just need to do the following:

1.  Add the scale factors to the parameter definitions _except_ for `kv4` and `kc4`, which might need to be renamed to indicate that they are exponents.

2.  Replace the various numbers in the setting of `thermodynamic_active_species_kv` and `thermodynamic_active_species_kc` with their respective parameter equivalent.

3.  Remove all `1.e-7_r8` and `1.e-5_r8` numbers from  `get_molecular_diff_coef` and `get_molecular_diff_coef_reference`.

The following model differences should then only be due to a (algebraically valid) change in the order of operations.

src/utils/physconst.F90
```diff
@@ -2017,8 +2019,8 @@ subroutine get_molecular_diff_coef(i0,i1,j0,j1,k1,nlev,temp,get_at_interfaces,sp
 
                temp_local = .5_r8*(temp(i,j,k-1)+temp(i,j,k))
                mbarvi = 0.5_r8*(mbarv(i,j,k-1)+mbarv(i,j,k))
-               kmvis(i,j,k) = kmvis(i,j,k)*mbarvi*temp_local**kv4*1.e-7_r8
-               kmcnd(i,j,k) = kmcnd(i,j,k)*mbarvi*temp_local**kc4*1.e-5_r8
+               kmvis(i,j,k) = kmvis(i,j,k)*mbarvi*temp_local**kv_temp_exp
+               kmcnd(i,j,k) = kmcnd(i,j,k)*mbarvi*temp_local**kc_temp_exp
```
fvitt, line 2023: What happened to the 1.e-7 and 1.e-5 factor here?

cacraigucar: The thermodynamic_active_species_kv (and _kc) now have the factors in them.  Those values are then summed in kmvis and kmcnd and the scaling was done at the end.    I believe that this is equivalent arithmetically, but if not, let me know.

fvitt: Ah okay.   Thanks for explaining.

fvitt (approved):

gold2718 (approved):
I like these changes and the answer changes seem acceptable.

jtruesdal (approved):

PR #650, opened 2022-08-31
CARMA models on unstructured grids

### LINKED ISSUE #649: Need capability to run existing CARMA models on unstructured grids

Many of the existing CARMA models are functional only on regular lat/lon grids. The ability to run the CARMRA on general unstructured physics grids is desired.

_No response_

I Don't Know

Yes

fvitt:
Fixes #649

### DIFF
--- a/cime_config/testdefs/testlist_cam.xml
+++ b/cime_config/testdefs/testlist_cam.xml
@@ -649,6 +649,15 @@
       <option name="wallclock">00:30:00</option>
     </options>
   </test>
+  <test compset="F2000climo" grid="ne30pg3_ne30pg3_mg17" name="ERS_D_Ln9_Vnuopc" testmods="cam/carma_dust">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="carma"/>
+    </machines>
+    <options>
+      <option name="comment" >CARMA dust test</option>
+      <option name="wallclock">00:30:00</option>
+    </options>
+  </test>
 <!--              carma sea_salt          -->
   <test compset="QPC5" grid="f10_f10_mg37" name="ERC_D_Ln9_Vnuopc" testmods="cam/carma_sea_salt">
     <machines>
@@ -781,7 +790,7 @@
       <option name="comment" >CARMA particle growth test</option>
       <option name="wallclock">00:30:00</option>
     </options>
-  </test>  
+  </test>
   <test compset="QPC6" grid="ne5pg3_ne5pg3_mg37" name="ERC_D_Ln9_Vnuopc" testmods="cam/carma_dust">
     <machines>
       <machine name="izumi" compiler="nag" category="carma"/>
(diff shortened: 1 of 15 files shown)


src/physics/carma/models/dust/carma_model_mod.F90
```diff
@@ -715,32 +708,36 @@ subroutine CARMA_ReadSoilErosionFactor(carma, rc)
...
-    real(r8)                                  :: lat(plat), lon(plon)
-    integer                                   :: i
+    type (interp_type)                        :: lat_wght, lon_wght
+    real(r8)                                  :: lat(pcols)            ! latitude index
+    real(r8)                                  :: lon(pcols)            ! longitude index
+    integer                                   :: i, ii
+    integer                                   :: lchnk                 ! chunk identifier
+    integer                                   :: ncol                  ! number of columns in chunk
+
+    real(r8), parameter   :: zero=0_r8, twopi=2_r8*pi, degs2rads = pi/180._r8
```
cacraigucar, line 722: I tried to find where `pi` is coming from and saw that all of the `use` statements are missing the `only` qualifiers.  This is probably far enough removed from the actual CAM code that CAM won't enforce this requirement, but I wanted to point out this deficiency in case CARMA would like to fix this issue.

Also, the only place I was able to find `pi` defined was in carma/models/cirrus_dust/carma_model_mod.F90 where it is hardcoded to 3.1416927_r8 in two places.  Again this contradicts a CAM coding standard, but again is far enough removed, that we can probably let it slide.   This practice is not recommended though.

src/physics/carma/models/dust/carma_model_mod.F90
```diff
@@ -715,32 +708,36 @@ subroutine CARMA_ReadSoilErosionFactor(carma, rc)
...
     call wrap_inq_dimlen(fid, fid_lat, f_nlat)
-  
+
     allocate(ero_lat(f_nlat))
     allocate(ero_lon(f_nlon))
     allocate(ero_factor (f_nlon, f_nlat))
-    allocate(ero_factor1(plon, plat))
-    allocate(soil_factor(plat, plon))
-    
+    allocate(soil_factor(pcols, begchunk:endchunk))
```
cacraigucar, line 739: Another CAM coding standard is to check all allocates.  Again, this code is far enough removed from CAM code, that I won't force the issue.   It could have an impact with a jobhalting and no useful error message if an allocate failed.

cacraigucar (approved):
There are several items which are not "best practices" which I've seen in the CARMA code.  Since they are in what we would consider being "the CARMA library", I am not going to force them to be changed for this PR.  I would suggest that CARMA consider a revision of their code at a future date to adopt the CAM coding standards as they prevent problems and aid with code debugging.

nusbaume (changes requested):
Looks good, just have one request to add "comments" to the new CAM-CARMA tests.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -771,6 +780,83 @@
       <option name="wallclock">00:30:00</option>
     </options>
   </test>
+  <test compset="QPC6" grid="ne5pg3_ne5pg3_mg37" name="ERC_D_Ln9_Vnuopc" testmods="cam/carma_dust">
```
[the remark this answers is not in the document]
fvitt: added

nusbaume (approved):
Looks great now, thanks!

Issue #564, opened 2022-04-11
Interpolation issue for SE dycore output? 


andrewgettelman:
I'm running an F-case with the se core on the cam_dev branch (res = ne30pg3_ne30pg3_mg17), tag is cam6_3_052 (but cam6_3_045 has the same problem)

When I set:
 interpolate_output = .true.
 interpolate_nlat = 192
 interpolate_nlon = 288

I get strange behavior in some of the output files. For some fields (AODVIS, TOT_CLD_VISTAU) but not others (SWCF) I get anomalies at high latitudes, even in a 3 year average:

![download](https://user-images.githubusercontent.com/9616874/162837476-07e76efa-ffef-48e7-b032-f3e65450a4f8.png)

Is this a known interpolation issue with the internal interpolation for CAM? It certainly looks like some imprinting at high latitudes.

Should I NOT use the interpolate function? Or just not use f09 style resolution> (I could do 180 x 360 for example).

Files are in: /glade/scratch/andrew/archive/pumasNewAcc_cam6352/atm/hist/*.h0.000[1-3]*

Thanks for any help/advice

adamrher:
Just putting this relevant info here regarding the apparent missing values at high-latitudes for these variables, from https://github.com/ESCOMP/CAM/issues/506

> @andrewgettelman this AODVIS issue that occurs polewards of about 65˚ seems to be present in all versions of the model. If you look at the AMWG diags ([here's an f09 L32 run w/ an older pre-cam_dev code base](https://webext.cgd.ucar.edu/FSCAM/AMIP_REPORTCARD/cam6_3_006.dev_FHIST_f09_f09_mg17_L32_1800pes_211105_6macmic_25nup_1a05b3alph_Nx2yrs/atm/cam6_3_006.dev_FHIST_f09_f09_mg17_L32_1800pes_211105_6macmic_25nup_1a05b3alph_Nx2yrs.1999_2017-cam6_3_006.dev_FHIST_f09_f09_mg17_L32_1800pes_211105_6macmic_cam6_Nx2yrs.1999_2017/)) the AODVIS variables seems to be set to missing values above that critical latitude. In fact, your values of order 1E34 suggests you're plotting missing values.
>
>We need to ping someone about why AODVIS is so strange at high-latitudes. And whether we need to worry about it.

andrewgettelman:
[...]

PR #658, opened 2022-09-23
Bug fixes for interpolate_output routines

### LINKED ISSUE #644: The interpolate_output namelist flag results in bad output values
### What happened?

As reported by @brianpm , a recent CAM simulation with the `ne30pg3_ne30pg3_mg17` and `ne30_g17` grids and the `interpolate_output` namelist flag(s) set to `.true.` resulted in history files with bad/missing values (history variables had O(10^-38) values).  

It should be noted that the non-interpolated history output was fine, so this is likely an issue with the interpolation itself.

Not entirely sure yet, but based on the example case below it appears that running an F2000climo compset for two days with one of the grids above and the following `user_nl_cam` settings:

```
nhtfrq = 0,-1,-1
mfilt = 1,24,24
ndens = 2,2,2
empty_htapes=.false.
interpolate_output=.true.,.true.,.true.
interpolate_nlat=192,192,192
interpolate_nlon=288,288,288

FINCL2 = 'U','V','T','Q','SOLIN','FSNT','FLUT','PRECC','PRECT'
FINCL3 = 'U','V','T','Q','SOLIN','FSNT','FLUT','PRECC','PRECT'

iradsw = -1
iradlw = -1
use_rad_dt_cosz = .false.
spectralflux = .false.
```

results in the bad history file output.  I will add a comment to this issue if I can find an even simpler setup which produces the same error.

Not sure, but will try with the latest tag and report the results here.

CISL machine (e.g. cheyenne)

Intel

/glade/work/brianpm/my_cases/test_cases/vanilla.f2000climo.ne30pg3.testinterp02

Yes

### Extra info

_No response_

nusbaume:
This Pull Request contains a few bug fixes for the `interpolate_output` option for the SE dycore.  Specifically, it allows for single-precision (e.g. `ndens = 2`) interpolated output, and allows for variables with missing values (e.g. `AODVIS`) to be output correctly in the standard monthly average (`h0`) files, at least for certain SE-CSLAM grids. 

Please note however that variables with missing values may still have "bad" values at the missing value margins when either using very high resolution grids or outputting at high frequencies (e.g. every hour or every time step).  To fix this issue completely will potentially require a much larger set of code modifications, which might be beneficial to do in conjunction with future developments (e.g. the creation of a dycore-agnostic interpolation routine), hence why I am creating this smaller PR now.

Fixes #644
Partially addresses #564 

Was reviewed in #659

### DIFF
--- a/src/control/cam_history.F90
+++ b/src/control/cam_history.F90
@@ -4261,7 +4261,6 @@ end subroutine print_active_fldlst
 
       !! Collect some field properties
       call AvgflagToString(tape(t)%hlist(f)%avgflag, tape(t)%hlist(f)%time_op)
-
       if ((tape(t)%hlist(f)%hwrt_prec == 8) .or. restart) then
         ncreal = pio_double
       else
@@ -4719,6 +4718,7 @@ end subroutine print_active_fldlst
     integer                          :: num_patches
     integer                          :: mdimsize   ! Total # on-node elements
     integer                          :: bdim3, edim3
+    integer                          :: ncreal     ! Real output kind (double or single)
     logical                          :: interpolate
     logical                          :: patch_output
     type(history_patch_t), pointer   :: patchptr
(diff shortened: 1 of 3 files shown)


peverwhee:
closing this PR - was included in cam6_3_081  ( #680 )

PR #672, opened 2022-10-10
Fix bug data reader used for WACCM-SC loss rates input 

### LINKED ISSUE #667: More restart non-bit-for-bit occurrences
### What happened?

From Keith Lindsay via email:

I have confirmed that my clone of one of Cecile's fully coupled experiments is not restarting exactly, when attempting to restart on particular dates.

The experiment restarts exactly early in the run, such as 0001-01-02 and 0001-01-03.
It also restarts exactly on 0001-12-01.
However, it does not restart exactly on 0001-12-31 or 0002-01-01.

In my tests of exact restartability, I've enabled cpl hist files (approximately) every timestep. (I need to move where history writing in the nuopc run sequence occurs to get this to happen.) So I'm getting a record of fields getting passed to and from the coupler. The first divergence of fields after a restart is in some fields sent from CAM to the coupler. Here's the RMS lines from cprnc comparing the cpl hist files:
 RMS atmImp_Faxa_rainc                7.1978E-15            NORMALIZED  4.7962E-10
 RMS atmImp_Faxa_snowc                1.5895E-16            NORMALIZED  3.5183E-09
 RMS atmImp_Sa_dens                   1.8034E-13            NORMALIZED  1.5064E-13
 RMS atmImp_Sa_pbot                   1.0914E-08            NORMALIZED  1.1155E-13
 RMS atmImp_Sa_ptem                   1.8113E-11            NORMALIZED  6.3387E-14
 RMS atmImp_Sa_shum                   1.2418E-14            NORMALIZED  1.3697E-12
 RMS atmImp_Sa_tbot                   1.8035E-11            NORMALIZED  6.3250E-14
 RMS atmImp_Sa_u                      7.9879E-13            NORMALIZED  1.5167E-13
 RMS atmImp_Sa_v                      4.5647E-12            NORMALIZED  9.7223E-13
 RMS atmImp_Sa_z                      9.1689E-12            NORMALIZED  1.4462E-13

So the differences are pretty small. But they grow.

This experiment has a number of CAM forcings with type CYCLICAL. Given that the problem occurs at the end of the year, but not only on Jan 1, my initial hypothesis is that there is a roundoff difference after restart in one of these forcing fields. That is, how the interpolation weights in time across the year boundary are computed.

I don't have high frequency CAM output enabled, yet, so I'm currently only able to diagnose differences in the fields the CAM sends to the coupler.

Given what I'm seeing, I'm guessing that this is reproducible in an F compset, and probably at lower resolution.

Do you have a preference on how I, or we, might proceed on further investigating this?

See above description

branch which Cecile is using

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

This issue will be evolving and becoming more specific as we gain more information

fvitt:
Fixes #667


cime_config/testdefs/testlist_cam.xml
```diff
@@ -2393,6 +2393,22 @@
...
+    </options>
+  </test>
+  <test compset="FWsc2000climo" grid="f10_f10_mg37" name="ERP_Ld3_Vnuopc" testmods="cam/outfrq1d_14dec">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="waccm"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:10:00</option>
+    </options>
+  </test>
```
nusbaume, line 2411: For both of these tests can you add a new comment option tag which describes in words what these tests are attempting to check, maybe something like:

`<option name="comment">Checks that exact restarts occur when crossing the December 14th date boundary in WACCM-SC</option>`

This will hopefully be the start of adding descriptions to our tests in this list so we'll know what's needed (or not) going forward.

nusbaume (approved):
Looks good!  Just one request to add descriptions to the new regression tests (which will hopefully be the start of a new trend).

cime_config/testdefs/testlist_cam.xml
fvitt: added

peverwhee:
reviewed and included in #680

Issue #678, opened 2022-10-25
Secondary bug in spectral-element advection


PeterHjortLauritzen:
### What happened?

The bug fix for spectral-element advection discussed in https://github.com/ESCOMP/CAM/issues/633 introduced a secondary bug identified by @jzhan166 (ACOM, NCAR). The problem can be seen in N2O (level 90hPa) as well as the uniform test tracer TT_UN (level 682hPa) (day 5 instantaneous values) when running with chemistry (CAM_CONFIG_OPTS="-phys cam_dev -microphys mg2 -age_of_air_trcs -chem trop_strat_mam4_vbs -nlev 58 -nadv 207 -nadv_tt 5"):

![Screen Shot 2022-10-25 at 9 09 06 AM](https://user-images.githubusercontent.com/8782989/197811734-48fc5ee5-70a0-48cf-989a-1e36276d0ecd.png)
![Screen Shot 2022-10-25 at 9 17 09 AM](https://user-images.githubusercontent.com/8782989/197813642-63d095e7-b35c-4ee1-99c3-eadf180f9068.png)

![Screen Shot 2022-10-25 at 9 18 40 AM](https://user-images.githubusercontent.com/8782989/197814038-acdca30e-a7c0-4796-8504-9443a3772307.png)

The problem is that when subtracting a reference profile in the hyper viscosity for spectral-element tracer advection the background reference value should be zero. This can be fixed with adding the following modifications to code in dynamics/se/dycore/prim_driver.F90 where the reference profiles are computed (the fix is subtracting dp0 from dp_ref so that the background value is zero when PHIS=0):

     do ie=nets,nete
       do k=1,nlev
         pmid_ref =hvcoord%hyam(k)*hvcoord%ps0 + hvcoord%hybm(k)*ps_ref(:,:,ie)
         dp0 = ( hvcoord%hyai(k+1) - hvcoord%hyai(k) )*hvcoord%ps0 + &
               ( hvcoord%hybi(k+1) - hvcoord%hybi(k) )*hvcoord%ps0    
         if (hvcoord%hybm(k)>0) then
           elem(ie)%derived%T_ref(:,:,k)    = T0+T1*(pmid_ref/hvcoord%ps0)**cappa
           !
           ! pel@ucar.edu: resolved noise issue over Antartica - bug not fixed in CESM2.2 release branch
           !
           elem(ie)%derived%dp_ref(:,:,k)   = elem(ie)%derived%dp_ref(:,:,k)-dp0
[...]

PR #685, opened 2022-11-03
cam6_3_084 : Secondary SE advection bug

### LINKED ISSUE #699: Increase history field length from 24 to 32
### Issue Type

Infrastructure Update

### Issue Description

copied from #686 :

This updates cam_history_support's fieldname_len parameter from 24 to 32, in order to accommodate longer names for history outputs. This is in support of implementing GEOS-Chem (https://github.com/ESCOMP/CAM/pull/484) into CESM, as we recently implemented a feature to automatically pass GEOS-Chem diagnostic quantities to CAM history. Some of the GEOS-Chem field names are longer than 24 characters and thus we need to increase the maximum length.

The model was verified to run correctly with this change so I'm making a pull request. It would be helpful to discuss here as well if the 24-character limitation was imposed due to impacts on other components, or it is artificial and can be expanded as required.

### Will this change answers?

No

No

peverwhee:
PR also contains code change from #686 
Fixes noise over Antarctica seen in spectral element advection.

TT_UN without and with fix:

Screen Shot 2022-11-03 at 11 17 24 AM

(thanks @PeterHjortLauritzen !)

Also: added a 2 month FHS94 test that would have caught this bug.

addresses #678 
closes #699


cime_config/testdefs/testlist_cam.xml
```diff
@@ -2493,6 +2493,16 @@
     </machines>
   </test>
 
+  <test compset="FHS94" grid="ne30_ne30_mg17" name="SMS_Lm2_Vnuopc" testmods="cam/se_advection">
+    <machines>
+      <machine name="cheyenne" compiler="intel" category="aux_cam"/>
+    </machines>
+    <options>
+      <option name="comment">Test for SE advection noise</option>
+      <option name="wallclock">01:00:00</option>
```
cacraigucar, line 2502: I am concerned about adding a 2 month test which takes about an hour to run to aux_cam since aux_cam tests are run pretty heavily.  Perhaps we could change the current 13 month FV 10 degree test to be an SE extremely coarse grid?  @PeterHjortLauritzen - what is the coarsest grid that SE can be run at (doesn't need to be scientifically valid)?  Also, would this very coarse grid capture this error?  If we decide to pursue this option, we'd need to see what the time impact is to making these mods.

If this is not feasible, perhaps we make this test be prealpha?  @fischer-ncar - What are your thoughts on this?  Typically we try to have aux_cam tests run to capture changes which CAM would introduce, but this wouldn't show up if that were the case until an alpha test is run.

A third alternative would be to introduce an aux_cam_periodic which we run periodically.  It wouldn't be run with every CAM tag, but run periodically (for instance every 10th CAM tag or something like that).  @peverwhee, @jtruesdal , @fvit, @nusbaume  - What are thoughts on this?  We'd reserve this tag for tests which are expensive, but we think are configurations which are pretty stable.

I'm open to any ideas on how to handle this.

cacraigucar (approved):
Assuming the test gets taken out of this PR as we discussed, I approve this PR

cime_config/testdefs/testlist_cam.xml
peverwhee: Test removed since it's taking longer than I'd hoped to verify that it isn't too costly - issue #698 opened

PeterHjortLauritzen (approved):

jtruesdal (approved):
Looks good.

PR #694, opened 2022-11-11
cam6_3_083: move emissions before CLUBB in cam_dev

### LINKED ISSUE #691: Move emissions before turbulence in cam_dev

With CAM6 and the addition of CLUBB, we degraded our long standing process ordering of having emissions prior to turbulence, and dry deposition after turbulence, in the time-loop. Instead in CAM6, turbulence precedes emissions and causes premature removal as the turbulence never has an opportunity to mix species upward prior to dry deposition. With the moving of CLUBB into tphysac in cam_dev, it is now trival to restore the correct process ordering of emissions - turbulence - deposition. I am proposing we do this.

_No response_

Yes

Yes

### LINKED ISSUE #693: part of cam6_3_083:  Bad ozone (Sa_o3) values in first coupling time step
### What happened?

In @fischer-ncar 's testing for CESM2_3_alpha10c, he noticed failures in B compsets run at f19_g17 resolution:

```
    FAIL ERS_Ld7.f19_g17.B1850.cheyenne_intel.allactive-defaultio RUN time=1817
    FAIL IRT_C3_Ld7.f19_g17.BHIST.cheyenne_intel.allactive-defaultio RUN time=799
    FAIL MCC.f19_g17.B1850.cheyenne_intel.allactive-defaultiomi RUN time=479
    FAIL MCC_Ld5.f19_g17.B1850G.cheyenne_intel.allactive-cism-test_coupling RUN time=486
    FAIL SMS_Ld5.f19_g17.B1850G.cheyenne_intel.allactive-cism-test_coupling RUN time=1819
```

These all fail in the same way (note that there is a mistake in the text of the error message: "output from CLM" should really be "input to CLM" here):

```
120: # of NaNs =            1
120: Which are NaNs =  F F F F F F F F F F F F F F F F F F F F T F F F F F F F F F F
120: F F F F F F F F F
120: NaN found in field Sa_o3 at gridcell index           21
120: ERROR:  ERROR: One or more of the output from CLM to the coupler are NaN
120:Image              PC                Routine            Line        Source            
120:cesm.exe           00000000042381C6  Unknown               Unknown  Unknown
120:cesm.exe           0000000003DB3657  shr_abort_mod_mp_         114  shr_abort_mod.F90
120:cesm.exe           0000000002929B4C  lnd_import_export         164  lnd_import_export_utils.F90
120:cesm.exe           0000000002924D82  lnd_import_export        1083  lnd_import_export.F90
120:cesm.exe           0000000002921C72  lnd_import_export         539  lnd_import_export.F90
120:cesm.exe           0000000002916D2B  lnd_comp_nuopc_mp         810  lnd_comp_nuopc.F90
```

The relevant change in this alpha tag is the inclusion of an updated CTSM: starting with `ctsm5.1.dev108`, CTSM imports the Sa_o3 field from the atmosphere. All CTSM coupling fields (input and output) are checked for NaN values in each time step, so now CTSM is aborting due to detecting a NaN value.

By removing the abort, I have verified that there is only a single grid cell with a NaN value, and only in the first coupling time step. However, by writing out cpl hist values at each time step (see steps to reproduce below) I see that the Sa_o3 values seem bad everywhere in the first coupling time step: they are 0 over most grid cells, but NaN over 3 grid cells. These are the values received from CAM's initialization. **So it seems that CAM is not properly setting the ozone coupling field in initialization, at least in this case that (I think) uses a prescribed ozone field rather than prognostic chemistry** (in contrast to other coupling fields, which seem to get valid values in model initialization).

The ideal solution would be for CAM to send the actual, correct ozone field in model initialization. If that is too difficult, can it at least send something like a reasonable, globally-constant value in initialization so that CTSM has something reasonable to work with in its first time step?

1. Checkout the latest CESM master: ESCOMP/CESM@bf1bf7f

2. Create, build and run a short test case with compset `B1850` and res `f19_g17` (e.g., one of the above tests or via create_newcase). Observe the failure.

3. To see more information, do the following:

Introduce this diff in components/clm to allow the model to continue past the initial abort:

```diff
diff --git a/src/cpl/nuopc/lnd_import_export.F90 b/src/cpl/nuopc/lnd_import_export.F90
index 0e7a5e2ee..8dccf3ddc 100644
--- a/src/cpl/nuopc/lnd_import_export.F90
+++ b/src/cpl/nuopc/lnd_import_export.F90
@@ -510,6 +510,8 @@ contains
     call NUOPC_ModelGet(gcomp, importState=importState, rc=rc)
     if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
+    write(iulog,*) 'WJS: lnd_import: nstep = ', get_nstep()
+
     ! Set bounds
     begg = bounds%begg; endg=bounds%endg
 
diff --git a/src/cpl/utils/lnd_import_export_utils.F90 b/src/cpl/utils/lnd_import_export_utils.F90
index 032cb19b6..495fcd8b9 100644
--- a/src/cpl/utils/lnd_import_export_utils.F90
+++ b/src/cpl/utils/lnd_import_export_utils.F90
@@ -161,7 +161,7 @@ contains
              write(iulog,*) "NaN found in field ", trim(fname), ' at gridcell index ',begg+i-1
           end if
        end do
-       call shr_sys_abort(' ERROR: One or more of the output from CLM to the coupler are NaN ' )
+       ! call shr_sys_abort(' ERROR: One or more of the output from CLM to the coupler are NaN ' )
     end if
   end subroutine check_for_nans
 ```

Change cplhist frequency to every time step with:

```
./xmlchange HIST_OPTION=nsteps,HIST_N=1
```

Also, change where the cpl hist files are written so that they are written with the field values before the LND run. This way, the first cpl hist file written will contain the atm -> mediator fields from CAM initialization. Do this by copying `run/nuopc.runseq` to your case directory, then moving the line `MED med_phases_history_write` from near the bottom of the file to just before the `LND` line (after the `ICE` line).

Then rerun the case.

4. Look at the `atmImp_Sa_o3` field in the first cpl hist file to see values of this field in the first LND coupling, from CAM initialization.

A case set up this way is here: `/glade/scratch/sacks/b1850_o3test_1110a`. See `/glade/scratch/sacks/archive/b1850_o3test_1110a/cpl/hist/b1850_o3test_1110a.cpl.hi.0001-01-01-01800.nc` or the file `b1850_o3test_1110a.cpl.hi.0001-01-01-01800_o3.txt` which contains the dump of atmImp_Sa_o3 (showing the 3 NaN values).

cam6_3_082

CISL machine (e.g. cheyenne)

Intel

/glade/scratch/sacks/b1850_o3test_1110a

No

### Extra info

_No response_

### PR DESCRIPTION
**Pull request by adamrher**
Fixes #691 
Fixes #693

### CODE REVIEW

**Comment by adamrher** [comment]
I am running a test overnight to double check the climate for this PR. Will advise when that completes.

**Review by cacraigucar — APPROVED** [review]
Looks good!

File: src/physics/cam/vertical_diffusion.F90 [hunk 1]
```diff
@@ -1174,7 +1180,7 @@ subroutine vertical_diffusion_tend( &
      tmp1(:ncol) = ztodt * gravit * state%rpdel(:ncol,pver)
      do m = 1, pmam_ncnst
         l = pmam_cnst_idx(m)
-        q_tmp(:ncol,pver,l) = q_tmp(:ncol,pver,l) + tmp1(:ncol) * cam_in%cflx(:ncol,l)
+        q_tmp(:ncol,pver,l) = q_tmp(:ncol,pver,l) + tmp1(:ncol) * cflux(:ncol,l)
```
peverwhee, line 1183: does this change answers for non-cam_dev? since cflux(:ncol,1) = 0 and cam_in%cflx(:ncol, 1) could be something else?

peverwhee (commented):
one question

src/physics/cam/vertical_diffusion.F90
adamrher: it shouldn't change answers because the ```l = pmam_cnst_idx(m)``` should never equal 1.  These indices refer to aerosol species and the 1 is reserved as the water vapor index (i.e., not an aerosol).

cacraigucar: So far the 20'ish izumi regression tests which have run are all BFB. I'm currently running the izumi tests but they have a ways to go.

peverwhee: ok! thanks! resolving.

peverwhee (approved):

src/physics/cam/clubb_intr.F90
```diff
@@ -4256,7 +4256,64 @@ subroutine clubb_tend_cam( &
    return
 #endif
   end subroutine clubb_tend_cam
-    
+
+  subroutine clubb_emissions_cam (state, cam_in, ptend)
```
fvitt, line 4260: Does this routine belong in the clubb_intr module?  All it is doing is setting emissions tendencies regardless of the source of the emissions.  Maybe I am missing something.

fvitt (commented):
A question

src/physics/cam/clubb_intr.F90
adamrher: I had originally came up with a generic module to do this, but decided that this is a clubb based product because it is adding all fluxes of tracers to the lowest model level *except* vapor, because that is handled in clubb_tend_cam. Is that too weak an argument, ya think?

src/physics/cam/clubb_intr.F90
```diff
@@ -4256,7 +4256,64 @@ subroutine clubb_tend_cam( &
...
+  logical  :: lq(pcnst)
+  real(r8) :: q_tmp(pcols,pver,pcnst)
+
+  ! ----------------------- !
+  ! Main Computation Begins !
+  ! ----------------------- !
+
+  ncol = state%ncol
+
+  lq(:) = .true.
```
fvitt, line 4300: Should this be like this?
```suggestion
  lq(2:) = .true.
```

src/physics/cam/clubb_intr.F90
```diff
@@ -4256,7 +4256,64 @@ subroutine clubb_tend_cam( &
...
+  call physics_ptend_init(ptend,state%psetcols, "clubb emissions", lq=lq)
+
+  ! Apply tracer fluxes to lowest model level (except vapor)
+  do m = 2,pcnst
+    ptend%q(:ncol,pver,m) = cam_in%cflx(:ncol,m)*state%rpdel(:ncol,pver)*gravit
+  end do
+
+  ! Convert tendencies of dry constituents to dry basis.
+  do m = 1,pcnst
```
fvitt, line 4309: And this?
```suggestion
  do m = 2,pcnst
```

fvitt (commented):
We are assuming the first constituent index is always water vapor.  I know this assumption is in other places of the code.

src/physics/cam/clubb_intr.F90
fvitt: I don't know.  I encourage others to weigh in on this.

cacraigucar: I see the point that @fvitt is making, but I also see @adamrher point as well.  If we think that this routine could ever be used anywhere other than within CLUBB, then putting it in a generic module would make sense.  On the other hand, it does seem like overkill for a simplistic routine like this which is now only used by CLUBB.  I don't have a strong feeling one way or the other.

adamrher: That the emissions have historically been dumped into vertical_diffusion.F90, which has always contained the interface to the PBL mixing scheme, until CLUBB came along. I think that it is consistent to keep emissions tied to the PBL scheme, in this case CLUBB. If we ever move to a different PBL scheme then I think it is the responsibility of the developer to make sure the emissions are properly coupled to the PBL scheme. 

@swrneale @JulioTBacmeister does what I say seem convincing? Or am I missing something?

src/physics/cam/clubb_intr.F90
adamrher: Sure, but do we need to additionally specify that ```lq(1)=.false.``` ?

fvitt: Yes, we should explicitly set `lq(1)=.false.`

src/physics/cam/clubb_intr.F90
fvitt: @adamrher You make a good case

fvitt (approved):

src/physics/cam/clubb_intr.F90
adamrher: OK I'm going to insist that this be in clubb_intr. I don't think we need to bother Julio and Rich about this.

src/chemistry/mozart/chemistry.F90
```diff
@@ -812,6 +814,11 @@ subroutine chem_init(phys_state, pbuf2d)
 
      call ocean_emis_init()
 
+     ! initialize srf ozone to zero
+     if (is_first_step()) then
```
fvitt, line 818: Might need to be like this:
```suggestion
     if (is_first_step() .and. srf_ozone_pbf_ndx>0) then
```

nusbaume (approved):
Looks good!  Just a couple small things.

src/physics/cam/clubb_intr.F90
```diff
@@ -4256,7 +4256,62 @@ subroutine clubb_tend_cam( &
...
+  !              except water vapor (applied in clubb_tend_cam)
+  !
+  ! Author: Adam Herrington, November 2022
+  ! Origin: Based on E3SM's clubb_surface subroutine
+  ! References:
+  !   None
+  !-------------------------------------------------------------------------------
+  use physics_types,      only : physics_ptend, physics_ptend_init, physics_state
+  use constituents,       only : cnst_type
+  use camsrfexch,         only : cam_in_t
```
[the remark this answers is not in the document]
adamrher: done

src/chemistry/mozart/chemistry.F90
adamrher: I believe this is the more correct statement. committed.

src/physics/cam/clubb_intr.F90
Katetc: If anybody cares, I think it's fine this is in clubb_intr :)

src/physics/cam/vertical_diffusion.F90
```diff
@@ -1096,7 +1097,12 @@ subroutine vertical_diffusion_tend( &
      tauy = 0._r8
      shflux = 0._r8
      cflux(:,1) = 0._r8
-     cflux(:,2:) = cam_in%cflx(:,2:)
+     if (cam_physpkg_is("cam_dev")) then
```
Katetc, line 1100: Is this going to be a long-term switch? I imagined that eventually cam_dev would become cam7 and would be the default cam. I suppose it won't be a lot of work to change the physpkg name, but are there a lot of "cam_dev" checks floating around?

src/physics/cam_dev/physpkg.F90
```diff
@@ -1571,6 +1571,16 @@ subroutine tphysac (ztodt,   cam_in,  &
 
     call t_stopf('tphysac_init')
 
+    !===================================================
+    ! Apply tracer surface fluxes to lowest model layer
+    !===================================================
+    call t_startf('clubb_emissions_tend')
+
+    call clubb_emissions_cam(state, cam_in, ptend)
+    call physics_update(state, ptend, ztodt, tend)
```
Katetc, line 1580: I feel like it would be nice to call check_energy_chng after this to make sure we don't have any unexpected leaks/sinks in the future.

adamrher: yes, good catch. it shouldn't change cam's energy, but that's why we check!

src/physics/cam/vertical_diffusion.F90
adamrher: this will be the second place in the code im aware of -- andrew put one in nucleate_ice_cam so as to not add preexisting ice for cam_dev. Is that what you're asking about when you ask if there are a lot of "checks"?

I recognize these hooks can get out of hand quickly. But I think that the real problem is with how've loaded up vertical_diffusion with a bunch of different physical processes (surface fluxes, pbl diffusion, molecular diffusion, beljaars and tms) that we pick and choose which to use depending on the physics and the configuration. This is also true for nucleate_ice, and it's cross-over with pumas. And this is exactly the type of thing that will need to be refactored as we try to ccpp'ize it ... and so I really view the current cam_dev hooks as a few ugly blips that I will plan to gut out once we transition to ccpp. Sorry that was a lot.

Katetc: Yeah, that's what I was asking about. I'd say less than 10 total is fine for the time being, but if we start piling up a ton of "if(cam_dev)" calls then maybe we should all have a discussion so there's not too much technical debt going into cam7.

src/physics/cam_dev/physpkg.F90
adamrher: im a bit surprised there is no check_energy_chng for the vertical_diffusion_tend call in tphysac. This must be an oversight?

adamrher: ok, I added the check_energy call to the emissions call, and verified it runs and doesn't print energy errors when you turn on it on.

adamrher: i may open an issue about other lapses i think im seeing in calling the energy and tracer checkers in tphysac

PR #703, opened 2022-11-21
Fixes initialization for cam_dev

### LINKED ISSUE #701: interpolation error on initialization in cam_dev
### What happened?

When running cam_dev with L93, the model crashes on startup with an interpolation error originating from rrtmg/cloud_rad_props.F90.  Details are [here](https://github.com/NCAR/amwg_dev/issues/201#issuecomment-1320246549).

From @cecilehannay:

## 
cd /glade/work/hannay/cesm_tags/cesm3_cam6_3_058_MOM_e/cime/scripts
./create_newcase --compset   FWscHIST  --res ne30pg3_ne30pg3_mg17 --case /glade/p/cesmdata/cseg/runs/cesm2_0/f.cesm3_cam058_mom_e.FWscHIST.ne30_L93.26c_non-orogw_on.001 --run-unsupported  --pecount 2160 --project 93300722 

## set up
cd /glade/p/cesmdata/cseg/runs/cesm2_0/f.cesm3_cam058_mom_e.FWscHIST.ne30_L93.26c_non-orogw_on.001  
./xmlchange CAM_CONFIG_OPTS="-phys cam_dev -microphys mg2 -chem waccm_sc_mam4 -nlev 93"
./case.setup

./xmlchange RUN_STARTDATE=1979-01-01
./xmlchange STOP_N=2
./xmlchange STOP_OPTION=nyears
./xmlchange RESUBMIT=5
./xmlchange RUN_TYPE=hybrid
./xmlchange RUN_REFCASE=f.e21.FWscHIST_BGC.ne30_ne30_mg17_L48_revert-J.001
./xmlchange RUN_REFDATE=1989-01-01
./xmlchange GET_REFCASE=TRUE
./xmlchange RUN_REFDIR=cesm2_init

Use the same user_nl_* and SourceMods as in: 
/glade/p/cesmdata/cseg/runs/cesm2_0/f.cesm3_cam058_mom_e.FWscHIST.ne30_L93.26c_non-orogw_on.001

cam6_3_058

CISL machine (e.g. cheyenne)

Intel

Cecile's above, or mine: /glade/scratch/stepheba/l93058test

### Extra info

The problem was that the variable `dei` was not initialized appropriately when running with cam_dev.  The fix is just to use a call to `pbuf_set_field` in order to initialize this field to zero.  This involves a one-line change to rrtmg/radiation.F90 and a few changes to rrtmg/cloud_rad_props.F90.

cacraigucar:
Closes #701


src/physics/rrtmg/cloud_rad_props.F90
```diff
@@ -113,6 +115,12 @@ subroutine cloud_rad_props_init()
    i_icgrauwp  = pbuf_get_index('ICGRAUWP',errcode=err) ! Available when using MG3
    i_degrau    = pbuf_get_index('DEGRAU',errcode=err)   ! Available when using MG3
```
adamrher, line 116: I presume this returns ```i_degrau <= 0``` if it's not avail? Can you confirm we don't need to initialize these ```i_``` variables before this call? Other places in the code they are initialized to =0 or =-1 at the module level, and so wondering if that may be necessary here.

adamrher (approved):
Looks good Ben. Just has one clarification about these i variables we use in the pbuf_get_index calls, and whether they need to be initialized.

src/physics/rrtmg/cloud_rad_props.F90
cacraigucar: The indices do not need to be initialized locally.  In pbuf_get_index, the index is initialized to -1.  If the field does not exist and errcode is passed in, then the index and errcode are both set to -1.  If the field does not exist and errcode is not passed in, the an endrun is called.

bstephens82: Ok thanks Cheryl, was going to check this with a print statement but sounds like it's okay.

gold2718: > I presume this returns `i_degrau <= 0` if it's not avail? Can you confirm we don't need to initialize these `i_` variables before this call? Other places in the code they are initialized to =0 or =-1 at the module level, and so wondering if that may be necessary here.

It is true that `pbuf_get_index` always returns a value (as an integer function must) that is the positive index or `-1`.
The philosophical reason to initialize the module variables is protection against any current or future logic that causes `pbuf_get_index` to not get called (a misplaced `ifdef(masterproc)` which has happened in the past). So initialization is overwritten if `pbuf_get_index` is called, but only if it is called.

cacraigucar: @gold2718  - That is a good point.  In Ben's case, the call to retrieve the index is right above.  That said, initializing all of the indices to a set value (I'll choose -1) make sense.

bstephens82 (approved):
This looks good to me, I can't think of anything else to improve it now that we are initializing the indices.

peverwhee (approved):
looks good to me!

jtruesdal (approved):
This all looks good to me too.

jtruesdal (approved):

peverwhee (approved):

Issue #582, opened 2022-04-26
moving CLUBB pdf closure to after advancing the CLUBB equations #582


adamrher:
In https://github.com/ESCOMP/CAM/pull/545, UWM wanted to move the pdf closure to the post position (i.e., call pdf closure after advancing the prognostic moments). I believe the motivation is because the pdf closure is CAM's saturation adjustment, which is usually the last process to be called because it provides an instantaneous adjustment. We decided against moving the pdf closure in https://github.com/ESCOMP/CAM/pull/545, and I instead have opened this issue to discuss what I perceive as the obstacles needed to be addressed before moving the pdf closure.

The pdf closure provides variables that are used to advance the prognostic equations, and we currently call the pdf closure at the top of CLUBB, so that all routines in CLUBB are able to use to those variables. By moving the pdf closure to the post position, the necessary variables will need to be saved from time-step to time-step, and fed in to CLUBB to be used to advance the CLUBB equations. UWM proposed saving these variables time-step to time-step by expanding their scope in the PBUF to 'global':

```
    call pbuf_add_field('WP2THVP',    'global', dtype_r8, (/pcols,pverp/), wp2thvp_idx)
    call pbuf_add_field('RTPTHVP',    'global', dtype_r8, (/pcols,pverp/), rtpthvp_idx)
    call pbuf_add_field('THLPTHVP',   'global', dtype_r8, (/pcols,pverp/), thlpthvp_idx)
    call pbuf_add_field('CLOUD_FRAC', 'global', dtype_r8, (/pcols,pverp/), cloud_frac_idx)
    call pbuf_add_field('ISS_FRAC',   'global',  dtype_r8, (/pcols,pverp/), ice_supersat_idx)
    call pbuf_add_field('RCM',        'global', dtype_r8, (/pcols,pverp/), rcm_idx)
```

The first three variables are probably fine to keep in the PBUF, but they are incorrect if we turn on advecting the moments via ```clubb_do_adv```. Looking through ```advance_clubb_core``` I don't believe CLOUD_FRAC is used anywhere prior to the post position of the pdf closure (I only see one place where CLOUD_FRAC might be used before the pdf closure, and that's for adjusting the C2 coefficient using cloud_frac (l_C2_cloud_frac), which as far as I can tell is defaulting to false).

[...]

PR #712, opened 2022-12-05
cam6_3_091: Katetc/new clubb cam6 3 080

### LINKED ISSUE #675: Bring in new CLUBB external!

The support we need to address the removal of the clubbtop limiter and some of the issues surrounding the monotonic flux limiter in CLUBB as well. Since only a few (mostly summer) months have passed since the last external update, changes should be minimal. 

Note that there will be changes to clubb_intr.F90 and namelist support files.

@adamrher @bstephens82 @JulioTBacmeister @vlarson

Yes

Yes

Katetc:
Fixes #675

Merging in changes from cam_development and uwm development. Namelist rearrange to alphabetical order for ease of checking going forward.


bstephens82 (approved):

src/physics/cam/clubb_intr.F90
```diff
@@ -3629,13 +3676,14 @@ subroutine clubb_tend_cam( &
     ! variable.
     if (clubb_do_energyfix) then
       do i=1, ncol
-        do k=clubbtop(i)+1,pver
+        do k=clubbtop(i),pver
```
adamrher, line 3679: comment to be removed after ensuring this branch doesn't crash:

I changed the depth of the active layer to be between pverp and clubbtop, instead of pverp and clubbtop+1. I think this is a better option since we are defining the active layer using the interface grid, not the mid-point grid from which it was defined. Therefore the interface above k_midpoint=clubbtop is k_interface=clubbtop.

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4116,206 @@ subroutine clubb_tend_cam( &
...
+    call outfld( 'WP2_CLUBB',        wp2,                   pcols, lchnk )
+    call outfld( 'UP2_CLUBB',        up2,                   pcols, lchnk )
+    call outfld( 'VP2_CLUBB',        vp2,                   pcols, lchnk )
+    call outfld( 'WP3_CLUBB',        wp3_output,            pcols, lchnk )
+    call outfld( 'UPWP_CLUBB',       upwp,                  pcols, lchnk )
+    call outfld( 'VPWP_CLUBB',       vpwp,                  pcols, lchnk )
+    call outfld( 'WPTHLP_CLUBB',     wpthlp_output,         pcols, lchnk )
+    call outfld( 'WPRTP_CLUBB',      wprtp_output,          pcols, lchnk )
+
+    temp2dp(:ncol,:) =  rtp2(:ncol,:)*1.0e6_r8
```
adamrher, line 4214: veto unit conversions as these are old mods and no longer consistent with units in history fields

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4116,206 @@ subroutine clubb_tend_cam( &
...
+    call outfld( 'RTP2_CLUBB',       temp2dp,                 pcols, lchnk )
+
+    rtpthlp_output(:ncol,:) = rtpthlp_output(:ncol,:) * 1000._r8
```
adamrher, line 4217: veto unit conversions as these are old mods and no longer consistent with units in history fields

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4116,206 @@ subroutine clubb_tend_cam( &
...
+    call outfld( 'RTPTHLP_CLUBB',    rtpthlp_output,          pcols, lchnk )
+
+    temp2dp(:ncol,:) = rcm(:ncol,:) * 1000._r8
```
adamrher, line 4220: veto unit conversions as these are old mods and no longer consistent with units in history fields

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4116,206 @@ subroutine clubb_tend_cam( &
...
+    call outfld( 'RCM_CLUBB',        temp2dp,                 pcols, lchnk )
+
+    temp2dp(:ncol,:) = wprcp(:ncol,:) * latvap
+    call outfld( 'WPRCP_CLUBB',      temp2dp,                 pcols, lchnk )
+
+    temp2dp(:ncol,:) = rcm_in_layer(:ncol,:) * 1000._r8
```
adamrher, line 4226: veto unit conversions as these are old mods and no longer consistent with units in history fields

adamrher (changes requested):
Request changes please. I noticed that some of the mods in the old clubb_intr, before my kitchen sink PR cleaning it up, have made there way back in (best example is the outfld call for rcm, trying to multiply it by 1000 to make it g/kg when it should be in kg/kg). It seems inevitable that I will miss something that I previously cleaned up and will get snuck back in ...

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1858,75 +1864,80 @@
...
+<clubb_l_lmm_stepping                             > .false. </clubb_l_lmm_stepping>
 <clubb_l_lscale_plume_centered                    > .false. </clubb_l_lscale_plume_centered>
 <clubb_l_min_wp2_from_corr_wx                     > .true.  </clubb_l_min_wp2_from_corr_wx>
 <clubb_l_min_xp2_from_corr_wx                     > .true.  </clubb_l_min_xp2_from_corr_wx>
+<clubb_l_mono_flux_lim_rtm                        > .false. </clubb_l_mono_flux_lim_rtm>
+<clubb_l_mono_flux_lim_spikefix                   > .false. </clubb_l_mono_flux_lim_spikefix>
+<clubb_l_mono_flux_lim_thlm                       > .false. </clubb_l_mono_flux_lim_thlm>
+<clubb_l_mono_flux_lim_um                         > .false. </clubb_l_mono_flux_lim_um>
+<clubb_l_mono_flux_lim_vm                         > .false. </clubb_l_mono_flux_lim_vm>
 <clubb_l_predict_upwp_vpwp                        > .false. </clubb_l_predict_upwp_vpwp>
```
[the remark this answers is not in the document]
bstephens82: Yes those are the right settings for those two flags.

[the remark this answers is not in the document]
adamrher: just to be clear ```clubb_l_mono_flux_lim_spikefix``` increases the lower limit on variance for thlm? whereas ```clubb_l_mono_flux_lim_rtm=.true.``` limits the fluxes? It is these two pieces that we agree to use as the recipe for (mostly) eliminating spikes, right? Or am I forgetting something?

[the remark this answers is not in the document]
bstephens82: Right the thlm mfl flag should be true also. The um and vm flags should be false, at least until we turn on prognostic momentum fluxes (I believe the mfl would not operate on um and vm unless we were using prognostic momentum fluxes).

src/physics/cam/clubb_intr.F90
```diff
@@ -1840,771 +1927,787 @@ end subroutine clubb_ini_cam
...
+      grid_dx, & 
+      grid_dy  ! CAM grid [m]      
+      
+    real(r8), dimension(state%ncol,sclr_dim) :: &
+      wpsclrp_sfc            ! Scalar flux at surface                        [{units vary} m/s]
+
+    real(r8), dimension(state%ncol,edsclr_dim) :: &
+      wpedsclrp_sfc        ! Eddy-scalar flux at surface                   [{units vary} m/s]
+
+    real(r8), dimension(state%ncol,pverp+1-top_lev) :: &
```
cacraigucar, line 2086: I am not keen on having a large number of arrays dimensioned to be "ncol" instead of "pcols".  The reason being that it makes it much harder to keep track of array sizes since there will be a mixture of both as all variables which are stored in the pbuf need to be dimensioned "pcols".  Typically in CAM we always set arrays to be pcols and if we need to operate on only ncol of them, we do that via `(:ncol)`  @Katetc - Could you elaborate on why you made this modification? 

Note that the PUMAS code has a mixture of array dimensions and I found it cumbersome to need to check which array was which size when I was making mods.  I'm not sure I like to see this bleeding into the CLUBB code as well.

@nusbaume - do you have an opinion?

I realize at this late date, we probably don't have a choice but to accept these changes as reverting it back to pcols would be very time consuming.

cacraigucar (commented):
First round of review comments

bld/namelist_files/namelist_defaults_cam.xml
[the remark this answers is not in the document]
bstephens82: Sorry when I was responding to this last night I was on my phone and I think I couldn't see the code you were referring to.  To give a more complete answer, the correct flag settings here are:
~~~
<clubb_l_mono_flux_lim_rtm                        > .true. </clubb_l_mono_flux_lim_rtm>
<clubb_l_mono_flux_lim_spikefix                   > .true. </clubb_l_mono_flux_lim_spikefix>
<clubb_l_mono_flux_lim_thlm                       > .true. </clubb_l_mono_flux_lim_thlm>
<clubb_l_mono_flux_lim_um                         > .false. </clubb_l_mono_flux_lim_um>
<clubb_l_mono_flux_lim_vm                         > .false. </clubb_l_mono_flux_lim_vm>
~~~
When we turn on prognostic momentum fluxes at some point (ie l_predict_upwp_vpwp=T) then we can presumably set the um and vm flags here to true. They are true in the UWM CLUBB repo (which uses prognostic momentum fluxes by default).

bld/namelist_files/namelist_definition.xml
```diff
@@ -3807,6 +3814,12 @@ Gaussian PDF, and also decreases the difference between the means of w from
 each Gaussian.
 </entry>
 
+<entry id="clubb_ipdf_call_placement" type="integer" category="pblrad"
+       group="clubb_params_nl" valid_values="1,2,3" >
+Option for the placement of the call to CLUBB's PDF closure. The options include: ipdf_pre_advance_fields (1) calls the PDF closure before advancing prognostic fields. ipdf_post_advance_fields (2) calls after advancing prognostic fields, and ipdf_pre_post_advance_fields (3) calls both before and after advancing prognostic fields.
+Default: 1
```
bstephens82, line 3820: The default here is "2" now I believe.  I agree with Cheryl's comment about having the defaults here being possibly problematic.

bld/namelist_files/namelist_definition.xml
```diff
@@ -3820,75 +3833,83 @@ scale (Lscale) in a grid-spacing dependent formula.  Increasing the value of
...
+Default: .false.
 </entry>
 
-<entry id="clubb_detice_rad" type="real" category="pblrad"
+<entry id="clubb_l_e3sm_config" type="logical" category="pblrad"
        group="clubb_params_nl" valid_values="" >
-Radius of detrained ice drops as they are used in the CLUBB parameterization in units of (m).
-Default: 25.0e-6 m
+Flag to run CLUBB with E3SM settings.
+Default: .true.
```
bstephens82, line 3883: The default for this flag is false.

bld/namelist_files/namelist_definition.xml
```diff
@@ -3922,6 +3943,36 @@ realizable until the correlation becomes realizable.  This correction is applied
 at the point in the code where xp2 is advanced one timestep.
 </entry>
 
+<entry id="clubb_l_mono_flux_lim_rtm" type="logical" category="pblrad"
+       group="clubb_params_nl" valid_values="" >
+Flag to turn on the clubb monotonic flux limiter for rtm.
+Default: .false.
```
bstephens82, line 3949: Default value here should be updated once it's fixed in namelist_defaults.

bld/namelist_files/namelist_definition.xml
```diff
@@ -3922,6 +3943,36 @@ realizable until the correlation becomes realizable.  This correction is applied
...
+</entry>
+
+<entry id="clubb_l_mono_flux_lim_spikefix" type="logical" category="pblrad"
+       group="clubb_params_nl" valid_values="" >
+Flag to turn on the smoothing code options for the monotonic flux limiter.
+Default: .false.
```
bstephens82, line 3955: Default value here should be updated once it's fixed in namelist_defaults.

bld/namelist_files/namelist_definition.xml
```diff
@@ -3922,6 +3943,36 @@ realizable until the correlation becomes realizable.  This correction is applied
...
+</entry>
+
+<entry id="clubb_l_mono_flux_lim_thlm" type="logical" category="pblrad"
+       group="clubb_params_nl" valid_values="" >
+Flag to turn on the clubb monotonic flux limiter for thlm.
+Default: .false.
```
bstephens82, line 3961: Default value here should be updated once it's fixed in namelist_defaults.

bstephens82 (changes requested):

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1858,75 +1864,80 @@
...
-<clubb_l_e3sm_config                              > .false.  </clubb_l_e3sm_config>
 <clubb_l_use_tke_in_wp2_wp3_K_dfsn                > .false. </clubb_l_use_tke_in_wp2_wp3_K_dfsn>
 <clubb_l_use_tke_in_wp3_pr_turb_term              > .false. </clubb_l_use_tke_in_wp3_pr_turb_term>
-<clubb_l_smooth_Heaviside_tau_wpxp                > .false. </clubb_l_smooth_Heaviside_tau_wpxp>
-<clubb_l_do_expldiff_rtm_thlm                     > .false. </clubb_l_do_expldiff_rtm_thlm>
+<clubb_l_vert_avg_closure                         > .true.  </clubb_l_vert_avg_closure>
+<clubb_mult_coef                                  > 1.0D0   </clubb_mult_coef>
+<clubb_nu2                                        > 5.0     </clubb_nu2>
+<clubb_nu9                                        > 20.0    </clubb_nu9>
+<clubb_penta_solve_method                         > 1       </clubb_penta_solve_method>
```
bstephens82, line 1935: In the namelist_definitions file this is said to default to 0 (which was my expectation since this is new code that to my knowledge is somewhat experimental).  So should this be set to zero here?

bld/namelist_files/namelist_definition.xml
```diff
@@ -4002,94 +4053,95 @@ grid box.
...
 
-<entry id="clubb_l_damp_wp2_using_em" type="logical" category="pblrad"
+<entry id="clubb_tridiag_solve_method" type="integer" category="pblrad"
        group="clubb_params_nl" valid_values="" >
-Flag to use a dissipation formula of -(2/3)*em/tau_zm, as in Bougeault (1981),
-in the wp2 (variance of vertical velocity) predictive equation.
-Default: .false.
+Specifier for method to solve tri-diagonal systems that are common in CLUBB.
+Valid values: 0 (lapack), 1 (penta_lu), 2 (penta_bicgstab)
+Default: 0
```
bstephens82, line 4130: These two lines I think need to be updated.  The parameter `clubb_tridiag_solve_method` only has one valid value currently, which is "1" (so that should be the default here---it's set correctly in namelist_defaults).  But the "Valid values" line should also be updated to reflect this.  Option "1" is lapack...I don't believe there's any other viable option.  CLUBB's code here will shut it down if this equals something other than the lapack option: https://github.com/larson-group/clubb/blob/873ae270b4c4c112e6e7da83ba0f7df4aff9b958/src/CLUBB_core/matrix_solver_wrapper.F90#L342-L359.

Katetc:
> Request changes please. I noticed that some of the mods in the old clubb_intr, before my kitchen sink PR cleaning it up, have made there way back in (best example is the outfld call for rcm, trying to multiply it by 1000 to make it g/kg when it should be in kg/kg). It seems inevitable that I will miss something that I previously cleaned up and will get snuck back in ...

Yes, Adam, and thank you for your help with this. After the last external update, our changes to clubb_intr.F90 didn't get pushed back to the UWM repo and they also did a lot of refactoring of this file this summer. This PR represents my best attempt at merging everybody's changes back together, but it is inevitable that some things don't merge exactly correctly. After this CAM PR we will make sure to push all these changes back into the UWM CAM repo.

src/physics/cam/clubb_intr.F90
```diff
@@ -295,12 +292,21 @@ module clubb_intr
...
-    clubb_l_update_pressure               ! Flag for having CLUBB update pressure and exner
-  
+    clubb_l_update_pressure,            & ! Flag for having CLUBB update pressure and exner
+    clubb_l_mono_flux_lim_thlm,         & ! Flag to turn on monotonic flux limiter for thlm
+    clubb_l_mono_flux_lim_rtm,          & ! Flag to turn on monotonic flux limiter for rtm
+    clubb_l_mono_flux_lim_um,           & ! Flag to turn on monotonic flux limiter for um
+    clubb_l_mono_flux_lim_vm,           & ! Flag to turn on monotonic flux limiter for vm
+    clubb_l_mono_flux_lim_spikefix,     & ! Flag to implement monotonic flux limiter code that
+                                          ! eliminates spurious drying tendencies at model top  
+    clubb_l_intr_sfc_flux_smooth = .false.  ! Add a locally calculated roughness to upwp and vpwp sfc fluxes
```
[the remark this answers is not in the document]
Katetc: The clubb_l_intr_sfc_flux_smooth variable is the namelist flag for Colin's "patch" that I just added in this PR. Since it only applies to code in clubb_intr, I didn't add the variable to the clubb_config_flags_api that sets the default values for the rest of the variables in a call to set_default_clubb_config_flags_api. So, I guess the answer to your fist question (convention) is yes, they are set in the config_flags_api. And yes, this is an exception because it is a flag not actually used by clubb.

bld/build-namelist
```diff
@@ -3151,76 +3153,80 @@ if ($clubb_sgs  =~ /$TRUE/io) {
...
-
+   add_default($nl, 'clubb_lambda0_stability_coef');
+   add_default($nl, 'clubb_lmin_coef');
    add_default($nl, 'clubb_l_brunt_vaisala_freq_moist');
    add_default($nl, 'clubb_l_call_pdf_closure_twice');
+   add_default($nl, 'clubb_l_damp_wp2_using_em');
    add_default($nl, 'clubb_l_damp_wp3_Skw_squared');
-   add_default($nl, 'clubb_l_lmm_stepping');
+   add_default($nl, 'clubb_l_do_expldiff_rtm_thlm');
    add_default($nl, 'clubb_l_e3sm_config');
```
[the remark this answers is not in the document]
Katetc: Ok, so, these are parameters introduced in this version of clubb_intr.F90 but the UWM guys didn't put them in the read namelist code. They actually don't seem to be in the namelist at the moment. There's no requirement that every logical parameter in clubb go into the CAM namelist, and I don't see that these were expected to be put into the namelist since they aren't in clubb_readnl. There are many other clubb parameters that aren't in clubb_readnl and aren't included in the CAM namelist. Similar to clubb_l_use_precip_frac, clubb_l_predict_upwp_vpwp and others. I assume these would be set using a source mod if somebody wanted to set them. And that UWM doesn't want them exposed to the namelist because they may not be a commonly used setting (or something they don't want us setting) in CAM.

bld/namelist_files/namelist_defaults_cam.xml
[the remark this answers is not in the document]
Katetc: Updated to these.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1858,75 +1864,80 @@
...
+<clubb_ipdf_call_placement                        > 2       </clubb_ipdf_call_placement>
+<clubb_lambda0_stability_coef                     > 0.04    </clubb_lambda0_stability_coef>
+<clubb_lmin_coef                                  > 0.1     </clubb_lmin_coef>
 <clubb_l_brunt_vaisala_freq_moist                 > .false. </clubb_l_brunt_vaisala_freq_moist>
 <clubb_l_call_pdf_closure_twice                   > .true.  </clubb_l_call_pdf_closure_twice>
+<clubb_l_damp_wp2_using_em                        > .false. </clubb_l_damp_wp2_using_em>
 <clubb_l_damp_wp3_Skw_squared                     > .false. </clubb_l_damp_wp3_Skw_squared>
+<clubb_l_do_expldiff_rtm_thlm                     > .false. </clubb_l_do_expldiff_rtm_thlm>
+<clubb_l_e3sm_config                              > .false. </clubb_l_e3sm_config>
+<clubb_l_godunov_upwind_wpxp_ta                   > .false. </clubb_l_godunov_upwind_wpxp_ta>
```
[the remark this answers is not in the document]
Katetc: See comment above. Since UWM didn't put these in clubb_readnl, I think they don't want them exposed as namelist options.

src/physics/cam/clubb_intr.F90
```diff
@@ -538,7 +544,7 @@ subroutine clubb_register_cam( )
     call pbuf_add_field('THLPTHVP',   'global', dtype_r8, (/pcols,pverp/), thlpthvp_idx)
     call pbuf_add_field('CLOUD_FRAC', 'global', dtype_r8, (/pcols,pverp/), cloud_frac_idx)
     call pbuf_add_field('ISS_FRAC',   'global',  dtype_r8, (/pcols,pverp/), ice_supersat_idx)
-    call pbuf_add_field('RCM',        'physpkg', dtype_r8, (/pcols,pverp/), rcm_idx)
+    call pbuf_add_field('RCM',        'global', dtype_r8, (/pcols,pverp/), rcm_idx)
```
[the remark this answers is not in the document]
Katetc: Ok so I was confused because the larson-group repo has this change with a fairly recent (February 2022) commit. But I have found all of the emails and this conversation: https://github.com/ESCOMP/CAM/issues/582 that explains the reasoning here. Reverted.

src/physics/cam/clubb_intr.F90
```diff
@@ -3617,7 +3664,7 @@ subroutine clubb_tend_cam( &
       te_b(i) = te_b(i) + (cam_in%shf(i)+cam_in%cflx(i,1)*(latvap+latice)) * hdtime      
 
       ! Compute the disbalance of total energy, over depth where CLUBB is active
-      se_dis(i) = (te_a(i) - te_b(i))/(state1%pint(i,pverp)-state1%pint(i,clubbtop(i)+1))
+      se_dis(i) = (te_a(i) - te_b(i))/(state1%pint(i,pverp)-state1%pint(i,clubbtop(i)))
```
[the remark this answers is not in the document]
Katetc: This did trigger a test failure in one cheyenne test where clubbtop(i) == pverp and we had a divide by zero. I updated to check for that before doing the division. But, looking at your comment above, I just wanted to let you know this was possible.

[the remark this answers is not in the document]
adamrher: ok, that was def. my fault. I need to go in and modify three lines of code to revert back then. I am going to try to do a PR against your branch correctly. stay tuned.

src/physics/cam/clubb_intr.F90
```diff
@@ -2891,20 +3001,41 @@ subroutine clubb_tend_cam( &
...
       wpthlp_sfc(i) = cam_in%shf(i)/(cpairv(i,pver,lchnk)*rho_zt(i,2))! Sensible heat flux
       wpthlp_sfc(i) = wpthlp_sfc(i)*inv_exner_clubb_surf(i)   ! Potential temperature flux
       wprtp_sfc(i)  = cam_in%cflx(i,1)/rho_zt(i,2)            ! Moisture flux  (check rho)
-      upwp_sfc(i)   = cam_in%wsx(i)/rho_zt(i,2)               ! Surface meridional momentum flux
-      vpwp_sfc(i)   = cam_in%wsy(i)/rho_zt(i,2)               ! Surface zonal momentum flux  
     end do
+
+    !  Other Surface fluxes provided by host model
+    if( (cld_macmic_num_steps > 1) .and. (clubb_l_intr_sfc_flux_smooth) ) then
+       do i=1,ncol
```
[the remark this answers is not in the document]
Katetc: Don't know why the PR isn't showing "outdated" but I did this.

src/physics/cam/clubb_intr.F90
Katetc: Ok so I just reimplemented  all of these changes and I had the distinct feeling that I've done this before. I think maybe I did integrate these changes back in but they ... got lost ... somewhere. I suspect I will find them on a branch in a different clone somewhere else.... Or an unsaved emacs file backup....

src/physics/cam/clubb_intr.F90
Katetc: Veto'd

src/physics/cam/clubb_intr.F90
Katetc: Veto'd

src/physics/cam/clubb_intr.F90
Katetc: Veto'd

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1829,25 +1829,31 @@
 <clubb_history               > .false. </clubb_history>
 <clubb_rad_history           > .false. </clubb_rad_history>
 
-<!-- CLUBB options -->
-<clubb_do_icesuper                                > .false. </clubb_do_icesuper>
+<!-- CLUBB pbl diff options -->
 <clubb_cloudtop_cooling                           > .false. </clubb_cloudtop_cooling>
 <clubb_rainevap_turb                              > .false. </clubb_rainevap_turb>
+<clubb_do_adv                                     > .false. </clubb_do_adv>
+<clubb_do_adv                           silhs="1" > .false. </clubb_do_adv>
```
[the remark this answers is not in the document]
Katetc: Previously there was ONLY a silhs line and I wasn't sure what namelist_defaults was doing in the non-silhs case. I added the other line to be explicit about what happens in default cases.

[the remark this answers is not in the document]
cacraigucar: Since you now have the default for all of CLUBB, I suggest you remove the line for SILHS, since it will be set for SILHS via line 1835.

bld/build-namelist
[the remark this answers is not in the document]
adamrher: OK that's fine by me if UWM doesn't want all namelists available at the CAM level. I propose we assume that if UWM puts it in the readnl subroutine, then it's intended for CAM namelists.  ```clubb_l_enable_relaxed_clipping``` is in readnl, whereas ```clubb_l_linearize_pbl_winds``` is not, I think.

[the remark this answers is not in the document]
Katetc: Sorry Adam, I don't see clubb_l_enable_relaxed_clipping in the `namelist /clubb_params_nl/ ` list. It is in the MPI_BCAST calls, but not in the namelist list. This is why I don't think it should be add to the cam namelist.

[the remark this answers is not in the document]
adamrher: ok, fine with me. ill resolve this convo once you remove the mpi bast call in readnl.

bld/namelist_files/namelist_defaults_cam.xml
Katetc: When I added both `penta_solve_method` and `tridiag_solve_method` to the clubb namelist files, I started with "0" for the same reasoning, but clubb crashed on the first timestep that way. I removed the namelist items and ran the test (and it worked) and the clubb log wrote out default values of 1 for both of these. When I put default values of 1 here, they work great. I'm running a test now to try `penta_solve=0` and `tridiag=1`. If that works I'm fine with that setting. But, the code suggests the defaults should be 1 for both.

bld/namelist_files/namelist_definition.xml
```diff
@@ -3671,53 +3657,59 @@ the damping of CLUBB's wp3 when skewness of w (vertical velocity) is large in
 magnitude.
 </entry>
 
-<entry id="clubb_C_wp3_pr_turb" type="real" category="pblrad"
+<entry id="clubb_C_invrs_tau_bkgnd" type="real" category="pblrad"
        group="clubb_params_nl" valid_values="" >
-Coefficient in the pressure-turbulence term of CLUBB's wp3 predictive equation.
+Coefficient of inverse tau term contributed by background constant value (units: none)
+Default: 1.0
```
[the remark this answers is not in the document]
Katetc: Hmm. Ok, so Lscale_from_tau is false by default, so, technically, the defaults here are the default. I don't think we usually document special case defaults here. For instance, many of these CLUBB fields have different defaults when SILHS is turned on. Should I document that here too?

bld/namelist_files/namelist_defaults_cam.xml
bstephens82: Ah, ok, I see what's going on...there is no "0" for penta_solve_method in the source code, it starts at "1" which is lapack, and then 2 and 3 are the newer experimental options.  So yeah I guess both should default to "1".

bld/namelist_files/namelist_definition.xml
```diff
@@ -3732,64 +3724,79 @@ Thermo of Kh_zm
 Default: 0.3
```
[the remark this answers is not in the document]
Katetc: I am totally in support of removing the Default: documentation here. So I did it.

bld/namelist_files/namelist_defaults_cam.xml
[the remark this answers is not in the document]
Katetc: Okee dokee

bld/namelist_files/namelist_defaults_cam.xml
Katetc: Yep. Test failed. Defaults are still 1.

bld/namelist_files/namelist_definition.xml
Katetc: Ah good catch. Updated.

bstephens82: yeah I think this was why I thought lapack was "0" haha.

bld/namelist_files/namelist_definition.xml
Katetc: Removing the default values from this file because it's annoying for everybody to keep them up to date.

bld/namelist_files/namelist_definition.xml
Katetc: Removing the default values from this file because it's annoying for everybody to keep them up to date.

bld/namelist_files/namelist_definition.xml
Katetc: Removing the default values from this file because it's annoying for everybody to keep them up to date.

bld/namelist_files/namelist_definition.xml
Katetc: Removing the default values from this file because it's annoying for everybody to keep them up to date.

bld/namelist_files/namelist_definition.xml
Katetc: Removing the default values from this file because it's annoying for everybody to keep them up to date.

src/physics/cam/clubb_intr.F90
Katetc: Added some warning labels. Comments that point out appropriate usage for NCOL or PCOL arrays.

bld/build-namelist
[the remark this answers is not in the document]
zarzycki: > Similar to clubb_l_use_precip_frac, **clubb_l_predict_upwp_vpwp and others**. I assume these would be set using a source mod if somebody wanted to set them.

I assume this was a typo (or I'm misinterpreting), but we definitely want `clubb_l_predict_upwp_vpwp` to be exposed via the namelist (it currently is). :)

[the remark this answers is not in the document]
zarzycki: FWIW, based on https://github.com/NCAR/amwg_dev/discussions/106#discussioncomment-3614557, we do apply a source mod to `advance_xm_wpxp_module` to change `l_enable_relaxed_clipping` as part of the CPT/CAM7 code. This now appears to be set as a default to .false. in `model_flags.F90` in the UWM repo.

```
l_enable_relaxed_clipping = .true.      !advance_xm_wpxp_module
```

So ideally we'd either set this to .true. across the board in the CLUBB external for CAM or we include it as a namelist option with a default of .false. to mimic CAM6 behavior but .true. for CAM_dev/CAM7. I think. I cannot remember why we toggled it to .true., I believe it had something to do with our runaway cooling over rough terrain in some 1deg simulations.

src/physics/cam/clubb_intr.F90
```diff
@@ -3010,10 +3015,27 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
+          upwp_sfc(i) = -state1%u(i,pver)*ustar**2/ubar
+          vpwp_sfc(i) = -state1%v(i,pver)*ustar**2/ubar
+       end do
+    else
+       do i=1,ncol
+          upwp_sfc(i)   = cam_in%wsx(i)/rho_zt(i,2)               ! Surface meridional momentum flux
+          vpwp_sfc(i)   = cam_in%wsy(i)/rho_zt(i,2)               ! Surface zonal momentum flux
+       end do
+    endif
+
```
zarzycki, line 3047: Looks good to me! The only difference from my previous PR is I nested this block inside the 1,ncol loop from above. I am not sure if optimal practice is to separate the thermodynamic surface flux setting (e.g., wpthlp_sfc and wprtp_sfc) from the momentum fluxes into two loops as above or if there should be one ncol loop with a nested if statement for the mom fluxes.

bld/build-namelist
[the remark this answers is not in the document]
Katetc: Ok, so, based on Colin's comment here, I did add clubb_l_enable_relaxed_clipping to the namelist list with a default value of .false. that can be changed by the camdev experiments.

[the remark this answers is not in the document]
adamrher: ok thanks @Katetc and @zarzycki for catching this (before its too late!)

bld/namelist_files/namelist_definition.xml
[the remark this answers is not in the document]
Katetc: Ok, I removed all of the "Default" documentation from namelist_definition.xml except the "_history" fields which have a single default value and I think that is useful information in the webpage.

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4205,206 @@ subroutine clubb_tend_cam( &
...
+    call outfld( 'WP2_CLUBB',        wp2,                   pcols, lchnk )
+    call outfld( 'UP2_CLUBB',        up2,                   pcols, lchnk )
+    call outfld( 'VP2_CLUBB',        vp2,                   pcols, lchnk )
+    call outfld( 'WP3_CLUBB',        wp3_output,            pcols, lchnk )
+    call outfld( 'UPWP_CLUBB',       upwp,                  pcols, lchnk )
+    call outfld( 'VPWP_CLUBB',       vpwp,                  pcols, lchnk )
+    call outfld( 'WPTHLP_CLUBB',     wpthlp_output,         pcols, lchnk )
+    call outfld( 'WPRTP_CLUBB',      wprtp_output,          pcols, lchnk )
+
+    temp2dp(:ncol,:) =  rtp2(:ncol,:)*1.0e6_r8
```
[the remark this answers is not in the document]
Katetc: Done

src/physics/cam/clubb_intr.F90
Katetc: I don't think it makes a lot of difference one way or the other. This way makes the most sense to me based on the namelists and macmic switches needed. Also, I have a vague feeling that pushing loops as low as possible is best for GPUs. But I could be wrong there.

bld/namelist_files/namelist_defaults_cam.xml
bstephens82: Resolved.

bstephens82 (approved):
I didn't find anything that I had questions or comments about...thanks Kate!

bld/build-namelist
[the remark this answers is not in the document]
adamrher: @Katetc I'm still not seeing l_enable_relaxed_clipping in the namelist?

src/physics/cam/clubb_intr.F90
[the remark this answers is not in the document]
adamrher: this has been fixed as Kate merged my PR into this PR branch.

bld/build-namelist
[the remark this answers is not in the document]
Katetc: Well, there was one unpushed commit on my branch but I think it was my merge. So, I don't know if this commit was there before but the namelist was added [in this commit
](https://github.com/ESCOMP/CAM/pull/712/commits/9dd39370c5b519804479c6e160c85b9782c0e940)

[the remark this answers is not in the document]
adamrher: i see it now. thx

bld/namelist_files/namelist_definition.xml
```diff
@@ -3671,131 +3642,135 @@ the damping of CLUBB's wp3 when skewness of w (vertical velocity) is large in
...
+Coefficient in the pressure-turbulence term of CLUBB's wp3 predictive equation.
 </entry>
 
-<entry id="clubb_C_invrs_tau_N2_xp2" type="real" category="pblrad"
+<entry id="clubb_detice_rad" type="real" category="pblrad"
        group="clubb_params_nl" valid_values="" >
-Coefficient of inverse tau term contributed by Brunt Vaisala frequency but for xp2_wpxp (units: none)
-Default: 0.2
+Radius of detrained ice drops as they are used in the CLUBB parameterization in units of (m).
+Default: 25.0e-6 m
```
cacraigucar, line 3743: This is the default for CAM runs, but cam_dev has a different default.  This comment will most likely be wrong for CAM7

cacraigucar (changes requested):
One minor change requested

src/physics/cam/clubb_intr.F90
adamrher: As I was noodling around with this coed block I realized I broke this clubbtop / energy fixer in cam6_3_066, and have opened an issue: https://github.com/ESCOMP/CAM/issues/739. Resolving this convo without full testing since it won't pass the energy fixer until this new issue is resolved in another PR.

src/physics/cam/clubb_intr.F90
```diff
@@ -3824,229 +3888,240 @@ subroutine clubb_tend_cam( &
...
+    ! ------------------------------------------------- !
+    ! Diagnose some output variables                    !
+    ! ------------------------------------------------- !
 
-   !  density
-   rho(1:ncol,1:pver) = rga*state1%pdel(1:ncol,1:pver)/(state1%zi(1:ncol,1:pver)-state1%zi(1:ncol,2:pverp))
-   rho(1:ncol,pverp) = rho(1:ncol,pver)
+    !  density
+    rho(:ncol,1:pver) = state1%pmid(:ncol,1:pver)/(rairv(:ncol,1:pver,lchnk)*state1%t(:ncol,1:pver))
+    rho(:ncol,pverp)  = state1%ps(:ncol)/(rairv(:ncol,pver,lchnk)*state1%t(:ncol,pver))
```
nusbaume, line 4022: Should these be explicitly looped over instead of using array slicing?  I noticed there were changes above where array slices were replaced with explicit loops, and it seems like doing that here would make things more consistent.  Plus it might make the actual code a little easier to read.

src/physics/cam/clubb_intr.F90
```diff
@@ -3824,229 +3888,240 @@ subroutine clubb_tend_cam( &
...
-         wpthlp_output(i,k)  = (wpthlp(i,k)-(apply_const*wpthlp_const))*rho(i,k)*cpair !  liquid water potential temperature flux
-         wprtp_output(i,k)   = (wprtp(i,k)-(apply_const*wprtp_const))*rho(i,k)*latvap  !  total water mixig ratio flux
-         rtpthlp_output(i,k) = rtpthlp(i,k)-(apply_const*rtpthlp_const)                !  rtpthlp output
-         wp3_output(i,k)     = wp3(i,k) - (apply_const*wp3_const)                      !  wp3 output
-         tke(i,k)            = 0.5_r8*(up2(i,k)+vp2(i,k)+wp2(i,k))                     !  turbulent kinetic energy
-         if (do_clubb_mf) then
-           mf_thlflx_output(i,k) = mf_thlflx_output(i,k)*rho(i,k)*cpair
-           mf_qtflx_output(i,k)  = mf_qtflx_output(i,k)*rho(i,k)*latvap
-         end if
+        eps = rairv(i,k,lchnk)/rh2o
```
nusbaume, line 4027: Can we make a `1._r8/rh2o` parameter and use that here instead of doing the division every time?

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4132,200 @@ subroutine clubb_tend_cam( &
...
-         thv(i,k) = th(i,k)*(1.0_r8+zvir*state1%q(i,k,ixq) - state1%q(i,k,ixcldliq))
+        th(i,k) = state1%t(i,k)*state1%exner(i,k)
+        thv(i,k) = th(i,k)*(1.0_r8+zvir*state1%q(i,k,ixq))
       enddo
-   enddo
+    enddo
  
-   call calc_ustar( ncol, state1%t(1:ncol,pver), state1%pmid(1:ncol,pver), cam_in%wsx(1:ncol), cam_in%wsy(1:ncol), &
+    ! diagnose surface friction and obukhov length (inputs to diagnose PBL depth)
+    rrho(1:ncol) = (1._r8/gravit)*(state1%pdel(1:ncol,pver)/dz_g(1:ncol,pver))
```
nusbaume, line 4190: You could replace `1._r8/gravit` with `rga` from `physconst` here.

nusbaume (approved):
Looks good!  Just some very minor code clean-up stuff.

adamrher (approved):
Sorry Kate, I found a handful of clubb_intr statements that were being reverted back to our old clubb_intr. Approving, assuming you will address them as the are straight forward.

src/physics/cam/clubb_intr.F90
```diff
@@ -2891,20 +3016,42 @@ subroutine clubb_tend_cam( &
...
       wpthlp_sfc(i) = cam_in%shf(i)/(cpairv(i,pver,lchnk)*rho_zt(i,2))! Sensible heat flux
       wpthlp_sfc(i) = wpthlp_sfc(i)*inv_exner_clubb_surf(i)   ! Potential temperature flux
       wprtp_sfc(i)  = cam_in%cflx(i,1)/rho_zt(i,2)            ! Moisture flux  (check rho)
-      upwp_sfc(i)   = cam_in%wsx(i)/rho_zt(i,2)               ! Surface meridional momentum flux
-      vpwp_sfc(i)   = cam_in%wsy(i)/rho_zt(i,2)               ! Surface zonal momentum flux  
     end do
+
+    !  Other Surface fluxes provided by host model
+    if( (cld_macmic_num_steps > 1) .and. (clubb_l_intr_sfc_flux_smooth) ) then
+       ! Adjust surface stresses using winds from the prior macmic iteration
```
adamrher, line 3034: based on internal discussions with Colin and Thomas, we need to add this statement below:
"! based on Thomas Toniazzo's implementation in NorESM"

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1862,75 +1867,81 @@
...
 <clubb_l_min_xp2_from_corr_wx                     > .true.  </clubb_l_min_xp2_from_corr_wx>
+<clubb_l_mono_flux_lim_rtm                        > .true.  </clubb_l_mono_flux_lim_rtm>
+<clubb_l_mono_flux_lim_spikefix                   > .true.  </clubb_l_mono_flux_lim_spikefix>
+<clubb_l_mono_flux_lim_thlm                       > .true.  </clubb_l_mono_flux_lim_thlm>
+<clubb_l_mono_flux_lim_um                         > .false. </clubb_l_mono_flux_lim_um>
+<clubb_l_mono_flux_lim_vm                         > .false. </clubb_l_mono_flux_lim_vm>
 <clubb_l_predict_upwp_vpwp                        > .false. </clubb_l_predict_upwp_vpwp>
 <clubb_l_rcm_supersat_adj                         > .false. </clubb_l_rcm_supersat_adj>
+<clubb_l_smooth_Heaviside_tau_wpxp                > .false. </clubb_l_smooth_Heaviside_tau_wpxp>
 <clubb_l_stability_correct_tau_zm                 > .true.  </clubb_l_stability_correct_tau_zm>
```
zarzycki, line 1924: FWIW, this is the set of defaults via build-namelist + namelist_defaults.xml that disagree with what we (cc: @bstephens82) have in: https://github.com/NCAR/amwg_dev/discussions/106#discussioncomment-3614557

If I add...

```
clubb_l_diag_lscale_from_tau=.true.
clubb_l_predict_upwp_vpwp=.true.
```

to user_nl_cam with `-phys cam_dev`

(new.txt is this PR, old.txt is my hand-modified sandbox)

```
zarzycki@cheyenne5:~/scratch/FHIST-ne30-free-x321/run> diff old.txt new.txt 
87c87
<  C_invrs_tau_N2_xp2           =   0.800000000000000     
---
>  C_invrs_tau_N2_xp2           =   5.000000000000000E-002
99a100
>  a3_coef_min                  =    1.00000000000000     
102a104,105
>  penta_solve_method =            1
>  tridiag_solve_method =            1
105,106c108,109
<  l_min_wp2_from_corr_wx =  F
<  l_min_xp2_from_corr_wx =  F
---
>  l_min_wp2_from_corr_wx =  T
>  l_min_xp2_from_corr_wx =  T
116c119
<  l_vert_avg_closure =  F
---
>  l_vert_avg_closure =  T
120,123c123,126
<  l_standard_term_ta =  T
<  l_partial_upwind_wp3 =  T
<  l_godunov_upwind_wpxp_ta =  T
<  l_godunov_upwind_xpyp_ta =  T
---
>  l_standard_term_ta =  F
>  l_partial_upwind_wp3 =  F
>  l_godunov_upwind_wpxp_ta =  F
>  l_godunov_upwind_xpyp_ta =  F
129c132
<  l_stability_correct_tau_zm =  F
---
>  l_stability_correct_tau_zm =  T
144c147
<  l_vary_convect_depth T
---
>  l_vary_convect_depth F
147c150,157
<  l_smooth_Heaviside_tau_wpxp =  T
---
>  l_smooth_Heaviside_tau_wpxp =  F
>  l_enable_relaxed_clipping =  F
>  l_linearize_pbl_winds =  F
>  l_mono_flux_lim_thlm =  T
>  l_mono_flux_lim_rtm =  T
>  l_mono_flux_lim_um =  F
>  l_mono_flux_lim_vm =  F
>  l_mono_flux_lim_spikefix =  T
```

Do we want to try and toggle these booleans when clubb_l_diag_Lscale_from_tau is true?

src/physics/cam/clubb_intr.F90
Katetc: There's actually several lines of code above these that weren't changed to explicit loops, so I think, in context, it makes sense. Also, Adam had an important request about the code in these lines below, which I think takes precedence here. I want the code to look the way he left it in his clean-up PR. I suspect it will eventually be turned to loops, especially as the GPU support comes in this year.

cacraigucar (approved):

src/physics/cam/clubb_intr.F90
```diff
@@ -190,8 +184,11 @@ module clubb_intr
                                  ! (double Gaussian) PDF type to use for the w, rt,
                                  ! and theta-l (or w, chi, and eta) portion of
                                  ! CLUBB's multivariate, two-component PDF.
-    clubb_ipdf_call_placement = unset_i    ! Selected option for the placement of the call to
+    clubb_ipdf_call_placement = unset_i, & ! Selected option for the placement of the call to
                                            ! CLUBB's PDF.
+    clubb_penta_solve_method = unset_i,  & ! Specifier for method to solve then penta-diagonal system
```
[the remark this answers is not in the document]
Katetc: Fixed

src/physics/cam/clubb_intr.F90
```diff
@@ -295,12 +292,21 @@ module clubb_intr
     clubb_l_use_tke_in_wp3_pr_turb_term,& ! Use TKE formulation for wp3 pr_turb term
     clubb_l_use_tke_in_wp2_wp3_K_dfsn,  & ! Use TKE in eddy diffusion for wp2 and wp3
     clubb_l_smooth_Heaviside_tau_wpxp,  & ! Use smooth Heaviside 'Peskin' in computation of invrs_tau
+    clubb_l_enable_relaxed_clipping,    & ! Flag to relax clipping on wpxp in xm_wpxp_clipping_and_stats
+    clubb_l_linearize_pbl_winds,        & ! Code to linearize PBL winds
```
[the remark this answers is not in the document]
Katetc: Added some "Flag" language

src/physics/cam/clubb_intr.F90
```diff
@@ -1840,771 +1928,801 @@ end subroutine clubb_ini_cam
...
+    real(r8) :: qclvar(pcols,pverp)              ! cloud water variance                          [kg^2/kg^2]
+    real(r8) :: zo(pcols)                               ! roughness height                              [m]
+    real(r8) :: dz_g(pcols,pver)                       ! thickness of layer                            [m]
+    real(r8) :: relvarmax
+
+    ! Local CLUBB variables dimensioned as NCOL (only useful columns) to be sent into the clubb run api
+    ! NOTE: THESE VARIABLS SHOULD NOT BE USED IN PBUF OR OUTFLD (HISTORY) SUBROUTINES
+    real(r8), dimension(state%ncol) :: &
+      fcor, &                  ! Coriolis forcing 			      	[s^-1]
+      sfc_elevation, &    		          ! Elevation of ground			      	[m AMSL]			      	[m]
```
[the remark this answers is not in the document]
Katetc: Lined everything up in an aesthetically pleasing manor.

src/physics/cam/clubb_intr.F90
```diff
@@ -2891,20 +3016,42 @@ subroutine clubb_tend_cam( &
...
     do i=1,ncol                                                                  
       wpthlp_sfc(i) = cam_in%shf(i)/(cpairv(i,pver,lchnk)*rho_zt(i,2))! Sensible heat flux
       wpthlp_sfc(i) = wpthlp_sfc(i)*inv_exner_clubb_surf(i)   ! Potential temperature flux
       wprtp_sfc(i)  = cam_in%cflx(i,1)/rho_zt(i,2)            ! Moisture flux  (check rho)
-      upwp_sfc(i)   = cam_in%wsx(i)/rho_zt(i,2)               ! Surface meridional momentum flux
-      vpwp_sfc(i)   = cam_in%wsy(i)/rho_zt(i,2)               ! Surface zonal momentum flux  
     end do
+
+    !  Other Surface fluxes provided by host model
+    if( (cld_macmic_num_steps > 1) .and. (clubb_l_intr_sfc_flux_smooth) ) then
```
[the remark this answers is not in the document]
Katetc: Removed

src/physics/cam/clubb_intr.F90
Katetc: Completed

src/physics/cam/clubb_intr.F90
Katetc: Done

src/physics/cam/clubb_intr.F90
```diff
@@ -3824,229 +3888,240 @@ subroutine clubb_tend_cam( &
...
-   ! ------------------------------------------------- !
+    ! ------------------------------------------------- !
+    ! Diagnose some output variables                    !
+    ! ------------------------------------------------- !
 
-   !  density
-   rho(1:ncol,1:pver) = rga*state1%pdel(1:ncol,1:pver)/(state1%zi(1:ncol,1:pver)-state1%zi(1:ncol,2:pverp))
-   rho(1:ncol,pverp) = rho(1:ncol,pver)
+    !  density
+    rho(:ncol,1:pver) = state1%pmid(:ncol,1:pver)/(rairv(:ncol,1:pver,lchnk)*state1%t(:ncol,1:pver))
```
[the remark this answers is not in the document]
Katetc: Sorry about this. I reverted this to the more recent code

src/physics/cam/clubb_intr.F90
```diff
@@ -3824,229 +3888,240 @@ subroutine clubb_tend_cam( &
...
-   ! --------------------------------------------------------------------------------- ! 
+    ! --------------------------------------------------------------------------------- !  
+    !  THIS PART COMPUTES CONVECTIVE AND DEEP CONVECTIVE CLOUD FRACTION                 !
+    ! --------------------------------------------------------------------------------- ! 
  
-   ! Initialize cloud fraction
-   deepcu(:,:) = 0.0_r8
-   shalcu(:,:) = 0.0_r8
+    deepcu(:,pver) = 0.0_r8
+    shalcu(:,pver) = 0.0_r8
```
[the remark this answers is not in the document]
Katetc: Reverted, thanks for catching it!

src/physics/cam/clubb_intr.F90
```diff
@@ -4057,203 +4132,200 @@ subroutine clubb_tend_cam( &
...
+    !  this is needed for aerosol code                                                  !
+    ! --------------------------------------------------------------------------------- !       
+    do i=1,ncol
       do k=1,pver
-         !use local exner since state%exner is not a proper exner
-         th(i,k) = state1%t(i,k)*inv_exner_clubb(i,k)
-         !thv should have condensate loading to be consistent with earlier def's in this module
-         thv(i,k) = th(i,k)*(1.0_r8+zvir*state1%q(i,k,ixq) - state1%q(i,k,ixcldliq))
+        th(i,k) = state1%t(i,k)*state1%exner(i,k)
+        thv(i,k) = th(i,k)*(1.0_r8+zvir*state1%q(i,k,ixq))
```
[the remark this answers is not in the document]
Katetc: Reverted

src/physics/cam/clubb_intr.F90
Katetc: Added

bld/namelist_files/namelist_defaults_cam.xml
adamrher: @zarzycki it had occurred to me to do that. But I went immediately to (a) there's no doubt that these tunings will change, and add/subtract more namelist values not incl. in your list, so we should probably wait to hard code these until we have a more final configuration. (b) weariness about using Lscale flag to toggle on all these non Lscale namelists. (c) let's just finish this PR already --- we're going to have to do more clubb_intr PR's over the next months anyways (e.g., I broke the energy fixer: https://github.com/ESCOMP/CAM/issues/739).

I'm fully expecting to have to relay a pretty loaded user_nl_cam w/ preferred namelists to Cecile for our CAM7 evaluations.

zarzycki: I think that is reasonable -- FWIW, I just tested and we can toggle everything except `l_vary_convect_depth`, `l_partial_upwind_wp3`, and `l_standard_term_ta` with simple user_nl_cam switches. Those 3 appear to be exposed in clubb_intr.F90 but do **not** have matches in namelist_definition.xml, so preview_namelist fails if one tries to set them in the case dir.

A couple quick thoughts. If @bstephens82 or @vlarson **absolutely** thinks we should change any of these bools in namelist_defaults, we should just do it, IMO.

If we could at least expose the above 3 variables to be toggled in user_nl_cam (by adding in build-namelist, namelist_defaults, and namelist_definition?) that would be extremely helpful, because then we can just inject a block of text for users to repro whatever our latest config is. Currently I hardcode them with CLUBB source mods, which is suboptimal.

adamrher: I 100% agree to expose those three namelists at the cam level. ```l_vary_convect_depth``` is very important to control for CLASP folks.

vlarson: Putting `l_vary_convect_depth`, `l_partial_upwind_wp3`, and `l_standard_term_ta` in the namelists sounds good to me.

src/physics/cam/clubb_intr.F90
adamrher: sorry we've moved the goal posts - it is now

"! implementation after Thomas Toniazzo (NorESM) and Colin Zarzycki (PSU)"

bld/namelist_files/namelist_defaults_cam.xml
zarzycki: OK, so perhaps we add them to CAM's namelist like other CLUBB logicals and have their defaults (for now) be:

```
l_vary_convect_depth F
l_standard_term_ta =  F
l_partial_upwind_wp3 =  F
```

which should match the internal CLUBB defaults. Then if we want to evaluate them moving forward they can be toggled in user_nl_cam. It is much easier to have a block of options carried about in user_nl_cam vs. asking folks to do sourcemods. Then in the future if we are 100% sure we want to switch them with the taus code we can do so in namelist_defaults (future PR).

Is that reasonable/doable, @Katetc?

Katetc: This sounds like a great plan to me. Thanks, guys, for the quick discussion!

src/physics/cam/clubb_intr.F90
Katetc: Updated!

zarzycki (approved):

PR #722, opened 2022-12-13
Update workflow

jedwards4b:
github workflow versions need to be updated, the current versions are deprecated and will soon no longer function.


nusbaume (approved):
Good catch, thanks!

Issue #627, opened 2022-07-26
Create a github repo for SPCAM and add it as a CAM external


cacraigucar:
### Issue Type

Infrastructure Update

### Issue Description

Currently SPCAM is part of the CAM code base.  The proposal is to create a github repo for SPCAM and be brought in via a CAM external.  This will mimic what is currently done for CLUBB and PUMAS.

### Will this change answers?

No

### Will you be implementing this yourself?

No

PR #725, opened 2022-12-13
part of cam6_3_087: Turn off unit 155 ECPP writing by default

### LINKED ISSUE #717: spcam output cleanup
### What happened?

spcam, specifically file  src/physics/spcam/ecpp/module_ecpp_td2clm.F90

is outputting a file to the run directory fort.155 in test 
ERS_Ln9_Vnuopc.f19_f19_mg17.FSPCAMM.cheyenne_intel.cam-outfrq9s

this should be removed or used only for DEBUG mode and renamed to something more easily identified.

test ERS_Ln9_Vnuopc.f19_f19_mg17.FSPCAMM.cheyenne_intel.cam-outfrq9s

cam6_3_084

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

_No response_

cacraigucar:
Closes #717

Note that this was just turned off like all of the other optional writing commands in ECPP with a -1.

This code mod was tested before and after making the change using ` ./create_test ERS_Ln9_Vnuopc.f19_f19_mg17.FSPCAMM.izumi_gnu.cam-outfrq9s` and the fort.155 file was not written out once the change was made.

### DIFF
--- a/src/physics/spcam/ecpp/module_ecpp_ppdriver2.F90
+++ b/src/physics/spcam/ecpp/module_ecpp_ppdriver2.F90
@@ -703,28 +703,7 @@ end type ptr2d_t
 !	and cloudy subareas
 
 !
-	idiagaa_ecpp(:) = 0
-!        idiagaa_ecpp(60:63) = 1
-        idiagaa_ecpp(60:63) = -1
-	idiagaa_ecpp(115:119) = 1 ; idiagaa_ecpp(118) = 111
-	idiagaa_ecpp(121:125) = 1
-        idiagaa_ecpp(131:135) = 1
-        idiagaa_ecpp(141:143) = 1
-
-	idiagaa_ecpp(155) = -1
-	idiagaa_ecpp(161) = 1 ; idiagaa_ecpp(162) = 1 ; idiagaa_ecpp(164) = 1
-
-        idiagaa_ecpp(131:135) = -1  ! not output in the MMF model
-        idiagaa_ecpp(115:117) = -1  ! not dump the original field in parampollu_td240clm
-        idiagaa_ecpp(118:119) = -1
-        idiagaa_ecpp(121:125) = -1
-        idiagaa_ecpp(141:143) = -1
-        idiagaa_ecpp(165:167) = -1
-        idiagaa_ecpp(164) = -1
-        idiagaa_ecpp(161) = -1
-        idiagaa_ecpp(162) = -1
-
-        idiagaa_ecpp(121) = -1
+	idiagaa_ecpp(:) = -1
 
 	do i = 1, 199
 	    ldiagaa_ecpp(i) = i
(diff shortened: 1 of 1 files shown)


src/physics/spcam/ecpp/module_ecpp_ppdriver2.F90
```diff
@@ -711,7 +711,7 @@ subroutine parampollu_driver2( aero_props,                &
         idiagaa_ecpp(131:135) = 1
         idiagaa_ecpp(141:143) = 1
 
-	idiagaa_ecpp(155) = 1
+	idiagaa_ecpp(155) = -1
```
fvitt, line 714: Is there a reason to not turn off all these diagnostics?

fvitt (commented):
Has this actually been tested?

src/physics/spcam/ecpp/module_ecpp_ppdriver2.F90
cacraigucar: Yes, it was tested (as I've described in the comment above) and the only "fort" file which was written before the mod was fort.155.  fort.155 was not created after the mod.  I guess the rest of the code was not being reached.  

I debated changing all of the settings, but since 155 was the only culprit, I made my change just to it.  If you would like for me to set the others, I can easily do so.

nusbaume: It looks like all of those other diagnostics are turned off immediately below this code block.  Maybe it would make sense to delete/combine these code blocks to avoid confusion (and needless repetition)?

cacraigucar: All of the SPCAM code does not follow our coding practices at all, but we also are not the maintainers of the code.  My approach to code changes in SPCAM was to make minimal changes and only as necessary so that the code retained its "look and feel" for the folks who wrote it.  Changing the value to -1 in this location seemed like it maintained the way that the developers would turn on/off this particular output (like line 708 did above and 717 did below).   For this reason, I thought this was the best place to make the change.  I don't believe that we should make any change other than what I made.  I am willing to be convinced otherwise by either @fvitt or @nusbaume

fvitt (approved):
I don't want to spend any time trying to figure out these diagnostic switches in this code. As long as the change proposed here turns off the fort.155 output and assuming there are no other fort.* files written then I am OK with the PR.

src/physics/spcam/ecpp/module_ecpp_ppdriver2.F90
nusbaume: While I agree that in theory we are not the maintainers of this code, I would argue that in practice we are, at least until issue #627 is resolved (and the fact that you are doing this PR at all).  Plus if cleaning up what clearly looks like copy/paste code is enough to throw off the maintainers then should they really be maintainers in the first place?

nusbaume (approved):
Looks good to me now, thanks!

cacraigucar:
This version was tested and determined that no fort.xx files were created nor did the cam.h0 differ from what it was previously (did not expect a change, but this was simple to verifiy)

PR #747, opened 2023-02-03
part of cam6_3_108: Cloud to ground lightning flash frequency passed to land model

### LINKED ISSUE #567: Pass lightning flash frequency to mediator for CTSM
See discussions in 
https://github.com/ESCOMP/CTSM/issues/1630

As suggested by @ekluzek, should set a flag in drv_flds_in so that surface models know that this is available.

fvitt:
Closes #567


src/chemistry/mozart/mo_lightning.F90
```diff
@@ -267,45 +340,63 @@ subroutine lightning_no_prod( state, pbuf2d,  cam_in )
...
-             !--------------------------------------------------------------------------------
-             ! 	... compute global no production rate in tgn/yr:
-             !           tgn per second: * 14.00674 * 1.65979e-24 * 1.e-12
-             !             nb: 1.65979e-24 = 1/avo
-             !           tgn per year: * secpyr
-             !--------------------------------------------------------------------------------
-             glob_prod_no_col(i,c) = 1.e17_r8*flash_energy(i,c) &
-                  * 14.00674_r8 * 1.65979e-24_r8 * 1.e-12_r8 * secpyr * factor
 
+             cld2grnd_flash_freq(i) = cam_in(c)%landfrac(i)*flash_freq_land *cgic(i,c) * factor ! cld-to-grnd flash frq (per min)
```
fvitt, line 356: @lkemmons Does this formulation look correct for the cloud-to-ground lightning flash frequency?  Is it correct to include `factor` here which is `0.1_r8*lght_no_prd_factor`, the namelist lightning NOx production rate tuning parameter.

lkemmons: No, 'factor' is just for getting the right amount of NO emissions per flash, and should not be applied to the flash frequency that is passed to the land.

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -39,109 +108,103 @@ subroutine lightning_inti( lght_no_prd_factor )
...
+       vdist(:,3) = (/  2.0_r8, 2.0_r8, 2.0_r8, 1.5_r8, 1.5_r8, 1.5_r8, 3.0_r8, 5.8_r8, &       ! trop cont
+            7.6_r8, 9.6_r8, 11.0_r8, 14.0_r8, 14.0_r8, 14.0_r8, 8.2_r8, 2.3_r8 /)
+
+       allocate( prod_no(pcols,pver,begchunk:endchunk),stat=astat )
+       if( astat /= 0 ) then
+          write(iulog,*) prefix, 'failed to allocate prod_no; error = ',astat
+          call endrun
+       end if
+       geo_factor = ngcols_p/(4._r8*pi)
+
```
lkemmons, line 159: The descriptions of these 3 fields should say 'lightning' not 'lighting'.

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -39,109 +108,103 @@ subroutine lightning_inti( lght_no_prd_factor )
...
+       call addfld( 'LNO_COL_PROD', horiz_only,  'I', 'TG N/YR', 'lightning column NO source' )
+       call addfld( 'LNO_PROD',     (/ 'lev' /), 'I', '/cm3/s',  'lightning insitu NO source' )
+       call addfld( 'FLASHENGY',    horiz_only,  'I', 'J',       'lightning flash energy' ) ! flash energy
```
fvitt, line 162: @lkemmons What should the units be on the flash energy?

lkemmons: The flash energy is in joules (J).
Could we change LNO_PROD units to 'molecules/cm3/s' - I think that is consistent with other chemical rates?

wwieder:
This PR caught the attention of the LMWG. Thanks for working on this, @fvitt!  

We're wondering what the timeline for this PR may be, as there's quite a bit of work this could spark on the land side if we can get interactive lightning (and therefore fire ignitions) into CESM3?

Externals.cfg
```diff
@@ -21,9 +21,9 @@ externals = Externals.cfg
 required = True
 
 [cmeps]
-tag = cmeps0.14.5
+branch = lightning_coupling
 protocol = git
-repo_url = https://github.com/ESCOMP/CMEPS.git
+repo_url = https://github.com/fvitt/CMEPS.git
```
cacraigucar, line 26: When will this come into CMEPS?

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -10,27 +10,96 @@ module mo_lightning
...
+    call mpi_bcast(lght_no_prd_factor, 1, mpi_real8, masterprocid, mpicom, ierr)
+    if (ierr/=mpi_success) then
+       call endrun(subname//': MPI_BCAST ERROR: lght_no_prd_factor')
+    end if
+
+    if (masterproc) then
+       write(iulog,*) subname,' lght_no_prd_factor: ',lght_no_prd_factor
+    end if
+
+    factor = 0.1_r8*lght_no_prd_factor
```
cacraigucar, line 86: Why are you scaling this factor instead of using directly the namelist setting?

cacraigucar (changes requested):
Preliminary review - may have additional comments, but wanted to get a couple of questions answered

src/chemistry/mozart/mo_lightning.F90
fvitt: @lkemmons Do you know the rationale for scaling down `lght_no_prd_factor` here. This is pre-existing code and I don't know the reason for it.  I would like the changes in this PR to be bit-for-bit type changes.   If scaling `lght_no_prd_factor` namelist settings by a factor of 0.1 does not cause any baseline test to fail then I am okay with it.  Otherwise, I would advocate leaving it as it is for this PR.

lkemmons: @fvitt I do not recall either why we included 0.1 in the code for additional scaling to the lght_no_prd_factor in the namelist.  Maybe 0.1 was the optimal value at some point for some configuration.  
If we remove the 0.1 factor here we will need to change all the default CAM-chem and WACCM namelists.  I agree it would be simpler to have a single factor and remove the 0.1, but I also agree with Francis it would be better to NOT change it at this point.

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -39,109 +111,103 @@ subroutine lightning_inti( lght_no_prd_factor )
...
+
+    cldtop_ndx = pbuf_get_index('CLDTOP',errcode=err)
+    cldbot_ndx = pbuf_get_index('CLDBOT',errcode=err)
+    calc_lightning = cldtop_ndx>0 .and. cldbot_ndx>0
+
+    if (.not.calc_lightning) return
+
+    calc_nox_prod = flsh_frq_ndx>0
+
+    if (calc_nox_prod) then
```
cacraigucar, line 136: Won't `calc_nox_prod` always be true if you reach this point?  I'm not seeing that the test `    if (calc_nox_prod) then` is needed here.  I also have the same question for line 274 below.

cacraigucar (changes requested):

src/chemistry/mozart/mo_lightning.F90
fvitt: Changed how `calc_nox_prod` is set.

brian-eaton (commented):
Looks good.  A couple of minor comments.

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -10,140 +10,205 @@ module mo_lightning
...
-  public  :: lightning_inti
+
+  public  :: lightning_readnl
+  public  :: lightning_register
+  public  :: lightning_init
   public  :: lightning_no_prod
   public  :: prod_no
 
-  save
+  real(r8),protected, allocatable :: prod_no(:,:,:)
```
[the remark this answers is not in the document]
fvitt: `prod_no` is public (always has been).  See line 24.  It is used in `src/chemistry/mozart/mo_setext.F90`.  Isn't it a good idea to protect it since it is public?

nusbaume (changes requested):
Looks good, just have one small change request and a question.

bld/build-namelist
```diff
@@ -4058,6 +4058,12 @@ if ($opts{'cmeps'}) {
   } else {
     add_default($nl, 'atm_ozone_frequency', 'val'=>'multiday_average');
   }
+  # for lightning flash freq to CTSM
+  if ($simple_phys or $aqua_mode) {
+    add_default($nl, 'atm_provides_lightning', 'val'=>'.false.');
+  } else {
+    add_default($nl, 'atm_provides_lightning', 'val'=>'.true.');
+  }
```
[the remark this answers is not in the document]
fvitt: done

src/chemistry/mozart/mo_lightning.F90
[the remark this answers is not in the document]
brian-eaton: @fvitt, sorry I missed that (but so did Steve:).  Stylewise I think it's preferable to have all the attributes for a module variable in one statement.  But a change here isn't necessary.

brian-eaton (approved):

jtruesdal (approved):
Looks Ok to me.

src/chemistry/mozart/mo_lightning.F90
```diff
@@ -10,140 +10,205 @@ module mo_lightning
...
+            5.0_r8, 7.0_r8, 9.0_r8, 14.0_r8, 16.0_r8, 14.0_r8, 8.0_r8, 0.5_r8 /)
+       vdist(:,2) = (/  2.5_r8, 2.5_r8, 2.5_r8, 2.5_r8, 2.5_r8, 2.5_r8, 2.5_r8, 6.1_r8, &       ! trop marine
+            17.0_r8, 15.4_r8, 14.5_r8, 13.0_r8, 12.5_r8, 2.8_r8, 0.9_r8, 0.3_r8 /)
+       vdist(:,3) = (/  2.0_r8, 2.0_r8, 2.0_r8, 1.5_r8, 1.5_r8, 1.5_r8, 3.0_r8, 5.8_r8, &       ! trop cont
+            7.6_r8, 9.6_r8, 11.0_r8, 14.0_r8, 14.0_r8, 14.0_r8, 8.2_r8, 2.3_r8 /)
+
+       allocate( prod_no(pcols,pver,begchunk:endchunk),stat=astat )
+       if( astat /= 0 ) then
+          write(iulog,*) prefix, 'failed to allocate prod_no; error = ',astat
+          call endrun
```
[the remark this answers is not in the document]
fvitt: fixed

src/control/camsrfexch.F90
```diff
@@ -302,6 +303,7 @@ subroutine atm2hub_alloc( cam_out )
        cam_out(c)%co2prog(:)  = 0._r8
        cam_out(c)%co2diag(:)  = 0._r8
        cam_out(c)%ozone(:)    = 0._r8
+       cam_out(c)%lightning_flash_freq(:) = -huge(1._r8)
```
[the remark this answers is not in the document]
fvitt: changed to 0.

cacraigucar (approved):

nusbaume (approved):

PR #766, opened 2023-03-17
cam6_3_100: add ghg_mam4 chemistry

### LINKED ISSUE #762: New chemistry mechanism for CAM7 prognostic GHGs

A WACCM-SC like mechanism is needed which will include prognostic CO2

@dan800 @lkemmons @tilmes

Yes

fvitt:
closes #762


brian-eaton (approved):
Looks good to me.

jtruesdal (approved):

adamrher:
I'm removing myself from the review as code-wise it's a bit out of my wheel-house, and so won't be able to spend the time I would need to review this properly.

I would ask that we clarify the LBC file(s) for use with ghg_mam4 --- if not in this PR, open a separate issue. As @dan800 and I just discussed, ghg_mam4 **should** point to a seasonally and latitudinally varying CO2_LBC. I could see holding off on making this decision until we've evaluated what the CO2 actually looks like in simulations w/ ghg_mam4. If that sounds reasonable I can just open the issue. If not, I'm all ears.

adamrher:
Thanks, Francis. That resolves my concerns. Carry on!

cacraigucar (approved):

PR #780, opened 2023-04-03
Initialize CO2 when not available in initial conditions file

### LINKED ISSUE #779: When CO2 is not in IC file, needs to initialize to a non-zero value
### What happened?

During Cecile's testing ( https://mail.google.com/mail/u/0/#inbox/FMfcgzGsltHJdwHmKsJSkrFBrKzmtmxF ), it was discovered that CO2 initializes to 0 when it is not in an initial condition file.  

Francis proposed the following fix:

In chem_init_cnst we have:

    if ( ghg_chem ) then
       do ilev = 1, nlev
          select case (name)
          case ('N2O')
             where(mask)
                q(:,ilev) = rmwn2o * chem_surfvals_get('N2OVMR')
             end where
          case ('CH4')
             where(mask)
                q(:,ilev) = rmwch4 * chem_surfvals_get('CH4VMR')
             end where
          case ('CFC11')
             where(mask)
                q(:,ilev) = rmwf11 * chem_surfvals_get('F11VMR')
             end where
          case ('CFC12')
             where(mask)
                q(:,ilev) = rmwf12 * chem_surfvals_get('F12VMR')
             end where
          end select
       end do
    end if
So CH4 is being initialized according the the LBC.
It looks like CO2 is being intialized to 1.e-38. It would be easy to add CO2 to the above code block.

Since there is already a 'CO2MMR' option, you could add the case like this:

          case ('CO2')
             where(mask)
                q(:,ilev) = chem_surfvals_get('CO2MMR')
             end where

See: https://mail.google.com/mail/u/0/#inbox/FMfcgzGsltHJdwHmKsJSkrFBrKzmtmxF

cam6_3_100

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

_No response_

cacraigucar:
Closes #779

### DIFF
--- a/src/chemistry/mozart/chemistry.F90
+++ b/src/chemistry/mozart/chemistry.F90
@@ -978,6 +978,10 @@ end function chem_is_active
              where(mask)
                 q(:,ilev) = rmwf12 * chem_surfvals_get('F12VMR')
              end where
+          case ('CO2')
+             where(mask)
+                q(:,ilev) = chem_surfvals_get('CO2MMR')
+             end where
           end select
        end do
     end if


nusbaume (approved):
Whew, this was a tough one!

fvitt:
Do we know if this gives the desired behavior?  I am curious to hear about @cecilehannay 's test with this change.

cacraigucar:
> Do we know if this gives the desired behavior? I am curious to hear about @cecilehannay 's test with this change.

@cecilehannay is just starting up the run.  My plan is to get this PR ready so that if/when she says the code change works, we can bring this in quickly.  I am currently planning on combining this PR with #777 which @brian-eaton is doing the initial review on.   I have added @cecilehannay as a reviewer so she can give us an approval when her job completes and she confirms that CO2 is being initialized as it should be.

adamrher:
@fvitt. @dan800 has confirmed these mods give the desired behavior here: https://github.com/NCAR/amwg_dev/issues/232#issuecomment-1496203672

brian-eaton (approved):

Issue #757, opened 2023-02-28
Introduce in-line photolysis rate calculations from TUV-x


mattldawson:
### What is the feature/what would you like to discuss?

This enhancement would use [TUV-x](https://github.com/NCAR/tuv-x) to calculate photolysis rates.

Science motivation: CESM currently uses a look-up table for photolysis rates that was pre-calculated for different ozone and cloud conditions. Aerosol changes have not been considered in the look-up tables. However, it has been shown that photolysis rates can be strongly changed by increases in aerosols in both the stratosphere (for example after volcanic eruptions) or in the troposphere with strong pollution events. TUV-x is an inline model that calculates photolysis rates and takes into account both changes from ozone and aerosols (not clouds currently). The inclusion of this model is expected to significantly improve the representation of photolysis rates in the model, with implications for, e.g.,  OH and ozone in polluted regions, and is also important for geoengineering studies that use stratospheric aerosols.

### Is there anyone in particular you want to be part of this conversation?

@dkinnison @stacywalters @tilmes @fvitt @lizziel

### Will this change (regression test) answers?

No

### Will you be implementing this enhancement yourself?

Yes

lizziel:
Please include me in this discussion as well. I am a member of the GEOS-Chem Support Team and am currently developing [Cloud-J](https://github.com/geoschem/Cloud-J ) to interface to other models, the first one being [GEOS-Chem](https://github.com/geoschem/geos-chem).

GEOS-Chem is in the pipeline to be included in CAM (see https://github.com/ESCOMP/CAM/pull/484). A future version of GEOS-Chem will require Cloud-J so I will need to integrate that into CESM as well. The plan is that Cloud-J will eventually become part of MUSICA. Seeing how TUV-x is integrated into the model will help guide me on how to submit Cloud-J.

mattldawson:
Hi @lizziel - sure thing! I just submitted a draft PR #784 for this if you're interested to take a look.

PR #784, opened 2023-04-04
Add TUV-x to CAM

mattldawson:
This PR contributes to #757 by including TUV-x in the CAM build and introducing a TUV-x wrapper to allow CAM-Chem photolysis rates to be calculated in-line by TUV-x. @stacywalters @tilmes @lkemmons @dkinnison

Using TUV-x for photolysis rate calculations in CAM runs is disabled pending science evaluation/implementation of:
- cross section and quantum yield data sets
- aerosol optical inputs to TUV-x
- cloud optical inputs to TUV-x
- heating rate calculations

This is my first CAM PR and I have some remaining questions about the correct way to do some things (which I will add as comments on this PR), so I'm leaving it in draft mode for now.


cime_config/buildnml
```diff
@@ -211,6 +211,17 @@ def buildnml(case, caseroot, compname):
         if (os.path.isfile(file1)) and (not os.path.isfile(file2)):
             shutil.copy(file1,file2)
 
+        # Temporary copy of TUV-x data for development
+        dest_data = os.path.join(rundir, "data")
+        if os.path.exists(dest_data):
+            shutil.rmtree(dest_data)
+        shutil.copytree(os.path.join(srcroot, "libraries", "musica", "lib", \
+                                     "tuv-x", "data"), dest_data)
```
mattldawson, line 219: There are a large number of small data files for cross sections and quantum yields that are currently part of the TUV-x repo (but may eventually move to a database of some kind). These are not CAM-specific files and will be used by other models that link to TUV-x. Is this the correct way to stage them for use in a CAM run? @nusbaume @cacraigucar

cime_config/buildnml
```diff
@@ -211,6 +211,17 @@ def buildnml(case, caseroot, compname):
...
+        shutil.copy2(os.path.join(srcroot, "cime_config", "tuvx_MOZART.json"), \
+                     os.path.join(rundir, "tuvx_MOZART.json"))
+        shutil.copy2(os.path.join(srcroot, "cime_config", "tuvx_MOZART_TS1.json"), \
+                     os.path.join(rundir, "tuvx_MOZART_TS1.json"))
```
mattldawson, line 223: There are two CAM-specific configuration files (each for specific MOZART chemical mechanisms) needed to use TUV-x in CAM. Should these be moved to the CAM dataset repository? @nusbaume @cacraigucar

Externals.cfg
```diff
@@ -21,9 +21,9 @@ externals = Externals.cfg
 required = True
 
 [cmeps]
-tag = cmeps0.14.12
+branch = develop-add-linked-libs
```
mattldawson, line 24: This is pending [CMEPS PR 356](https://github.com/ESCOMP/CMEPS/pull/356)

cime_config/buildnml
lizziel: This handling for config files looks similar to what we originally did for GEOS-Chem. Ultimately we decided to keep the config files in the source code, copy them at case setup time to the case root via [cam/cime_config/case.case_setup.py](https://github.com/ESCOMP/CAM/blob/cam_development/cime_config/cam.case_setup.py), and then copy from case root to run directory upon submit (via [this code](url) in buildnml). This is good if you think your config file template will be tied to source code version and you want users to be able to custom edit it prior to run-time(from case root).

mnlevy1981:
Are there plans to bring biological effects in to this PR? @coupewx has coupled a version of WACCM4 with TUV (from @cbardeen) to POP in CESM2 to compute UV inhibition in ocean plankton groups; we would like to bring this work into CESM3 (with MOM6 instead of POP), and are trying to scope out what would be involved. If this update to CAM provides fields like UV index and phytoplankton damage per Cullen et al (see [this block of code](https://github.com/cbardeen/CAM/blob/cam_cesm2_1_rel_41-photo/src/chemistry/mozart/radxfr_cam.F90#L621-L708) from the version of CAM he is currently using), that would reduce the workload on our end. We skimmed through the changes in this PR and didn't see anything - if we missed something, could you please point us to it?

fvitt:
> Are there plans to bring biological effects in to this PR? @coupewx has coupled a version of WACCM4 with TUV (from @cbardeen) to POP in CESM2 to compute UV inhibition in ocean plankton groups; we would like to bring this work into CESM3 (with MOM6 instead of POP), and are trying to scope out what would be involved. If this update to CAM provides fields like UV index and phytoplankton damage per Cullen et al (see [this block of code](https://github.com/cbardeen/CAM/blob/cam_cesm2_1_rel_41-photo/src/chemistry/mozart/radxfr_cam.F90#L621-L708) from the version of CAM he is currently using), that would reduce the workload on our end. We skimmed through the changes in this PR and didn't see anything - if we missed something, could you please point us to it?

@dkinnison should weigh in on this

fvitt:
@mattldawson 
Is the version of TUVx used in your branch near the head of NCAR/tuv-x?
Do you have a test case for WACCM including the json file needed for Js and chemical heating?

cime_config/buildlib
```diff
@@ -113,12 +114,150 @@ def _build_cam(caseroot, libroot, bldroot):
...
+        macro_path = os.path.abspath(os.path.join(caseroot, "cmake_macros", ""))
+        args = "-DCONVERT_TO_MAKE=ON "
+        args += "-DCASEROOT={} ".format(caseroot)
+        args += "-DCOMPILER={} ".format(build.get_value("COMPILER"))
+        args += "-DOS={} ".format(build.get_value("OS"))
+        args += "-DMACH={} ".format(case.get_value("MACH"))
+        args += "-DCMAKE_C_COMPILER_WORKS=1 "
+        args += "-DCMAKE_Fortran_COMPILER_WORKS=1 "
+        args += "-DCMAKE_CXX_COMPILER_WORKS=1 "
+        cmd = "cmake {} .".format(args)
```
fvitt, line 141: @mattldawson 
I believe to get the correct compiler flags you need this line:
```
       args += "-DDEBUG={} ".format(build.get_value("DEBUG"))
```

cacraigucar:
@mattldawson - Is this ready for review?  If so, then please move it out of draft

mattldawson:
> @mattldawson - Is this ready for review?  If so, then please move it out of draft

Hi @cacraigucar - not quite yet. we're just merging in the latest commit in cam_development and testing

mattldawson:
closing so that @fvitt and @jzhan166 can resubmit from a branch on Francis's fork (for after I'm gone)

PR #788, opened 2023-04-13
Reimplement zonal_mean_mod::Invert_Matrix using LAPACK DGESV

### LINKED ISSUE #736: Replace "Invert_Matrix" subroutine in "zonal_mean_mod.F90" with LAPACK version
### Issue Type

Code Clean-up

### Issue Description

PR #677 will bring in a new "zonal_mean_mod.F90" that has its own matrix inversion subroutine (`Invert_Matrix`).  However, given that CAM already uses LAPACK, and that LAPACK has optimized routines to do matrix inversion, it would make sense to replace this `Invert_Matrix` subroutine with one from LAPACK.

### Will this change answers?

I Don't Know

Any CAM SE can do this

closes #745

### LINKED ISSUE #745: Bug in zonal mean "Invert_Matrix" subroutine
### What happened?

As reported by @patcal, line 1432 in the `Invert_Matrix` subroutine in `zonal_mean_mod.F90`, which is:

`do ii=11,Nbas`

Should instead be:

 `do ii=1,Nbas`

Otherwise some elements will be missed during the inversion.

@patcal Found the bug when expanding the spherical harmonics routines to manage quantities that aren't just purely zonal.

cam6_3_092

CISL machine (e.g. cheyenne)

Intel

_No response_

Any CAM SE can do this

### Extra info

The SEs need to decide if we will fix this bug directly, or if we'll instead try and resolve issue #736 instead.

brian-eaton:
The Invert_Matrix subroutine in module zonal_mean_mod has been reimplemented using the LAPACK subroutine DGESV.  This change causes roundoff level differences in the diagnostic zonal mean fields compared to the original version of Invert_Matrix with the fix from issue #745 applied. 

closes #736
fixes #745

### DIFF
--- a/src/utils/zonal_mean_mod.F90
+++ b/src/utils/zonal_mean_mod.F90
@@ -1318,104 +1318,31 @@ contains
       !
       ! Invert_Matrix: Given the NbasxNbas matrix, calculate and return
       !                the inverse of the matrix.
+      !
+      ! Implemented with the LAPACK DGESV routine.
+      !
       !====================================================================
-      real(r8),parameter:: TINY = 1.e-20_r8
       !
       ! Passed Variables
       !------------------
-      real(r8),intent(in ):: I_Mat   (:,:)
-      integer ,intent(in ):: Nbas
-      real(r8),intent(out):: O_InvMat(:,:)
+      real(r8), intent(inout) :: I_Mat(:,:)  ! input matrix contains P*L*U
+                                             ! decomposition on output
+      integer,  intent(in)    :: Nbas
+      real(r8), intent(out)   :: O_InvMat(:,:)
       !
       ! Local Values
       !-------------
-      real(r8),allocatable:: Mwrk(:,:),Rscl(:)
-      integer ,allocatable:: Indx(:)
-      real(r8):: Psgn,Mmax,Mval,Sval
-      integer :: ii,jj,kk,ndx,i2,ii_max, astat
+      integer, allocatable :: Indx(:)  ! pivot indices
+      integer :: astat, ii
       character(len=*), parameter :: subname = 'Invert_Matrix'
+      character(len=80) :: msg
+
+      external DGESV
 
       ! Allocate work space
       !---------------------
-      allocate(Mwrk(Nbas,Nbas), stat=astat)
-      call handle_allocate_error(astat, subname, 'Mwrk')
-      allocate(Rscl(Nbas), stat=astat)
-      call handle_allocate_error(astat, subname, 'Rscl')
       allocate(Indx(Nbas), stat=astat)
       call handle_allocate_error(astat, subname, 'Indx')
-
-      ! Copy the Input matrix so it can be decomposed
-      !-------------------------------------------------
-      Mwrk(1:Nbas,1:Nbas) = I_Mat(1:Nbas,1:Nbas)
-
-      ! Initailize Row scales
-      !----------------------
-      Psgn = 1._r8
-      do ii=1,Nbas
-        Mmax = 0._r8
-        do jj=1,Nbas
-          if(abs(Mwrk(ii,jj))>Mmax) Mmax = abs(Mwrk(ii,jj))
-        end do
-        if(Mmax==0._r8) then
-          call endrun('Invert_Matrix: Singular Matrix')
-        endif
-        Rscl(ii) = 1._r8/Mmax
-      end do
-
-      ! Decompose the matrix
-      !-----------------------
-      do jj=1,Nbas
-
-        if(jj>1) then
-          do ii=1,(jj-1)
-            Sval = Mwrk(ii,jj)
-            if(ii>1) then
-              do kk=1,(ii-1)
-                Sval = Sval - Mwrk(ii,kk)*Mwrk(kk,jj)
-              end do
-              Mwrk(ii,jj) = Sval
-            endif
-          end do
-        endif
-
-        Mmax = 0._r8
-        do ii=jj,Nbas
-          Sval = Mwrk(ii,jj)
-          if(jj>1) then
-            do kk=1,(jj-1)
-              Sval = Sval - Mwrk(ii,kk)*Mwrk(kk,jj)
-            end do
-            Mwrk(ii,jj) = Sval
-          endif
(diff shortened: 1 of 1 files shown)


peverwhee (approved):
Looks good to me!

src/utils/zonal_mean_mod.F90
```diff
@@ -1318,147 +1318,48 @@ subroutine Invert_Matrix(I_Mat,Nbas,O_InvMat)
...
       !
       ! Passed Variables
       !------------------
-      real(r8),intent(in ):: I_Mat   (:,:)
-      integer ,intent(in ):: Nbas
-      real(r8),intent(out):: O_InvMat(:,:)
+      real(r8), intent(inout) :: I_Mat(:,:)  ! input matrix contains P*L*U
+                                             ! decomposition on output
+      integer,  intent(in)    :: Nbas
+      real(r8), intent(out)   :: O_InvMat(:,:)
```
cacraigucar, line 1331: How is this array having its dimensionality set?   It must have the dimensions, since it is working.  I thought that it would need to be intent (inout) as intent(out) is undefined upon entry.  I believe that allocatable arrays are deallocated when they are passed into a subroutine as intent(out)

fvitt (approved):
Pleasantly surprised that the differences in the zonal means are only roundoff level differences

src/utils/zonal_mean_mod.F90
patcal: The dimensions can/should be set to (Nbas,Nbas)

patcal (approved):
Adding the dimensions to the passed arrays provides clarity and is a good safeguard against a compiler that may not like the (:,:) usage.

src/utils/zonal_mean_mod.F90
brian-eaton: @cacraigucar, the actual arguments in the 2 calls to Invert_Matrix are allocated to be shape (I_nbas,I_nbas).  The intent(out) attribute of the corresponding dummy argument means the compiler is free to make the elements of the array undefined on entry to the subroutine, but I believe that has no impact on the shape of the array or its allocation status.  I've been reading my Fortran reference and doing some searching on the web and I think this code is OK.  Do you have a reference that indicates that allocation status is affected by an intent(out) assumed shape dummy argument?  My head hurts reading about argument association rules:(

cacraigucar: https://stackoverflow.com/questions/72336483/intentout-and-allocatable-fortran-arrays-what-is-really-done was the reference I looked at to "confirm" my memory.  

It's first sentence is
`I read on many posts on Stack Overflow that an allocatable array is deallocated when it is passed in a subroutine where the dummy argument is intent(out).`

Since that confirmed what I thought I remembered, I didn't read any further.  But when I read the full post, I see the following response:

`
You are slightly misunderstanding what happens here. `

`Arrays which are deallocated on entry to a procedure with intent(out) are those allocatable arrays which correspond to an allocatable dummy.`

`What intent(out) means for an argument depends entirely on the characteristics of the dummy not actual argument.`

`Allocatable actual arguments which correspond to ordinary dummy arguments are not deallocated. (If they were, the dummy argument which is not allocatable would have to be not allocated!)`

`Instead, the allocation status of the actual argument remains unchanged and the shape of the assumed (not deferred) shape dummy argument remains the same.`

So between my bad memory and the first sentence in Stack Overflow "confirming" what I thought, I led us down a rabbit hole - sorry!

cacraigucar (approved):

src/utils/zonal_mean_mod.F90
brian-eaton: Thanks for the lesson!

gold2718:
> Adding the dimensions to the passed arrays provides clarity and is a good safeguard against a compiler that may not like the (:,:) usage.

For the record, adding the dimensions in the declaration statement disables shape checks so if the array passed into that routine does not match the declared size*, the array references will be different than in the calling routine which usually leads to unexpected results.
In my view, this makes Brian's approach safer.

*in at least one dimension that is not the last one.

patcal:
Great, I am just happy that everyone decided that I had it coded correctly
in the first place.

sjsprecious:
> > Adding the dimensions to the passed arrays provides clarity and is a good safeguard against a compiler that may not like the (:,:) usage.
> 
> For the record, adding the dimensions in the declaration statement disables shape checks so if the array passed into that routine does not match the declared size*, the array references will be different than in the calling routine which usually leads to unexpected results. In my view, this makes Brian's approach safer.
> 
> *in at least one dimension that is not the last one.

Thanks @gold2718 for this additional information, which I did not know before. I just wanted to add another point here: based on David Appelhans's talk at GTC23, using assumed-shape variables could easily hurt the GPU performance, compared to using the variables with explicit size.

gold2718:
> I just wanted to add another point here: based on David Appelhans's talk at GTC23, using assumed-shape variables could easily hurt the GPU performance, compared to using the variables with explicit size.

This seems like a classic tension (engineering for flexibility and testability vs. engineering for performance).
@sjsprecious, do you have any data on this? It would be great to know what the penalty is on real current GPU architectures. How big is the penalty?

sjsprecious:
According to the examples in David's talk, the performance penalty could be between 2x and 40x on NVIDIA's A100 GPU when using the assumed shape variables. You could find more details by searching his GTC23 talk titled "Best Practices for Programming GPUs using Fortran, OpenACC, and CUDA". I could send you a copy of the slides too in case it is no longer accessible for the public.

brian-eaton:
I think in the current context using the assumed shape arrays is fine because Invert_Matrix is just a wrapper for calling DGESV.  The dummy arrays in the DGESV interface are assumed size arrays which means the compiler is going to create a copy to provide contiguous storage for the array references that are passed down.

sjsprecious:
Thanks @brian-eaton. I agreed with you that the assumed shape arrays might not be an issue here since you were calling an LAPACK interface and not using any GPU. Just want to mention this potential performance issue somewhere for record but probably I leave it at the wrong place (apologize if that is the case).

In addition, you raised another good point and I want to make it clear: according to David's talk, using assumed size arrays won't hurt the GPU performance but using assumed shape arrays will.

PR #791, opened 2023-04-19
remove shr_file stuff and replace with shr_log_ calls

jedwards4b:
Replaces calls to shr_file_[gs]etLogUnit with calls to shr_log_setLogUnit.   
I tested this change in case ERS_D.f09_g17.B1850.cheyenne_intel.allactive-defaultio  and it reduces the size of the cesm log by removing the deprecated statements generated by the cam component.  It did not substantially change the contents of the cam.log files.


brian-eaton:
This change has already been implemented in cam6_3_087.  Closing.

PR #815, opened 2023-05-23
cam6_3_115: Add support for additional SE grids

jtruesdal:
Add support for running SE ne3pg3 out of the box.  Mostly used for testing.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,6 +1,5 @@
 [ccs_config]
 tag = ccs_config_cesm0.0.71
-#tag = ccs_config_cesm0.0.59
 protocol = git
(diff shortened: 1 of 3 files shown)


doc/ChangeLog
```diff
@@ -1,4 +1,80 @@
 ===============================================================
+Tag name: cam6_3_11X
+Originator(s): aherring
```
cacraigucar, line 3: @jtruesdal  - Make sure you add your name here as well.  You are the SE making the commit

cacraigucar (approved):

doc/ChangeLog
jtruesdal: Thanks I will do that for the final update to ChangeLog .

PR #820, opened 2023-05-25
cam6_3_116: CAM tuning F and fix Derecho CSLAM performance

### LINKED ISSUE #812: Increase PE layout for MT runs
`###` Issue Type

Infrastructure Update

### Issue Description

MT runs out-of-the-box error out while writing restart files.

@tilmes reported that the following PE layout worked for her and preliminary testing indicates it does solve the problem:

```
</entry>
    <entry id="NTASKS">
      <type>integer</type>
      <values>
        <value compclass="ATM">-50</value>
        <value compclass="CPL">-50</value>
        <value compclass="OCN">-50</value>
        <value compclass="WAV">1</value>
        <value compclass="GLC">1</value>
        <value compclass="ICE">-50</value>
        <value compclass="ROF">-50</value>
        <value compclass="LND">-50</value>
        <value compclass="ESP">1</value>
      </values>
      <desc>number of tasks for each component</desc>
    </entry>
    <entry id="NTASKS_PER_INST">
      <type>integer</type>
      <values>
        <value compclass="ATM">1800</value>
        <value compclass="OCN">1800</value>
        <value compclass="WAV">1</value>
        <value compclass="GLC">1</value>
        <value compclass="ICE">1800</value>
        <value compclass="ROF">1800</value>
        <value compclass="LND">1800</value>
        <value compclass="ESP">1</value>
      </values>
      <desc>Number of tasks per instance for each component. DO NOT EDIT: Set automatically by case.setup based on NTASKS, NINST and MULTI_DRIVER</desc>
    </entry>
    <entry id="NTHRDS">
      <type>integer</type>
      <values>
        <value compclass="ATM">1</value>
        <value compclass="CPL">1</value>
        <value compclass="OCN">1</value>
        <value compclass="WAV">1</value>
        <value compclass="GLC">1</value>
        <value compclass="ICE">1</value>
        <value compclass="ROF">1</value>
        <value compclass="LND">1</value>
        <value compclass="ESP">1</value>
      </values>
      <desc>number of threads for each task in each component</desc>
    </entry>


```
### Will this change answers?

No

Any CAM SE can do this

### LINKED ISSUE #818: Update tunings for LT and MT based on the "Tuning F"

See discussion in https://github.com/NCAR/amwg_dev/discussions/275#discussioncomment-5994091

This is the version of the tuning we would like to put on the trunk for LT and MT

Tuning (this is the so-called Tuning F):

clubb_l_predict_upwp_vpwp=.true. 
clubb_l_mono_flux_lim_um   = .true.  
clubb_l_mono_flux_lim_vm   = .true.  
clubb_c_uu_shr = 0.1
clubb_c7=0.1

micro_mg_dcs=0.00025
micro_mg_vtrmi_factor=0.5
microp_aero_wsubi_scale=2.5

dust_emis_fact         = 0.80D0
seasalt_emis_scale=1.5

@cecilehannay 

No

Yes

### LINKED ISSUE #829: New gw tuning to add to the trunk.

We are ready to add the following setting to LT and MT compsets. 
```
use_gw_convect_dp = .true.
use_gw_front = .true.
tau_0_ubc = .true.
```

Yes

No

### LINKED ISSUE #838: More tuning for LT and MT

Follow up on a 6/13 discussion with @cacraigucar.

The latest tuning for LT and MT configuration is:
- **for both LT and MT**
```
clubb_l_predict_upwp_vpwp=.true. 
clubb_l_mono_flux_lim_um   = .true.  
clubb_l_mono_flux_lim_vm   = .true.  
clubb_c_uu_shr = 0.1
clubb_c7=0.1

dust_emis_fact         = 0.80D0

micro_mg_dcs=0.00025
micro_mg_vtrmi_factor=0.5
microp_aero_wsubi_scale=2.5
seasalt_emis_scale=1.5

use_gw_convect_dp = .true.
use_gw_front = .true.
tau_0_ubc = .true.
effgw_rdg_beta = 0.5D0
effgw_rdg_beta_max = 0.5D0
```
- **for LT** 
```
effgw_beres_dp  = 0.30D0
```
- **for MT**
```
effgw_beres_dp = 0.7 
```

You already have some of these in your branch I believe. But I am giving the full list to make sure you have everything

_No response_

Yes

No

cacraigucar:
Introduce tuning "F" 
Increase PE count for MT compsets

closes #818 
closes #829
closes #838
closes #827 
closes #845

### DIFF
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -2384,7 +2384,7 @@ if (($chem =~ /_mam4/ or $chem =~ /_mam5/) and ($phys =~ /cam6/ or $phys =~ /cam
     }
 
     # for cam6 add h2o source from methane oxidation
-    if ($chem =~ /trop_mam/ or $chem =~ /ghg_mam/) {
+    if ($chem =~ /trop_mam/) {
         %species = (%species, 'h2o_ch4ox_ext_file' => 'H2O');
     }
     # for mechanisms without OCS use prescribed OCS oxidation rates for
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3724,6 +3724,8 @@ if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
      $do_gw_convect_dp or $do_gw_convect_sh )) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
+} elsif ($phys =~ /cam_dev/) {
+    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
 } elsif (!$simple_phys) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.false.');
 }
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3072,10 +3072,10 @@ if ($cfg->get('microphys') =~ /^mg/) {
             $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
         }
         elsif ($phys eq 'cam6') {
-            $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
+            $micro_mg_dcs = '500.D-6';   # default for cam6
         }
         elsif ($phys eq 'cam_dev') {
-            $micro_mg_dcs = '250.D-6';   # default for cam_dev/mg2
+            $micro_mg_dcs = '250.D-6';   # default for cam_dev
         }
     }
 
@@ -3722,8 +3722,7 @@ if (($do_gw_rdg_beta or $do_gw_rdg_gamma)) {
 #    use tau_0_ubc = .false. to avoid changing answers.
 if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
-     $do_gw_convect_dp or $do_gw_convect_sh or
-     $do_gw_oro or $do_gw_rdg_beta or $do_gw_rdg_gamma)) {
+     $do_gw_convect_dp or $do_gw_convect_sh )) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
 } elsif (!$simple_phys) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.false.');
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -498,8 +498,6 @@ my $dtime = $nl->get_value('dtime');
 #-----------------------------------------------------------------------------------------------
 # Add defaults for the CAM component
 
-my $model_top = $cfg->get('model_top');
-
 my $chem = $cfg->get('chem');
 my $waccm_chem = ($chem =~ "waccm_");
 my $waccm_phys = $cfg->get('waccm_phys');
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3700,8 +3700,6 @@ if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
      $do_gw_convect_dp or $do_gw_convect_sh)) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
-} elsif ($phys =~ /cam_dev/) {
-    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
 } elsif (!$simple_phys) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.false.');
 }
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3064,8 +3064,14 @@ if ($cfg->get('microphys') =~ /^mg/) {
 
     my $micro_mg_dcs = '400.D-6';        # default for cam5/mg1
 
-    if ($microphys eq 'mg2') {
-        if ($phys eq 'cam6') {
+    if ($microphys =~ /mg2|mg3/) {
+        if ($silhs eq '1') {
+            $micro_mg_dcs = '390.D-6';       # default for SIHLS
+        }
+        elsif ($hgrid =~ /1.9x2.5/ and $phys eq 'cam6') {
+            $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
+        }
+        elsif ($phys eq 'cam6') {
             $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
         }
         elsif ($phys eq 'cam_dev') {
@@ -3073,16 +3079,6 @@ if ($cfg->get('microphys') =~ /^mg/) {
         }
     }
 
-    if ($microphys eq 'mg3') {
-        $micro_mg_dcs = '250.D-6';       # default for cam_dev/mg3
-    }
-
-    if ($microphys =~ /mg2|mg3/ and $silhs eq '1') {
-        $micro_mg_dcs = '390.D-6';       # default for SIHLS
-    }
-    elsif ($microphys =~ /mg2|mg3/ and $hgrid =~ /1.9x2.5/) {
-        $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
-    }
 
     add_default($nl, 'micro_mg_dcs', 'val'=>$micro_mg_dcs);
 }
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3564,6 +3564,11 @@ if ($waccm_phys or
     }
     add_default($nl, 'gw_qbo_hdepth_scaling', 'val'=>$hdepth_scaling);
     add_default($nl, 'gw_top_taper');
+} elsif ($phys =~ /cam_dev/) {
+    # cam_dev settings for nlev<60 (Other cam_dev set above)
+    add_default($nl, 'use_gw_front'     , 'val'=>'.true.');
+    add_default($nl, 'use_gw_convect_dp', 'val'=>'.true.');
+    add_default($nl, 'gw_qbo_hdepth_scaling', 'val'=>'1.0D0');
 } else {
     add_default($nl, 'use_gw_front'     , 'val'=>'.false.');
     add_default($nl, 'use_gw_convect_dp', 'val'=>'.false.');
@@ -3694,6 +3699,8 @@ if (($do_gw_rdg_beta or $do_gw_rdg_gamma)) {
 if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
      $do_gw_convect_dp or $do_gw_convect_sh)) {
+    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
+} elsif ($phys =~ /cam_dev/) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
 } elsif (!$simple_phys) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.false.');
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3082,7 +3082,7 @@ if ($cfg->get('microphys') =~ /^mg/) {
     }
     elsif ($microphys =~ /mg2|mg3/ and $hgrid =~ /1.9x2.5/) {
         $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
-    }    
+    }
 
     add_default($nl, 'micro_mg_dcs', 'val'=>$micro_mg_dcs);
 }
@@ -3669,8 +3669,8 @@ if ($do_gw_rdg_beta) {
         die "$ProgName - ERROR: beta ridge scheme requires data from a topo file.\n";
     }
     add_default($nl, 'n_rdg_beta',          'val'=>'10');
-    add_default($nl, 'effgw_rdg_beta',      'val'=>'1.0D0');
-    add_default($nl, 'effgw_rdg_beta_max',  'val'=>'1.0D0');
+    add_default($nl, 'effgw_rdg_beta');
+    add_default($nl, 'effgw_rdg_beta_max');
     add_default($nl, 'trpd_leewv_rdg_beta', 'val'=>'.false.');
     add_default($nl, 'rdg_beta_cd_llb',     'val'=>'1.0D0');
 }
@@ -3726,7 +3726,8 @@ if (($do_gw_rdg_beta or $do_gw_rdg_gamma)) {
 #    use tau_0_ubc = .false. to avoid changing answers.
 if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
-     $do_gw_convect_dp or $do_gw_convect_sh)) {
+     $do_gw_convect_dp or $do_gw_convect_sh or
+     $do_gw_oro or $do_gw_rdg_beta or $do_gw_rdg_gamma)) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
 } elsif (!$simple_phys) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.false.');
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -498,6 +498,8 @@ my $dtime = $nl->get_value('dtime');
 #-----------------------------------------------------------------------------------------------
 # Add defaults for the CAM component
 
+my $model_top = $cfg->get('model_top');
+
 my $chem = $cfg->get('chem');
 my $waccm_chem = ($chem =~ "waccm_");
 my $waccm_phys = $cfg->get('waccm_phys');
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -3029,7 +3029,6 @@ if ($cfg->get('microphys') =~ /^mg/) {
     add_default($nl, 'nucleate_ice_subgrid_strat');
     add_default($nl, 'nucleate_ice_use_troplev');
     add_default($nl, 'cld_macmic_num_steps', 'dtime'=>$dtime);
-    add_default($nl, 'micro_mg_dcs');
     add_default($nl, 'micro_mg_precip_frac_method');
     add_default($nl, 'micro_mg_berg_eff_factor');
     add_default($nl, 'nucleate_ice_incloud');
@@ -3057,6 +3056,35 @@ if ($cfg->get('microphys') =~ /^mg/) {
 	add_default($nl, 'micro_mg_accre_sees_auto');
 	add_default($nl, 'micro_mg_implicit_fall');
     }
+
+    # tunings for micro_mg_dcs
+    my $microphys = $cfg->get('microphys');
+    my $hgrid = $cfg->get('hgrid');
+    my $silhs = $cfg->get('silhs');
+
+    my $micro_mg_dcs = '400.D-6';        # default for cam5/mg1
+
+    if ($microphys eq 'mg2') {
+        if ($phys eq 'cam6') {
+            $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
+        }
+        elsif ($phys eq 'cam_dev') {
+            $micro_mg_dcs = '250.D-6';   # default for cam_dev/mg2
+        }
+    }
+
+    if ($microphys eq 'mg3') {
+        $micro_mg_dcs = '250.D-6';       # default for cam_dev/mg3
+    }
+
+    if ($microphys =~ /mg2|mg3/ and $silhs eq '1') {
+        $micro_mg_dcs = '390.D-6';       # default for SIHLS
+    }
+    elsif ($microphys =~ /mg2|mg3/ and $hgrid =~ /1.9x2.5/) {
+        $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
+    }    
+
+    add_default($nl, 'micro_mg_dcs', 'val'=>$micro_mg_dcs);
 }
 
 # Aerosol Namelist options
--- a/bld/config_files/definition.xml
+++ b/bld/config_files/definition.xml
@@ -47,7 +47,9 @@ Switch to turn on analytic initial conditions for the dynamics state:
   1 => yes.
 </entry>
 <entry id="model_top" valid_values="none,lt,mt" value="none">
-Model top specifier
+Model top specifier - set by compset definition
+  lt: ~ 40 km top
+  mt: ~ 80 km top
 </entry>
 <entry id="waccmx" valid_values="0,1" value="0">
 Option to turn on waccmx thermosphere/ionosphere extension: 0 => no, 1 => yes
--- a/bld/config_files/definition.xml
+++ b/bld/config_files/definition.xml
@@ -46,6 +46,9 @@ Switch to turn on analytic initial conditions for the dynamics state:
   0 => no
   1 => yes.
 </entry>
+<entry id="model_top" valid_values="none,lt,mt" value="none">
+Model top specifier
+</entry>
 <entry id="waccmx" valid_values="0,1" value="0">
 Option to turn on waccmx thermosphere/ionosphere extension: 0 => no, 1 => yes
 </entry>
--- a/bld/namelist_files/namelist_defaults_cam.xml
+++ b/bld/namelist_files/namelist_defaults_cam.xml
@@ -741,7 +741,9 @@
 <gw_limit_tau_without_eff chem="trop_strat_mam5_vbs">.true. </gw_limit_tau_without_eff>
 <gw_limit_tau_without_eff chem="trop_strat_mam5_vbsext">.true. </gw_limit_tau_without_eff>
 <gw_apply_tndmax                        >.true. </gw_apply_tndmax>
-<gw_apply_tndmax          waccm_phys="1">.false.</gw_apply_tndmax>
+<gw_apply_tndmax chem="ghg_mam4" nlev="93">.false.</gw_apply_tndmax>
+<gw_apply_tndmax chem="ghg_mam4" nlev="58">.false.</gw_apply_tndmax>
+<gw_apply_tndmax            waccm_phys="1">.false.</gw_apply_tndmax>
 <gw_apply_tndmax chem="trop_strat_mam4_vbs">.false.</gw_apply_tndmax>
 <gw_apply_tndmax chem="trop_strat_mam4_vbsext">.false.</gw_apply_tndmax>
 <gw_apply_tndmax chem="trop_strat_mam5_vbs">.false.</gw_apply_tndmax>
@@ -2977,6 +2979,7 @@
 <se_hypervis_subcycle                                            > 3 </se_hypervis_subcycle>
 <se_hypervis_subcycle  waccm_phys="1"                            > 2 </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne16np4"                           > 4 </se_hypervis_subcycle>
+<se_hypervis_subcycle  hgrid="ne30np4" npg="3" nlev="58"         > 4 </se_hypervis_subcycle>
 <se_hypervis_subcycle  se_refined_mesh="1"                       > 3 </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4CONUS.ne30x8" waccm_phys="1" > 1 </se_hypervis_subcycle>
 
--- a/bld/namelist_files/namelist_defaults_cam.xml
+++ b/bld/namelist_files/namelist_defaults_cam.xml
@@ -700,6 +700,8 @@
 <!-- gravity wave settings -->
 <effgw_beres_dp>0.7D0</effgw_beres_dp>
 <effgw_beres_dp hgrid="0.9x1.25"                              >0.4D0</effgw_beres_dp>
+<effgw_beres_dp model_top="lt"                                >0.30D0</effgw_beres_dp>
+<effgw_beres_dp model_top="mt"                                >0.70D0</effgw_beres_dp>
 <effgw_beres_dp waccm_phys="0"                                >0.55D0</effgw_beres_dp>
 <effgw_beres_dp waccm_phys="1" hgrid="0.9x1.25"               >0.5D0</effgw_beres_dp>
 <effgw_beres_dp waccm_phys="1" hgrid="ne30np4" nlev="70"      >0.5D0</effgw_beres_dp>
@@ -708,7 +710,16 @@
 <effgw_beres_dp chem="trop_strat_mam4_vbsext" hgrid="0.9x1.25">0.5D0</effgw_beres_dp>
 <effgw_beres_dp chem="trop_strat_mam5_vbs" hgrid="0.9x1.25"   >0.5D0</effgw_beres_dp>
 <effgw_beres_dp chem="trop_strat_mam5_vbsext" hgrid="0.9x1.25">0.5D0</effgw_beres_dp>
+
 <effgw_oro      nlev="60"     >0.0625D0</effgw_oro>
+
+<effgw_rdg_beta                    >1.0D0</effgw_rdg_beta>
+<effgw_rdg_beta model_top="lt"     >0.5D0</effgw_rdg_beta>
+<effgw_rdg_beta model_top="mt"     >0.5D0</effgw_rdg_beta>
+
+<effgw_rdg_beta_max                >1.0D0</effgw_rdg_beta_max>
+<effgw_rdg_beta_max model_top="lt" >0.5D0</effgw_rdg_beta_max>
+<effgw_rdg_beta_max model_top="mt" >0.5D0</effgw_rdg_beta_max>
 
 <!-- setting for gravity waves from shallow convection. -->
 <effgw_beres_sh>0.03D0</effgw_beres_sh>
@@ -2976,7 +2987,7 @@
 <se_hypervis_subcycle                                                           > 3 </se_hypervis_subcycle>
 <se_hypervis_subcycle  waccm_phys="1"                                           > 2 </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne16np4"                                          > 4 </se_hypervis_subcycle>
-<se_hypervis_subcycle  phys="cam_dev" hgrid="ne30np4" npg="3" model_top="lt"    > 4 </se_hypervis_subcycle>
+<se_hypervis_subcycle  hgrid="ne30np4"  model_top="lt"                          > 4 </se_hypervis_subcycle>
 <se_hypervis_subcycle  se_refined_mesh="1"                                      > 3 </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4CONUS.ne30x8" waccm_phys="1"                > 1 </se_hypervis_subcycle>
 
@@ -3007,7 +3018,7 @@
 
 <se_nu_top                                   > 1.25e5 </se_nu_top>
 <se_nu_top waccmx="1"                        > 1.0e6 </se_nu_top>
-<se_nu_top phys="cam_dev" model_top="mt"     > 1.0e6 </se_nu_top>
+<se_nu_top model_top="mt"                    > 1.0e6 </se_nu_top>
 
 <se_molecular_diff               >  0.0 </se_molecular_diff>
 <se_molecular_diff waccmx="1"    >  1.0 </se_molecular_diff>
@@ -3038,7 +3049,7 @@
 
 <se_rsplit                                          > 3 </se_rsplit>
 <se_rsplit waccm_phys="1"                           > 4 </se_rsplit>
-<se_rsplit phys="cam_dev" model_top="mt"            > 6 </se_rsplit>
+<se_rsplit model_top="mt"                           > 6 </se_rsplit>
 <se_rsplit hgrid="ne0np4CONUS.ne30x8" waccm_phys="1"> 3 </se_rsplit>
 
 <se_fvm_supercycling>-1</se_fvm_supercycling>
--- a/bld/namelist_files/namelist_defaults_cam.xml
+++ b/bld/namelist_files/namelist_defaults_cam.xml
@@ -125,7 +125,7 @@
 <ncdata dyn="se" hgrid="ne30np4" nlev="93" chem="trop_strat_mam5_vbs">atm/cam/inic/se/FCMTHIST_ne30pg3_1980-01-01_c221214.nc</ncdata>
 <ncdata dyn="se" hgrid="10x15"   nlev="32" chem="trop_strat_mam5_vbs">atm/cam/inic/fv/FC2000mam5_f10_0002-01-01_c221214.nc</ncdata>
 
-<ncdata dyn="se" hgrid="ne30np4" npg="3" nlev="58">atm/cam/inic/se/FWsc_ne30pg3_58L_GRID_48_taperstart10km_lowtop_BL10_v3_beta1p75_Top_43km.nc</ncdata>
+<ncdata dyn="se" hgrid="ne30np4" npg="3" nlev="58">atm/cam/inic/se/FLT_L58_ne30pg3_IC_c220623.nc</ncdata>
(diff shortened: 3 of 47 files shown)


bld/config_files/definition.xml
```diff
@@ -46,6 +46,9 @@ Switch to turn on analytic initial conditions for the dynamics state:
   0 => no
   1 => yes.
 </entry>
+<entry id="model_top" valid_values="none,lt,mt" value="none">
+Model top specifier
```
fvitt, line 50:
```suggestion
Model top specifier
  lt : ~ 40 km top
  mt : ~ 80 km top
```

cime_config/config_pes.xml
```diff
@@ -146,6 +146,43 @@
...
+      <pes pesize="any" compset="CAM%DEV%MT">
+        <comment>none</comment>
+        <ntasks>
+          <ntasks_atm>-50</ntasks_atm>
+          <ntasks_lnd>-50</ntasks_lnd>
+          <ntasks_rof>-50</ntasks_rof>
+          <ntasks_ice>-50</ntasks_ice>
+          <ntasks_ocn>-50</ntasks_ocn>
+          <ntasks_glc>-50</ntasks_glc>
+          <ntasks_wav>-50</ntasks_wav>
```
fvitt, line 160: I believe the glacier and wave models are stubs.   Can these be removed and use the default `ntasks`, `nthrds`, and `rootpe` for these?

cime_config/config_component.xml
```diff
@@ -166,8 +166,8 @@
       <value compset="_CAM\d0%CC.*%SDYN">-nlev 56</value>
       <value compset="_CAM\d0%WC.*%SDYN">-nlev 88</value>
       <value compset="_CAM\d0%WX.*%SDYN">-nlev 145</value>
-      <value compset="_CAM.*%LT">-nlev 58</value>
-      <value compset="_CAM.*%MT">-nlev 93</value>
+      <value compset="_CAM.*%LT">-nlev 58 -model_top lt</value>
+      <value compset="_CAM.*%MT">-nlev 93 -model_top mt</value>
```
fvitt, line 170: I guess you are not setting a default number of layers for `lt` or `mt` model top.  So the`-nlev` setting is still needed.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3030,18 +3039,18 @@
...
+<se_nsplit hgrid="ne30np4"  waccm_phys="1"                >  5  </se_nsplit>
+<se_nsplit phys="cam_dev" hgrid="ne30np4"  model_top="mt" >  3  </se_nsplit>
+<se_nsplit hgrid="ne0np4CONUS.ne30x8" waccm_phys="1"      > 10  </se_nsplit>
 
 <se_nsplit hgrid="ne0np4TESTONLY.ne5x4"> 7 </se_nsplit>
 
 <se_rsplit                                          > 3 </se_rsplit>
 <se_rsplit waccm_phys="1"                           > 4 </se_rsplit>
-<se_rsplit nlev="93"                                > 6 </se_rsplit>
+<se_rsplit phys="cam_dev" model_top="mt"            > 6 </se_rsplit>
```
fvitt, line 3053: I imagine some of the setting for the SE dycore are more dependent on number of model layers rather the physics and altitude of the of the model top.  I am not going to fret over it too much...

fvitt (commented):
A few questions/comments.

bld/config_files/definition.xml
brian-eaton: It would be useful to point out that `model_top` is being set by the compset definition.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -741,8 +741,8 @@
 <gw_limit_tau_without_eff chem="trop_strat_mam5_vbs">.true. </gw_limit_tau_without_eff>
 <gw_limit_tau_without_eff chem="trop_strat_mam5_vbsext">.true. </gw_limit_tau_without_eff>
 <gw_apply_tndmax                        >.true. </gw_apply_tndmax>
-<gw_apply_tndmax chem="ghg_mam4" nlev="93">.false.</gw_apply_tndmax>
-<gw_apply_tndmax chem="ghg_mam4" nlev="58">.false.</gw_apply_tndmax>
+<gw_apply_tndmax phys="cam_dev" chem="ghg_mam4" model_top="mt">.false.</gw_apply_tndmax>
+<gw_apply_tndmax phys="cam_dev" chem="ghg_mam4" model_top="lt">.false.</gw_apply_tndmax>
```
brian-eaton, line 745: This seems like overkill on the attribute specification.  Either `phys="cam_dev" chem="ghg_mam4"` or `model_top="lt"` would suffice.  Using `phys="cam_dev" chem="ghg_mam4"` here would be my choice.

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: If Francis isn't going to fret, then neither am I.  I agree that it's hard to evaluate whether these attribute settings make sense without knowing what the requirements are.

cime_config/config_component.xml
brian-eaton: I believe `nlev` will always be needed for the initial file (unless the model infrastructure were refactored to be totally dynamic).  It's the initial file which sets the model top.

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: A question I had when first looking at the changes in `namelist_defaults_cam.xml` was whether the new settings that were being enabled by the `phys="cam_dev"` attribute were intended to be applied to the `F2000dev` compset as well.  I now see that that issue was discussed here [https://github.com/NCAR/amwg_dev/discussions/275](url) and that @adamrher was OK with it.  But the value of `gw_apply_tndmax` is `.true.` for `F2000dev` and is `.false.` for the `LT` compsets.  Since both are low top models using `cam_dev` physics I've been trying to understand what the important difference is between these compsets to help inform how to set the attributes that provide the different default values.  I'm guessing now that the difference is more related to the vertical resolution difference than it is to model top.  This may be related to @fvitt's comment above about dependence on number of layers.  Adam, what's your take?

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: The `chem="ghg_mam4"` attribute is what would keep this setting from being used with `F2000dev`.  So the resolution here depends on @adamrher  confirming that indeed the setting for `F2000dev` should be `.true.` as it is for `F2000climo`, or not...

adamrher: Apologies Brian, I missed this question to me since I get email prompts even when I'm not mentioned. I would like the gw settings in F2000dev to be identical to whatever the latest FLTHIST compset is. The only difference from FLTHIST is the ```chem="ghg_mam4"``` where I'd like to retain the chem options from F2000climo in F2000dev. Does that answer your question(s) Brian?

brian-eaton: Thanks Adam.  I think that should make it easier to just use `phys="cam_dev"` as the attribute for many of the new `LT/MT` compset settings.

nusbaume: I agree with everyone that `phys="cam_dev"` should be the first control on the namelist defaults, with the `model_top` attribute only needed when a namelist variable changes with respect to `LT` or `MT`, or some other model top setting (so in this particular case where they are the same I don't think `model_top` is needed at all).

In terms of my original request, basically the one thing I don't want to see (for now) is a `model_top` attribute that doesn't have `phys="cam_dev"` listed first.  This is because these LT and MT settings are only in the context of CAM7 physics, and by requiring the `phys` attribute it will protect us from mistakenly modifying other model configuration settings (e.g. CAM6).   

Anyways, I hope that helps, and let me know if you are still confused or concerned about this particular ask!

cacraigucar: Set it just to "cam_dev"

cime_config/config_pes.xml
cacraigucar: No longer setting the ntasks for MT, since further testing indicated that it is not a memory issue

bld/build-namelist
```diff
@@ -3695,6 +3700,8 @@ if ((not $waccm_phys) and
     ($do_gw_front or $do_gw_front_igw or
      $do_gw_convect_dp or $do_gw_convect_sh)) {
     add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
+} elsif ($phys =~ /cam_dev/) {
+    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
```
adamrher, line 3704: Do we even need this additional entry? cam_dev satisfies the first if block:

```
(not $waccm_phys)
use_gw_front=.true.
use_gw_convect_dp=.true.
```

Also, what if nothing satisfied this if block ... is tau_o_ubc just not set? Shouldn't it be initialized to false somewhere above this code block, as a fall back value?

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2044,9 +2045,12 @@
 <clubb_l_mono_flux_lim_spikefix                   > .true.  </clubb_l_mono_flux_lim_spikefix>
 <clubb_l_mono_flux_lim_thlm                       > .true.  </clubb_l_mono_flux_lim_thlm>
 <clubb_l_mono_flux_lim_um                         > .false. </clubb_l_mono_flux_lim_um>
+<clubb_l_mono_flux_lim_um phys="cam_dev"          > .true. </clubb_l_mono_flux_lim_um>
```
adamrher, line 2048: @bstephens82 the mlf limiter on um/vm wouldn't get invoked if clubb_l_predict_upwp_vpwp=F, or does it?

bstephens82: Correct. It only operates on the winds if using prognostic momentum fluxes.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2156,6 +2160,7 @@
 <micro_mg_dcs         microphys="mg3">                                 500.D-6    </micro_mg_dcs>
 <micro_mg_dcs         microphys="mg3"                   silhs="1"   >  390.D-6    </micro_mg_dcs>
 <micro_mg_dcs         microphys="mg3" hgrid="1.9x2.5">                 200.D-6    </micro_mg_dcs>
+<micro_mg_dcs         microphys="mg3" phys="cam_dev">                  250.D-6    </micro_mg_dcs>
```
adamrher, line 2163: What happens if we have mg3=T, hgrid="1.9x2.5" can cam_dev=T?

I'm inclined to recommend a larger hammer -- can we just condition this on cam_dev, and not mg3? I want it to work for cam_dev + mg2, but I'm not sure whether we should have another entry for cam_dev+mg2, or just have the large hammer to simplify the interpretation.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2195,6 +2201,7 @@
 <micro_mg_homog_size  microphys="mg2"  >  25.D-6    </micro_mg_homog_size>
 
 <micro_mg_vtrmi_factor                  >  1.D0    </micro_mg_vtrmi_factor>
+<micro_mg_vtrmi_factor phys="cam_dev"   >  0.5D0    </micro_mg_vtrmi_factor>
 <micro_mg_vtrmi_factor microphys="mg2"  >  1.D0    </micro_mg_vtrmi_factor>
```
adamrher, line 2205: same issue - what if I turn on mg2 and cam_dev? Does the last entry that satisfies the logic the one that wins out?

Actually I would just delete the mg2 entry, as its set to the default value. That would alleviate my concerns.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2976,12 +2984,12 @@
...
-<se_hypervis_subcycle                                            > 3 </se_hypervis_subcycle>
-<se_hypervis_subcycle  waccm_phys="1"                            > 2 </se_hypervis_subcycle>
-<se_hypervis_subcycle  hgrid="ne16np4"                           > 4 </se_hypervis_subcycle>
-<se_hypervis_subcycle  hgrid="ne30np4" npg="3" nlev="58"         > 4 </se_hypervis_subcycle>
-<se_hypervis_subcycle  se_refined_mesh="1"                       > 3 </se_hypervis_subcycle>
-<se_hypervis_subcycle  hgrid="ne0np4CONUS.ne30x8" waccm_phys="1" > 1 </se_hypervis_subcycle>
+<se_hypervis_subcycle                                                           > 3 </se_hypervis_subcycle>
+<se_hypervis_subcycle  waccm_phys="1"                                           > 2 </se_hypervis_subcycle>
+<se_hypervis_subcycle  hgrid="ne16np4"                                          > 4 </se_hypervis_subcycle>
+<se_hypervis_subcycle  phys="cam_dev" hgrid="ne30np4" npg="3" model_top="lt"    > 4 </se_hypervis_subcycle>
```
adamrher, line 2990: I don't see why this needs to be conditioned on npg="3". I'd prefer it not. We may need to wait for @PeterHjortLauritzen to get his take.

adamrher (changes requested):
Hi I've provided some comments asking for clarification. 

I accidentally added some comments as one-off comments, and not part of the review. oops.

bld/namelist_files/namelist_defaults_cam.xml
adamrher: In that case, I would suggest that the default for these (```clubb_l_mono_flux_lim_um``` and ```clubb_l_mono_flux_lim_um```) always be set to true. request to get rid of the cam_dev qualifier. If anything it should be qualified on ```clubb_l_predict_upwp_vpwp```, but as Ben said, it's not really necessary since its not used when predict_upwp=F.

bld/build-namelist
cacraigucar: @adamrher - I wasn't sure that cam_dev will always satisfy the "and" part of the statement.  Since I was told it should be set for all cam_dev runs, that is what I was doing, even if the other settings were to be changed.  It sounds like the finer granularity of the above if statement is sufficient, so I should remove this section?

tau_0_ubc is initialized to false in gw_drag.F90 and keeps that setting unless it is overriden with a namelist setting.  If none of these settings are executed, then the atm_in file will not have the variable appear in it.  

Removing the "large hammer" cam_dev check

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: Always set to true now

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: I'm happy with using the large `cam_dev` hammer since it sounds like the values will be the same.  We can fine tune it with additional qualifiers if needed in the future.

In answer to the question, I believe with the namelist_defaults_cam.xml file, the last value which matches is what is ultimately set.

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: Deleted unnecessary line qualified by mg2

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: That question I can not answer, as it was already being selected with `npg=3`.  Do you want to hold up this PR until @PeterHjortLauritzen returns from PTO or should we just open an issue to investigate it?

bld/build-namelist
adamrher: I think that the first logic block is sufficient; if all of those gw namelists are false, meaning only orographic **could** be on, I don't think tau_0_ubc is consequential. Let me ask Julio tmrw, and resolve this convo then.

bld/namelist_files/namelist_defaults_cam.xml
adamrher: After thinking a little more on this; I'm going to advocate for the qualifiers, but that we should (a) also include one for mg2 and cam_dev, and (b) move both just above the mgX and hgrid=2deg entries; I'd like to retain the hgrid=2deg tunings even if we do turn on cam_dev. When people run 2deg I'd like them to have the 2deg tunings. If it needs to be retuned with cam_dev, then we can make that change then.

bld/namelist_files/namelist_defaults_cam.xml
adamrher: I'm going to make some executive decisions here. Dycore time-stepping should not have either (a) cam_dev qualifiers, or (b) npg qualifiers. These are both physics, and are orthogonal to the dynamics. Time-stepping should only be a function of the hgrid, the number of levels and the model top location. I suspect the waccm physics qualifier is a proxy for changes in vertical levels and/or model top.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3030,18 +3036,18 @@
...
-<se_nsplit hgrid="ne16np4"  waccm_phys="1" waccmx="1" >  5  </se_nsplit>
-<se_nsplit hgrid="ne16np4"  waccm_phys="1"            >  3  </se_nsplit>
-<se_nsplit hgrid="ne30np4"  waccm_phys="1"            >  5  </se_nsplit>
-<se_nsplit hgrid="ne30np4"  nlev="93"                 >  3  </se_nsplit>
-<se_nsplit hgrid="ne0np4CONUS.ne30x8" waccm_phys="1"  > 10  </se_nsplit>
+<se_nsplit hgrid="ne5np4"   waccm_phys="1"                >  3  </se_nsplit>
+<se_nsplit hgrid="ne16np4"  waccm_phys="1" waccmx="1"     >  5  </se_nsplit>
+<se_nsplit hgrid="ne16np4"  waccm_phys="1"                >  3  </se_nsplit>
+<se_nsplit hgrid="ne30np4"  waccm_phys="1"                >  5  </se_nsplit>
+<se_nsplit phys="cam_dev" hgrid="ne30np4"  model_top="mt" >  3  </se_nsplit>
```
adamrher, line 3043: remove cam_dev qualifier.

bld/namelist_files/namelist_defaults_cam.xml
adamrher: unresolving this conversation because I don't think physics should qualify the dycore time-stepping. If someone wants to try to run -phys cam6 w/ mt, then have at it. Please remove the cam_dev qualifier.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3008,9 +3014,9 @@
 
 <se_nu_p>-1 </se_nu_p>
 
-<se_nu_top                    > 1.25e5 </se_nu_top>
-<se_nu_top waccmx="1"         > 1.0e6 </se_nu_top>
-<se_nu_top nlev="93"          > 1.0e6 </se_nu_top>
+<se_nu_top                                   > 1.25e5 </se_nu_top>
+<se_nu_top waccmx="1"                        > 1.0e6 </se_nu_top>
+<se_nu_top phys="cam_dev" model_top="mt"     > 1.0e6 </se_nu_top>
```
adamrher, line 3019: remove cam_dev qualifier; if someone wants to run mt w/ a different physics they should have this value of nu_top.

adamrher (changes requested):
My requests here reflect that the dycore time-stepping should not be conditioned on physics. If someone wants to run something other than cam_dev, than they should still get stable time-stepping.

bld/namelist_files/namelist_defaults_cam.xml
adamrher: I've not much to add since Brian is in agreement with me here. And I won't step into the debate on use cases vs namelist_defaults because I admittedly have very little familiarity with use cases and their pros / cons.

> There's no obvious correlation here between model top and the setting of se_nsplit.

As I've been dragged into understanding high top modeling, I've learned that b/c the mt lid is right above where there's a lot of wave breaking, that this is a particularly bad combination as waves are reflecting off the lid and the sponge layer apparently can't completely rid this dynamic. Or at least thats my cartoon understanding of it. So I think the non-monotonic relationship of model top and cfl limiting time-step is to be expected.

So I will add that the ```waccm_pyhs="1"``` modifier is very opaque since its keying off the vertical grid / model top aspect that one should not associate with physics. I figure we let this be grandfathered in at this point, but that we not introduce anymore of this type of logic going forward.

bld/build-namelist
adamrher: I've checked with Julio, and have come up with a better solution. Yes, lets keep that cam_dev hammer out of the logic, and instead expand the first if block to include the three orographic options:

```
    ($do_gw_front or $do_gw_front_igw or
     $do_gw_convect_dp or $do_gw_convect_sh or
     $do_gw_oro or $do_gw_rdg_beta or $do_gw_rdg_gamma)) {
    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
```

Since tau_0_ubc controls the b.c.'s for all the various gw schemes, this ensures that we have tau_0_ubc=T if any of the gw schemes are on (and not running waccm_phys).

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: I'll add for the record here that when the namelist defaults have an equal number of attribute matches, it's the first one that gets returned.  But it's not robust to depend on this behavior (mainly because maintenance is difficult).  The attributes should always be chosen to avoid a tie.  When the logic gets complicated, it's best to move it to `build-namelist` where the availability of programming logic makes it easier to ensure the desired defaults.  That might be the case here since it looks like `silhs` may already getting its default using the assumption that an attribute count tie results in the first element being chosen.

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: @nusbaume - Since you specifically asked me to use `phys="cam_dev"` with all model_top settings, I want to make sure you are okay with Adam's rationale.

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: @nusbaume - Another explanation on why I am being advised to remove the `phys="cam_dev"` qualifier from the lines with `-model_top`

bld/namelist_files/namelist_defaults_cam.xml
adamrher: I'd advocate for moving this particular namelist setting to build-namelist, since it's getting so complicated.

bld/namelist_files/namelist_defaults_cam.xml
nusbaume: I am happy to drop the `phys` qualifer.  I mostly just wanted it to prevent any unexpected namelist behavior, but since it sounds like including it could itself produce (at least scientifically) unexpected behavior then we can go ahead and drop it.

bld/namelist_files/namelist_defaults_cam.xml
nusbaume: I am fine to remove it here too, as I agree that in general the physics shouldn't control the dycore settings.

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: @adamrher, here is proposed code for `build-namelist` to implement the `micro_mg_dcs` defaults.  Please have a look and let me know if this provides the defaults you have in mind.

    # tunings for micro_mg_dcs
    my $microphys = $cfg->get('microphys');
    my $hgrid = $cfg->get('hgrid');
    my $silhs = $cfg->get('silhs');

    my $micro_mg_dcs = '400.D-6';        # default for cam5/mg1

    if ($microphys eq 'mg2') {
        if ($phys eq 'cam6') {
            $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
        }
        elsif ($phys eq 'cam_dev') {
            $micro_mg_dcs = '250.D-6';   # default for cam_dev/mg2
        }
    }

    if ($microphys eq 'mg3') {
        $micro_mg_dcs = '250.D-6';       # default for cam_dev/mg3
    }

    if ($microphys =~ /mg2|mg3/ and $silhs eq '1') {
        $micro_mg_dcs = '390.D-6';       # default for SIHLS
    }
    elsif ($microphys =~ /mg2|mg3/ and $hgrid =~ /1.9x2.5/) {
        $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
    }    

    add_default($nl, 'micro_mg_dcs', 'val'=>$micro_mg_dcs);

bld/namelist_files/namelist_defaults_cam.xml
PeterHjortLauritzen: agreed

bld/namelist_files/namelist_defaults_cam.xml
adamrher: The build-namelist mods look good to me, Brian.

I will chat with Julio tmrw about the 2deg tunings, to get his take, and report back here.

fvitt (approved):

bld/namelist_files/namelist_defaults_cam.xml
adamrher: I talked w/ Julio --- lets discuss this on Tuesday w/ the bigger group. Assuming you weren't planning on completing this PR before Tues?

cacraigucar: I'd rather get the PR "correct" than to get it in quickly.  I've added an item to Tuesday's agenda for discussion

bld/build-namelist
cacraigucar: This change is introducing answer changes for tests like FHIST.  Is that okay? @adamrher and @JulioTBacmeister what are your thoughts ?

bld/build-namelist
```diff
@@ -3057,6 +3056,35 @@ if ($cfg->get('microphys') =~ /^mg/) {
...
+        if ($phys eq 'cam6') {
+            $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
+        }
+        elsif ($phys eq 'cam_dev') {
+            $micro_mg_dcs = '250.D-6';   # default for cam_dev/mg2
+        }
+    }
+
+    if ($microphys eq 'mg3') {
+        $micro_mg_dcs = '250.D-6';       # default for cam_dev/mg3
```
cacraigucar, line 3077: This change is also changing answers for compsets like F2000climo (which uses mg3).  Is this wanted?  @adamrher @JulioTBacmeister

bld/build-namelist
adamrher: hmmm ... we'll need to discuss this on Tues as well. It depends on what we want the non-FLT/FMT compsets to "mean" in CESM3.

I suspect we may want to go back to adding an explicit cam_dev if statement, but lets talk on Tues.

bld/build-namelist
adamrher: I think @brian-eaton may have made a typo? The default for mg3 in namelist_defaults was:

```
<micro_mg_dcs microphys="mg3"> 500.D-6 </micro_mg_dcs>
```

so I think this needs to be changed to 500E-6.

adamrher: actually I don't think its a typo, I think an explicit cam_dev if statement needs to be added to the mg3 block, that sets it to 250E-6. Otherwise use the default 500E-6.

brian-eaton: @cacraigucar, `F2000climo` uses `cam6` which implies `mg2`.  I don't think this build-namelist code should be a change for `F2000climo` which should still be getting `micro_mg_dcs=500E-6`.  That said, I was looking at your testing on cheyenne and see that there's an `F2000climo` compset run with `C96` which uses testmods from `cam/outfrq9s_mg3`.  That test using a `cam6/mg3` combo, which is not a configuration from a standard compset, should see diffs.
@adamrher, since `cam_dev` implies `mg3` (at least that's what's in configure), I intentionally set the default for `mg3` to 250e-6 which I thought was the value we wanted for the FLT* and FLM* compsets.  So I'm still thinking the way I set these defaults up is OK.  What do you think?

adamrher: right, ok, so the issue is that for compsets using cam6 physics, you shouldn't be running tests using ```cam/outfrq9s_mg3```, since cam6 by definition in configure is mg2

```
elsif ($phys_pkg eq 'cam6') {
    $microphys_pkg = 'mg2';
}
```
I think that we need to ensure that tests using cam6 compsets (F2000climo,FHIST,etc...) are not xml'ing in mg3. I can't think of why this was done in the first place, but it's possible I'm missing something.

adamrher: But to answer your broader question whether we can equate cam_dev with mg3, and therefore set the default mg3 values to represent cam_dev settings , I actually would like to revisit this at our tuesday meeting ... mg3 adds two additional dynamics tracers and I'm not sure we should be adding this additional computational cost to the already very expensive cam_dev/cam7, as it is intended to run with the increased levels L32->L58. Andrew keeps telling me the mg3 enhancements don't make a difference at 1˚.

brian-eaton: In the past we've often set up tests using phys/chem combos or non-standard physics and/or chem configurations that a researcher may well like to try (e.g., `cam6/mg3`).  I don't think that's a bad idea.  In the case of running `F2000climo/mg3` I think the issue is just that we always need to justify "baseline comparison" failures in our regression tests.  This particular failure would be justified by saying it was due to changing an mg3 default namelist setting, if that's what we decide to do.  The issue the test raised turns out to be a good one:)

bld/build-namelist
adamrher: Following our convo at todays mtg, we decided we should preserve the L32 cam6 compsets (which will need to be renamed, and cheryl is planing to open a git issue to get that convo started) and so lets revert this logic back to how we first had it with the explicit cam_dev logic:

```
if ((not $waccm_phys) and
    ($do_gw_front or $do_gw_front_igw or
     $do_gw_convect_dp or $do_gw_convect_sh)) {
    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
} elsif ($phys =~ /cam_dev/) {
    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
```

This will maintain the value of false for cam6 compsets, which only uses orographic gw settings.

gold2718: Out of curiousity, what is going to happen with 20 or so `science_support` tags? Are we still claiming they give answers compatible with publications?

bld/namelist_files/namelist_defaults_cam.xml
adamrher: Following today's convo, I'd like to divorce cam_dev from mg version type. I'd like to make the 2deg tunings to be explicit they are for cam6 physics. So I propose:

```
# tunings for micro_mg_dcs
my $microphys = $cfg->get('microphys');
my $hgrid = $cfg->get('hgrid');
my $silhs = $cfg->get('silhs');

my $micro_mg_dcs = '400.D-6';        # default for cam5/mg1

if ($microphys =~ /mg2|mg3/) {
    if ($phys eq 'cam6') {
        $micro_mg_dcs = '500.D-6';   # default for cam6
    }
    elsif ($phys eq 'cam_dev') {
        $micro_mg_dcs = '250.D-6';   # default for cam_dev
    }
}

if ($microphys =~ /mg2|mg3/ and $silhs eq '1') {
    $micro_mg_dcs = '390.D-6';       # default for SIHLS
}
elsif ($microphys =~ /mg2|mg3/ and $hgrid =~ /1.9x2.5/ and $phys eq 'cam6') {
    $micro_mg_dcs = '200.D-6';       # default for FV 2-deg/cam6
}    

add_default($nl, 'micro_mg_dcs', 'val'=>$micro_mg_dcs);
```
maybe this can be done a bit more cleanly by throwing the 2deg if statement into the first ```if ($phys eq 'cam6')``` block?

My proposal is to also add ```phys="cam6"``` to the 2deg tunings, requiring modifying two entries in namelist_defaults:

```
<clubb_gamma_coef hgrid="1.9x2.5" phys="cam6" > 0.280   </clubb_gamma_coef>
```
...
```
<seasalt_emis_scale ver="strat" clubb_sgs="1" hgrid="1.9x2.5" phys="cam6">1.10D0</seasalt_emis_scale>
```

I'm not too concerned if this fails baselines (e.g., 2deg runs not running cam6); these 2deg tunings are intended to be specific to cam6.

bld/build-namelist
adamrher: Not sure about all 20, but in terms of the CMIP6 CESM2 FV1deg workhorse, technically, that has **not** been scientifically supported since we transitioned from cesm2.1.X for cesm2.2 or later, and started adding in new bells and whistles activated by the cam6 compsets.

bld/build-namelist
adamrher: my proposal above (https://github.com/ESCOMP/CAM/pull/820#discussion_r1235619211) addresses this convo as well. Will resolve this convo when we resolve that convo. Those mods should actually cause this baseline to pass, I think, but that's more of a bonus. Based on this conversation I'm of the opinion it doesn't matter too much if this baseline fails since it was a unique combo of cam6 and mg3, and therefore susceptible to changing the mg3 defaults.

And in response to my other claim

```
I think that we need to ensure that tests using cam6 compsets (F2000climo,FHIST,etc...) are not xml'ing in mg3. I can't think of why this was done in the first place, but it's possible I'm missing something.
```

Brian made clear that I was missing something; these exotic test combinations are useful as our user base will use exotic combinations, and so its nice to get these baseline failures so it reminds us how changing defaults changes impacts exotic combos.

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: > maybe this can be done a bit more cleanly by throwing the 2deg if statement into the first if ($phys eq 'cam6') block?

I agree.

cacraigucar: I have already put it and the SILHS check in the upper logic and am running tests now.  If it seems to be behaving properly, I'll go ahead and commit my changes so folks can see the proposed implementation.

brian-eaton (approved):

bld/build-namelist
```diff
@@ -3057,6 +3056,31 @@ if ($cfg->get('microphys') =~ /^mg/) {
...
+            $micro_mg_dcs = '200.D-6';       # default for FV 2-deg
+        }
+        elsif ($phys eq 'cam6') {
+            $micro_mg_dcs = '500.D-6';   # default for cam6/mg2
+        }
+        elsif ($phys eq 'cam_dev') {
+            $micro_mg_dcs = '250.D-6';   # default for cam_dev/mg2
+        }
+    }
+
```
adamrher, line 3081: remove the ".../mg2" from the comments "default for ..."

adamrher (changes requested):
You did not revert the tau_0_ubc logic block back to how it was before. See this convo that I will resolve once completed: https://github.com/ESCOMP/CAM/pull/820/files#r1235512849

bld/build-namelist
cacraigucar: Sorry - I missed this when I made the edits.  Do I get "brownie points" for the issue I opened about making sure we review our `science_support` tags before our next code freeze?   The change has been made to build-namelist now.

bld/build-namelist
cacraigucar: done

bld/build-namelist
adamrher: Once you insert the cam_dev hammer back into this logic block (sorry if I wasn't clear on that), you'll get ur brownie points for the science support issue.

```
} elsif ($phys =~ /cam_dev/) {
    add_default($nl, 'tau_0_ubc', 'val'=>'.true.');
```

cacraigucar: Added the cam_dev hammer back in.  Looking for my brownie points now?

adamrher: brownie points granted. see brian dobbins for sweets.

adamrher (approved):

nusbaume (approved):

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2043,10 +2055,11 @@
 <clubb_l_mono_flux_lim_rtm                        > .true.  </clubb_l_mono_flux_lim_rtm>
 <clubb_l_mono_flux_lim_spikefix                   > .true.  </clubb_l_mono_flux_lim_spikefix>
 <clubb_l_mono_flux_lim_thlm                       > .true.  </clubb_l_mono_flux_lim_thlm>
-<clubb_l_mono_flux_lim_um                         > .false. </clubb_l_mono_flux_lim_um>
-<clubb_l_mono_flux_lim_vm                         > .false. </clubb_l_mono_flux_lim_vm>
+<clubb_l_mono_flux_lim_um                         > .true.  </clubb_l_mono_flux_lim_um>
+<clubb_l_mono_flux_lim_vm                         > .true.  </clubb_l_mono_flux_lim_vm>
```
cacraigucar, line 2059: I have an unexpected answer change for the SILHS regression test and have confirmed it is due to these two namelist settings being changed.  Is this expected and more importantly, are these the appropriate settings for SILHS?
@bstephens82 @adamrher @JulioTBacmeister @PeterHjortLauritzen

adamrher: @cacraigucar can you confirm this failing test does **not** invoke cam_dev?

@bstephens82 if Cheryl confirms this not a cam_dev test, then I think this means that the u/v mfl limiters are consequential for running diagnostic momentum fluxes (since only cam_dev can toggle on predict_upwp=T), contrary to what we thought.

cacraigucar:
@adamrher - I confimed that `cam_physpkg = 'cam6'`.  This was the only cheyenne test which had answer changes which I did not expect (all the rest were cam_dev).  

Should I halt the regression tests as we'll be changing something?

bld/namelist_files/namelist_defaults_cam.xml
adamrher: Ah I see what's happening. For some reason, that predates this PR, there is this namelist entry:

```
<clubb_l_predict_upwp_vpwp              silhs="1" > .true.  </clubb_l_predict_upwp_vpwp>
```

I have no idea who did this or why. Nonetheless this would explain the baseline failure. @bstephens82 do you know about this? I'm inclined to delete this line from namelist_defaults.xml in this PR unless someone can tell me why we are turning on prognostic momentum with silhs.

cacraigucar: @adamrher - I confirmed that this was the setting from day 1 when SILHS was introduced into CAM.

bstephens82: Yeah I don't know about that. Im not aware of a connection between SILHS and prognostic momentum fluxes.

adamrher: hmm ... my gut is now saying b/c this is grandfathered in, and plus SILHS is a bit of a mess right now, that we shouldn't mess with SILHS in this PR and perhaps just accept the baseline failure.

But I can only make sense of this baseline failure if there was also a silhs baseline failure for the last clubb externals update PR, where we introduced these nl's (```clubb_l_mono_flux_lim_um```/```clubb_l_mono_flux_lim_vm```) and had them default to false.

cacraigucar: In cam6_3_091, when the CLUBB and SILHS libraries were updated, the SILHS regression test along with all the CLUBB tests had answer changes and the explanation was "Expected answer changes due to new CLUBB code".

adamrher: ah, should've seen that coming. I say proceed with the testing and accept the silhs baseline failure. good to know going forward that silhs has prognostic momentum on from its inception.

PR #855, opened 2023-07-14
Bug fix for SILHS history output being garbled.

### LINKED ISSUE #854: Subcolumn output getting scrambled
### What happened?

I was trying to run SILHS with cam6 physics (with an old release tag, cam_cesm2_2_rel), and I found that the SILHS output in the history files seemed weird and scrambled.  Probably the lower half of the vertical levels (i.e. the levels closer to the ground) were all populated with zeros, while the upper levels were nonzero but seemed unphysical, jumping around weirdly and also sometimes showing very large values of vertical velocity, i.e. on the order of 10,000 m/s or more.  From runtime print statements it became clear that what was happening in the code was not being represented correctly in the history files.

Check out ESCOMP/CAM and use probably any tag since `cam_cesm2_2_rel`, create a new case with FHIST compset and f09_f09_mg17 res, and set CAM_CONFIG_OPTS="-phys cam6 -nlev 58 -silhs -psubcols 4", compile and run with some SILHS output, for example fincl1='SILHS_WM_SCOL'.

You will probably also need `fv_nspltrac=2` in user_nl_cam.

cam_cesm2_2_rel and cam6_3_117

CISL machine (e.g. cheyenne)

Intel

/glade/scratch/stepheba/lidar_paper

Yes

### Extra info

I think I figured out what the problem is with this bug.

From runtime print statements it became clear that what was happening in the SILHS code was not being represented correctly in the history files.  Somehow it was getting scrambled or garbled.  So I looked into the [outfld call](https://github.com/ESCOMP/CAM/blob/6019d2de13e9eb5c82b233da18f5a2286f374313/src/physics/cam/subcol_SILHS.F90#L1628) (this is linking to the current trunk of cam_development, but I think the problem still exists in the current code).  

Eventually I tracked the problem down to [this block of code](https://github.com/ESCOMP/CAM/blob/6019d2de13e9eb5c82b233da18f5a2286f374313/src/control/cam_history.F90#L5875C1-L5884C11) in src/control/cam_history.F90: 

    if (size(dimnames) > 0) then
      if (trim(dimnames(1)) == 'psubcols') then
        if (listentry%field%decomp_type /= cam_grid_id('physgrid')) then
          write(errormsg, *) "Cannot add ", trim(fname),                      &
               "Subcolumn history output only allowed on physgrid"
          call endrun("ADDFLD: "//errormsg)
          listentry%field%is_subcol = .true.
        end if
      end if
    end if

In particular you can see here that the line "listentry%field%is_subcol = .true." is inside the innermost if statement, which looks weird to me.  I think the point of this block of code is to decide whether a field is a subcolumn or not, and so if dimnames(1) is psubcols, it should set is_subcol=T.  But becuase the line where this happens is in the innermost if block, where endrun is being called, I think this doesn't happen the way it should.  In other words, I think this block of code should look like this:

     if (size(dimnames) > 0) then
      if (trim(dimnames(1)) == 'psubcols') then
        if (listentry%field%decomp_type /= cam_grid_id('physgrid')) then
          write(errormsg, *) "Cannot add ", trim(fname),                      &
               "Subcolumn history output only allowed on physgrid"
          call endrun("ADDFLD: "//errormsg)
        end if
          listentry%field%is_subcol = .true.
      end if
    end if  

with the is_subcol=T statement put within the second nested if statement. When I make this change, it appears to resolve the problem for the SILHS history output, both in the old tag and in the newest tag that I've tried.

bstephens82:
Closes #854

### DIFF
--- a/src/control/cam_history.F90
+++ b/src/control/cam_history.F90
@@ -5878,8 +5878,8 @@ end subroutine print_active_fldlst
           write(errormsg, *) "Cannot add ", trim(fname),                      &
                "Subcolumn history output only allowed on physgrid"
           call endrun("ADDFLD: "//errormsg)
-          listentry%field%is_subcol = .true.
         end if
+        listentry%field%is_subcol = .true.
       end if
     end if
     ! Levels


gold2718 (approved):
Looks correct.

Also -- oops (assuming it was me but it goes back before CAM6 and I forgot my svn pw)!

cacraigucar:
@brian-eaton - perhaps this could also be added to your upcoming misc PR - since the work is all done.

nusbaume:
These changes were brought in via PR https://github.com/ESCOMP/CAM/pull/937, so closing this PR now.

Issue #190, opened 2020-08-18
Reimplement Convective Organization Augmentations


swrneale:
Reconfiguring and reimplementing Zhang-McFarlabe code from a while back. Shouldn't be too difficult I hope, but will involve questions about the physics buffer and maybe CCPP down the line.

PR #890, opened 2023-09-29
ZM clean up in preparation for using via CCPP and remove zmconv_microp feature

### LINKED ISSUE #889: Remove microphysics embedded in ZM
### Issue Type

Code Clean-up

### Issue Description

Scientists have indicated that the microphysics embedded in ZM is not the correct implementation of doing this.  It will be removed in anticipation of CCPP'izing ZM.

An email was sent off to a user who is using this capability in an older CESM release and they indicated that it would not interfere with their work.

### Will this change answers?

No

Yes

cacraigucar:
This PR is the first step in the CCPP conversion process.  It has the ZM module separated into its key components and all of the CAM "use" statements lower down have been removed and variables are passed in the calling lists.

zmconv_microp and the zm_microphysics are being removed with this PR as well.  It was deemed that this implementation is not the correct way to embed the microphysics in ZM.

Closes #889


bld/configure
```diff
@@ -1802,6 +1802,9 @@ if ($usr_cppdefs and $print>=2) { print "Commandline CPP definitions: \'$usr_cpp
 # the CPP definitions that were explicitly set in the defaults file or by the user on the commandline.
 my $cfg_cppdefs = ' ';
 
+# Turn on CCPP "OLD_CAM"
+$cfg_cppdefs .= " -DOLD_CAM";
```
nusbaume, line 1806: Is this used anymore?  If not then I might remove it.

nusbaume (approved):
Looks fine to me!  I have some minor concerns that will need to be addressed eventually before conversion is complete, but nothing  that should hold up this PR (especially if it is needed for science work).

mwaxmonsky (approved):
Looks great!  Had a few questions discussed during our weekly SIMA meeting so no need for any changes that I can see.

Thanks for tackling this one @cacraigucar!

src/physics/cam/zm_conv_intr.F90
```diff
@@ -155,6 +155,7 @@ subroutine zm_conv_register
    ! convective mass fluxes
    call pbuf_add_field('CMFMC_DP', 'physpkg', dtype_r8, (/pcols,pverp/), mconzm_idx)
 
+!CACNOTE - Is zm_org really a constituent or was it just a handy structure to use for an allocatable which persists in the run?
```
nusbaume, line 158: Question for @swrneale

src/physics/cam/zm_conv_intr.F90
```diff
@@ -485,7 +486,10 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
    integer  :: ii
 
    real(r8),pointer :: zm_org2d(:,:)
-   real(r8),pointer :: orgt(:,:), org(:,:)
+   real(r8),allocatable :: orgt_alloc(:,:), org_alloc(:,:)
```
nusbaume, line 489: I don't think these two variables are needed, as they don't appear to be used anywhere (besides the allocation step below).

src/physics/cam/zm_conv_intr.F90
```diff
@@ -485,7 +486,10 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
...
+
+   real(r8) :: zm_org2d_noalloc(state%ncol,pver)
+   real(r8) :: orgt_noalloc(state%ncol,pver), org_noalloc(state%ncol,pver)
```
nusbaume, line 492: I might rename these variable to something like `zm_org2d_ncol` and `orgt_ncol`, because if you removed the unused `alloc` variables above then the `noalloc` statements won't really make much sense.

src/physics/cam/zm_conv_intr.F90
```diff
@@ -855,11 +911,16 @@ subroutine zm_conv_tend_2( state,  ptend,  ztodt, pbuf)
...
-                     nstep,   fracis,  ptend%q, dpdry, ztodt)
+
+!REMOVECAM - no longer need this when CAM is retired and pcols no longer exists
+   ptend%q(:,:,:) = 0._r8
+!REMOVECAM_END
+
+      call zm_conv_convtran_run (ncol, pver,          &
+                  ptend%lq,state%q(:ncol,:,:), pcnst,  mu(:ncol,:), md(:ncol,:),   &
+                  du(:ncol,:), eu(:ncol,:), ed(:ncol,:), dp(:ncol,:), dsubcld(:ncol),  &
+                  jt(:ncol), maxg(:ncol), ideep(:ncol), 1, lengath,  &
```
nusbaume, line 922: I am probably being paranoid, but there should be a check sometime before this call to check if `lengath` is bigger than `ncol`, and if so then just use `ncol` here for `lengath`?  Otherwise it seems at least theoretically possible for `lengath` to cause an array out-of-bounds error inside the CCPP-ized code.

src/physics/spcam/crmclouds_camaerosols.F90
```diff
@@ -732,11 +732,14 @@ subroutine crmclouds_convect_tend(state,  ptend,  ztodt,  pbuf)
...
-                  du(:,:), eu(:,:), ed(:,:), dp(:,:), dsubcld(:),  &
-                  jt(:),maxg(:),ideep(:), 1, lengath,  &
-                  nstep,   fracis,  ptend%q, dpdry, ztodt  )
+   !REMOVECAM - no longer need these when CAM is retired and pcols no longer exists
+   ptend%q(:,:,:) = 0._r8
+   !REMOVECAM_END
+   call zm_conv_convtran_run (ncol,pver,                                    &
+                  ptend%lq,state%q(:ncol,:,:), pcnst,  mu(:ncol,:), md(:ncol,:),   &
+                  du(:ncol,:), eu(:ncol,:), ed(:ncol,:), dp(:ncol,:), dsubcld(:ncol),  &
+                  jt(:ncol),maxg(:ncol),ideep(:ncol), 1, lengath,  &
```
nusbaume, line 741: Same note here about ensuring that `lengath` is not bigger than `ncol`.

nusbaume (changes requested):
Have some minor cleanup and a question about the `lengath` variable, but once those are resolved then this PR should be ready to go!

src/physics/cam/zm_conv_intr.F90
swrneale: They are not currently needed, but it is part of my ancient yet grand plan to include convective organization. I have a dusty issue on this #190 . Feel free to remove 'org' related infrastructure, I can just refer back to an old tag to remind myself of the implementation if need be.

src/physics/cam/zm_conv_intr.F90
cacraigucar: renamed

mwaxmonsky (approved):
Looks great @cacraigucar!

src/physics/cam/zm_conv_intr.F90
cacraigucar: Good catch.  I had several different flavors of variables while I was trying to get this all to work.  I neglected to delete these two.  Since @swrneale is going to be working on implementing this, I will leave the 'org' things here.

nusbaume (approved):
Looks great now, thanks @cacraigucar!

PR #869, opened 2023-08-03
Change "time" value for averaged quantities to midpoint of averaging period

peverwhee:
### Overview
Per #159 - this PR changes the "time" variable in a history file with averaged (and/or min/max/stdev quantities) to be the midpoint of the averaging period (basically, the average of the time_bounds variable).

Side affects of the above/related changes in this PR (all implemented):
* Instantaneous history files (avgflag_pertape = 'I') cannot have averaged fields due to the time mismatch
* Averaged history files (avgflag_pertape /= 'I') cannot have instantaneous fields (same reason)
* Changed 'time_bnds' variable name to 'time_bounds'
* Ensured the 'cell_methods' metadata field always includes "time: XXXXX" (where XXXXX refers to the avgflag) for history fields so that it's always clear what operation is being done on/for that field

### Tests performed locally
* Tested out-of-the-box Kessler (h0 is instantaneous) - runs and time field is current time
* Tested trying to add an average field to h0 - fails as expected
* Tested adding an average tape (h1) - runs and time field is midpoint(s)
* Tested adding an instantaneous field to h1 - fails as expected
* Tested adding instantaneous field (":I") to h0 and average field (":M") to h1 - works as expected

Related to: https://github.com/ESCOMP/CTSM/pull/2019

closes #159

PR #903, opened 2023-10-13
cam6_3_140: Separate history tapes into hXi and hXa

### LINKED ISSUE #159: Change "time" value for averaged quantities to midpoint of averaging period
A longstanding quirk of CAM is that fields saved as averages have the "time" coordinate value of the end of the averaging period. A January monthly average has a timestamp of 1 February, for example. Changing the "time" coordinate value to represent the middle of the averaging period assists our users in understanding CAM output and doing proper analysis.

Fields saved as instantaneous values don't need to be changed at all.

Having one time axis for instantaneous fields and one for averaged fields would allow both kinds of output to reside in the same file, but netCDF-3 doesn't allow for more than one UNLIMITED dimension in a single file. netCDF-4 allows multiple UNLIMITED dimensions, however, pnetcdf is built on netCDF-3, so multiple time axes aren't possible.

### LINKED ISSUE #554: Modify naming and attributes of time variables on history files to be consistent with other CESM components
These changes are being requested to improve the ease-of-access of CESM data and to normalize the look of the metadata across CESM components. See https://github.com/ESCOMP/CESM/issues/194 and especially the google doc referenced from https://github.com/ESCOMP/CESM/issues/194#issuecomment-961443965.

Current CAM time array output:

	double time(time) ;
		time:long_name = "time" ;
		time:units = "days since YYYY-MM-DD HH:MM:SS" ;
		time:calendar = "noleap" ;
		time:bounds = "time_bnds" ;
	double time_bnds(time, nbnd) ;
		time_bnds:long_name = "time interval endpoints" ;

Proposed CAM time array output:

	double time(time) ;
		time:long_name = "time" ;
		time:units = "days since YYYY-MM-DD HH:MM:SS" ;
		time:calendar = "noleap" ;
		time:bounds = "time_bounds" ;
	double time_bounds(time, nbnd) ;
		time_bounds:long_name = "time interval endpoints" ;
		time_bounds:units = "days since YYYY-MM-DD HH:MM:SS" ;
		time_bounds:calendar = "noleap" ;

Changes made: time_bnds renamed as time_bounds, added units/calendar attributes to time_bounds, change bounds attribute of time to time_bounds

@billsacks @phillips-ad @strandwg are filing similar issues for each component.

peverwhee:
### Overview
Per https://github.com/ESCOMP/CAM/issues/159 - this PR changes the "time" variable in a history file with averaged (and/or min/max/stdev quantities) to be the midpoint of the averaging period (basically, the average of the time_bounds variable).

Side affects of the above/related changes in this PR (all implemented):
* Each history tape _can_ now result in TWO files - hXa and hXi. The 'i' file is ALWAYS created and will, at minimum, contain time-dependent variables (like solar forcing data, etc). If only the instantaneous file is created, it will still carry the 'i' flag in the filename
* Changed 'time_bnds' variable name to 'time_bounds'
* Ensured the 'cell_methods' metadata field always includes "time: XXXXX" (where XXXXX refers to the avgflag) for history fields so that it's always clear what operation is being done on/for that field
* Any user-specified history file name or template will also have "a" or "i" appended to it and may result in two files
* `time`, `date`, `datesec` are all calculated based on the midpoint time for the accumulated file

### Implementation
* amending the active_entry DDT "File" property to be an array of files called "Files" and adding a num_files attribute to determine if we have one or two files
* iterating over Files array and adding all necessary metadata and variables to _both_ "i" and "a" files when necessary

### Testing
* All regression tests run successfully (except for known failures)
* Wrote a script to compare the results vs baselines; confirmed no fields were lost in the transition and no answers changed (except the expected midpoint time variables on the accumulated file)

Related to: https://github.com/ESCOMP/CTSM/pull/2019

closes #159 
closes #554 
Supersedes #869


src/control/cam_history.F90
```diff
@@ -126,7 +127,7 @@ module cam_history
   !   The size of these parameters should match the assignments in restart_vars_setnames and restart_dims_setnames below
   !
   integer, parameter :: restartvarcnt              = 45
-  integer, parameter :: restartdimcnt              = 10
+  integer, parameter :: restartdimcnt              = 11
```
cacraigucar, line 130: Why did this get bumped up to 11?

cacraigucar (commented):
Asking a basic question to aid my understanding during the review process

src/control/cam_history.F90
peverwhee: I had to add a new restart dimension for the variables that hold the names of the history files (nhfil and cpath). The new dimension is defined on ~line 1424 and is set to the parameter maxsplitfiles(=2) from cam_grid_support.

nhfil and cpath used to just be have one dimension which was the # of tapes, but now there's a second dimension for the two potential split files.

src/control/cam_history.F90
```diff
@@ -3933,103 +3975,116 @@ subroutine h_inquire (t)
...
+               ! this is the accumulated file - skip instantaneous fields
+               if (tape(t)%hlist(fld)%avgflag == 'I') then
+                  cycle
+               end if
+            else
+               ! this is the instantaneous file - skip accumulated fields
+               if (tape(t)%hlist(fld)%avgflag /= 'I') then
+                  cycle
+               end if
+            end if
```
cacraigucar, line 4042: This is a little fragile - assuming that the first file is accumulated and the second is instantaneous.  With this current implementation, that is the case, but if someone flipped them down at line 5650, then it would no longer work.  Since this history is nearing end-of-life and the likelihood of it getting broken is low, I won't require this be fixed, but if it is easy to determine which is which, I'd suggest that be used instead.

src/control/cam_history.F90
```diff
@@ -2392,7 +2434,7 @@ character(len=max_string_len) function get_hfilepath( tape )
     !
     integer, intent(in) :: tape  ! Tape number
 
-    get_hfilepath = cpath( tape )
+    get_hfilepath = cpath( tape, 1 )
```
cacraigucar, line 2437: I was questioning if this should be for both files or just the first.   When I searched through the code tree, I do not see it being used at all.  I will still ask the question though

src/control/cam_history.F90
```diff
@@ -4380,412 +4409,495 @@ subroutine h_define (t, restart)
...
+               ! this is the accumulated file - skip instantaneous fields
+               if (tape(t)%hlist(fld)%avgflag == 'I') then
+                  cycle
+               end if
+            else
+               ! this is the instantaneous file - skip accumulated fields
+               if (tape(t)%hlist(fld)%avgflag /= 'I') then
+                  cycle
+               end if
+            end if
```
cacraigucar, line 4653: If there is an easy way to determine which file is accumulated instead of just relying on it being the first, I would add that here as well.

src/control/cam_history.F90
```diff
@@ -4380,412 +4409,495 @@ subroutine h_define (t, restart)
...
+           if (tape(t)%hlist(fld)%avgflag == 'I') then
+              str = 'instantaneous'
+           else
+              str = tape(t)%hlist(fld)%time_op
+           end if
+           if (len_trim(cell_methods) > 0) then
+              cell_methods = trim(cell_methods)//' '//'time: '//str
+           else
+              cell_methods = trim(cell_methods)//'time: '//str
+           end if
```
cacraigucar, line 4833: Previously this was not being set for instantaneous fields.  Does this matter?

cacraigucar (changes requested):
My brain is officially fried and I'm just doing the review!!  I am impressed with the changes you've made!  

First batch of review comments - more may be coming tomorrow

src/control/cam_history.F90
```diff
@@ -2342,27 +2369,42 @@ subroutine read_restart_history (File)
...
+            write(iulog,*)'READ_RESTART_HISTORY: nf_close(',t,')=',nhfil(t,1), mfilt(t)
           end if
           do f=1,nflds(t)
             deallocate(tape(t)%hlist(f)%varid)
             nullify(tape(t)%hlist(f)%varid)
           end do
-          call cam_pio_closefile(tape(t)%File)
+          do f = 1, tape(t)%num_files
+             call cam_pio_closefile(tape(t)%Files(f))
+          end do
```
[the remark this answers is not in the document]
gold2718: I have not looked much into this but I'm pretty sure there used to be a restart history file for any history file that had non-Instantaneous (last value) fields. Has that changed with this PR?

src/control/cam_history.F90
gold2718: Maybe assign a parameter to the instantaneous file?

src/control/cam_history.F90
```diff
@@ -5187,128 +5303,128 @@ subroutine dump_field (f, t, restart)
...
                  mdimsize, ncreal, fdecomp)
-          else if (tape(t)%hlist(f)%field%zonal_complement > 0) then
+          else if (tape(t)%hlist(fld)%field%zonal_complement > 0) then
             ! We don't want to double write so do nothing here
 !            compind = tape(t)%hlist(f)%field%zonal_complement
 !            compid => tape(t)%hlist(compind)%varid(index)
-!            call write_interpolated(tape(t)%File, compid, varid,              &
+!            call write_interpolated(tape(t)%Files(f), compid, varid,              &
 !                 tape(t)%hlist(compind)%hbuf, tape(t)%hlist(f)%hbuf,          &
 !                 mdimsize, PIO_DOUBLE, fdecomp)
```
cacraigucar, line 5350: This  commented out section is wrong with hlist(f).  I suggest you either delete the section or fix the index

src/control/cam_history.F90
```diff
@@ -2117,6 +2140,11 @@ subroutine read_restart_history (File)
     gridsontape = -1
     do t = 1, ptapes
       do f = 1, nflds(t)
+        if (tape(t)%hlist(f)%avgflag .eq. 'I') then
+           hfile_inst(t) = .true.
+        else
+           hfile_accum(t) = .true.
+        end if
```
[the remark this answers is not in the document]
cacraigucar: There are still some functions which use "f".  I did a `grep hlist\(f\)` to find them.  I would suggest it might be good to be fully consistent throughout this routine.

cacraigucar (changes requested):

src/control/cam_history.F90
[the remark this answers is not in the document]
peverwhee: yep - good call. I missed the ones outside of a loop context. fixed.

src/control/cam_history.F90
```diff
@@ -4234,33 +4289,60 @@ subroutine h_define (t, restart)
...
     end if
     if(is_satfile(t)) then
       interpolate = .false. ! !!XXgoldyXX: Do we ever want to support this?
       patch_output = .false.
-      call cam_pio_def_dim(tape(t)%File, 'ncol', pio_unlimited, timdim)
-      call cam_pio_def_dim(tape(t)%File, 'nbnd', 2, bnddim)
+      do f = 1, tape(t)%num_files
+         call cam_pio_def_dim(tape(t)%Files(f), 'ncol', pio_unlimited, timdim)
+         call cam_pio_def_dim(tape(t)%Files(f), 'nbnd', 2, bnddim)
+      end do
```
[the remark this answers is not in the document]
peverwhee: loop removed.

src/control/cam_history.F90
```diff
@@ -4234,33 +4289,60 @@ subroutine h_define (t, restart)
...
+         call cam_pio_def_var(tape(t)%Files(f), 'lat', pio_double, (/timdim/),       &
+              latvar(1)%vd)
+         ierr=pio_put_att (tape(t)%Files(f), latvar(1)%vd, 'long_name', 'latitude')
+         ierr=pio_put_att (tape(t)%Files(f), latvar(1)%vd, 'units', 'degrees_north')
+
+         call cam_pio_def_var(tape(t)%Files(f), 'lon', pio_double, (/timdim/),       &
+              lonvar(1)%vd)
+         ierr=pio_put_att (tape(t)%Files(f), lonvar(1)%vd,'long_name','longitude')
+         ierr=pio_put_att (tape(t)%Files(f), lonvar(1)%vd,'units','degrees_east')
+      end do
```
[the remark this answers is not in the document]
peverwhee: loop removed

src/control/cam_history.F90
peverwhee: before the decision to split the files, this discussion here: https://github.com/ESCOMP/CTSM/pull/2019#issuecomment-1660639903

was about how to tell what type of file it is. For consistency with what CLM has, I added the instantaneous cell_value string. But I also see how this isn't super necessary anymore because the filename has the specifier in it already...

cacraigucar: I'll leave it up to the developer on what is best - I was just asking the question...

src/control/cam_history.F90
peverwhee: I agree @cacraigucar that this is fragile. Good point.

@gold2718 what do you mean by "assign a parameter to the instantaneous file"?

My initial thought is to add "accumulate_file_index" and "instantaneous_file_index" parameters and use those instead of "1" and "2" throughout the code. But I'm open to other options (I just don't yet understand yours...)

src/control/cam_history.F90
```diff
@@ -6239,8 +6440,10 @@ subroutine wrapup (rstwr, nlend)
         ! Is this the 0 timestep data of a monthly run?
         ! If so, just close primary unit do not dispose.
         !
-        if (masterproc) write(iulog,*)'WRAPUP: nf_close(',t,')=',trim(nhfil(t))
-        if(pio_file_is_open(tape(t)%File)) then
+        if (masterproc) then
+           write(iulog,*)'WRAPUP: nf_close(',t,')=',trim(nhfil(t,1))
+        end if
+        if(pio_file_is_open(tape(t)%Files(1))) then
```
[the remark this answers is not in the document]
peverwhee: This shouldn't need to loop over files. The split history files don't have separate field lists, so a second loop through here wouldn't do anything (all the field varids will have already been deallocated).

src/control/cam_history.F90
gold2718: That is exactly what I was talking about, just using real language :)
I think those parameter names are fine except that I still don't really like 'instantaneous' because that does not describe which instant (I like last value because you get the last sample you happened to talk before writing the file).

cacraigucar: I would encourage you to keep "instantaneous" as this is what the ultimate file is called.  If we were to change to "last value", we'd need to overhaul everything (including changing the "I" designator) and coordinate with CTSM.  I believe "I" is what we are currently stuck with.

peverwhee: @cacraigucar agreed. Parameters added.

src/control/cam_history.F90
peverwhee: Since it's not used, I also pretty much ignored it. But, for the sake of cleanliness, I amended the routine to take in a flag and either output the accumulated or instantaneous filepath

src/control/cam_history.F90
peverwhee: parameter added for the indices - check is now against accumulated_file_index

src/control/cam_history.F90
peverwhee: fixed the index.

src/control/cam_history.F90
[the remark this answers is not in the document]
peverwhee: Still mulling over this. I _think_ it needs to be a loop. This is after we read the single restart history file (which we closed on line 2339) and when we're setting up the other history files.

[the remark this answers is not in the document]
cacraigucar: It doesn't hurt to keep it as a loop

src/control/cam_history.F90
```diff
@@ -5579,64 +5749,104 @@ subroutine wshist (rgnht_in)
...
+                ! We have two files - one for accumulated and one for instantaneous fields
+                if (f == accumulated_file_index) then
+                   ! accumulated tape - time is midpoint of time_bounds
+                   ierr=pio_put_var (tape(t)%Files(f), tape(t)%timeid, (/start/),(/count1/),(/(tdata(1) + tdata(2)) / 2._r8/))
+                else
+                   ! instantaneous tape - time is current time
+                   ierr=pio_put_var (tape(t)%Files(f), tape(t)%timeid, (/start/),(/count1/),(/time/))
+                end if                   
+             else
+                if (hfile_accum(t) .and. .not. restart) then
```
cacraigucar, line 5763: Could you ever get here with a restart file (i.e. why are you checking restart) since you've already determined that there is more than one file?  If it is important to check for a restart file, then you probably need to do so on the else section as well.

src/control/cam_history.F90
```diff
@@ -5579,64 +5749,104 @@ subroutine wshist (rgnht_in)
...
+              if(associated(tape(t)%hlist(fld)%varid)) then
+                deallocate(tape(t)%hlist(fld)%varid)
+                nullify(tape(t)%hlist(fld)%varid)
               end if
             end do
-            call cam_pio_closefile(tape(t)%File)
+            do f = 1, tape(t)%num_files
+               call cam_pio_closefile(tape(t)%Files(f))
+            end do
           else
```
cacraigucar, line 5846: Since this is a restart, I believe the loop is unnecessary

src/control/cam_history.F90
peverwhee: i would lean toward overkill, but am willing to be overruled.

cacraigucar (changes requested):
Whew - I finished my review and I've resolved the conversations which are fixed.  Again, good job on these changes!

src/control/cam_history.F90
cacraigucar: Since I suggested overkill as well, I'd say go for it!

src/control/cam_history.F90
peverwhee: this else is for when there is only one file. I think it's right to not check for restart in the else section since we want the non-averaged time variable on restart files

src/control/cam_history.F90
peverwhee: removed.

src/control/cam_history.F90
cacraigucar: Oops - the highlighting didn't work too well.  It is the else loop right above this one (the one for the instantaneous tape)

cacraigucar: I see you removed the restart check that I was worried about, so all good now

src/control/cam_history.F90
```diff
@@ -2342,27 +2377,44 @@ subroutine read_restart_history (File)
...
+              call cam_pio_openfile(tape(t)%Files(instantaneous_file_index), locfn, PIO_WRITE)
+           else if (hfile_accum(t)) then
+              tape(t)%num_files = 1
+              call getfil (cpath(t,accumulated_file_index), locfn)
+              call cam_pio_openfile(tape(t)%Files(1), locfn, PIO_WRITE)
+           else if (hfile_inst(t)) then
+              tape(t)%num_files = 1
+              call getfil (cpath(t,instantaneous_file_index), locfn)
+              call cam_pio_openfile(tape(t)%Files(1), locfn, PIO_WRITE)
+           end if
```
cacraigucar, line 2394: Is this section where the aaccumulated_file_index and instantaneous_file_index need to be set (and not as a parameter above?  As it is right now, the instananeous is assigned to 2, but should be 1 if there is only one file.   I believe the "Files(1)" on the two lines here also need to have the variables used instead of "1"

src/control/cam_history.F90
```diff
@@ -4234,34 +4305,57 @@ subroutine h_define (t, restart)
...
+         tape(t)%num_files = 2
+         call cam_pio_createfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), amode)
+         call cam_pio_createfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), amode)
+      else if (hfile_accum(t)) then
+         tape(t)%num_files = 1
+         call cam_pio_createfile (tape(t)%Files(1), nhfil(t,accumulated_file_index), amode)
+      else if (hfile_inst(t)) then
+         tape(t)%num_files = 1
+         call cam_pio_createfile (tape(t)%Files(1), nhfil(t,instantaneous_file_index), amode)
+      end if
```
cacraigucar, line 4340: I think these "Files(1)" should also use the varaibles (after they are assigned in the previous comment.

src/control/cam_history.F90
```diff
@@ -6273,7 +6490,14 @@ subroutine wrapup (rstwr, nlend)
...
           !
           if (.not.nlend .and. .not.lfill(t)) then
-            call cam_PIO_openfile (tape(t)%File, nhfil(t), PIO_WRITE)
+            if (hfile_accum(t) .and. hfile_inst(t)) then
+               call cam_PIO_openfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), PIO_WRITE)
+               call cam_PIO_openfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), PIO_WRITE)
+            else if (hfile_accum(t)) then
+               call cam_PIO_openfile (tape(t)%Files(1), nhfil(t,accumulated_file_index), PIO_WRITE)
+            else if (hfile_inst(t)) then
+               call cam_PIO_openfile (tape(t)%Files(1), nhfil(t,instantaneous_file_index), PIO_WRITE)
```
cacraigucar, line 6499: I think these "Files(1)" should also use the varaibles (after they are assigned in the previous comment.

cacraigucar (changes requested):
Thanks for adding those new parameters.  It helped me find some areas which I believe are bugs.

src/control/cam_history.F90
peverwhee: Yeah, the indexing here was technically correct (since the index parameters are set before we know how many files we have), but very confusing. As a result, I've changed the looping over num_files to instead be over the size of Files and then checking whether or not each file is open before operating on it. For example, if you only have an instantaneous file, you'd only be doing things during the second iteration of the loop because instantaneous_file_index = 2

TLDR: these indexes are now set to the parameters.

src/control/cam_history.F90
peverwhee: same here.

src/control/cam_history.F90
peverwhee: and also here.

src/control/cam_history.F90
```diff
@@ -5187,128 +5333,128 @@ subroutine dump_field (f, t, restart)
...
+          else if (tape(t)%hlist(fld)%field%zonal_complement > 0) then
             ! We don't want to double write so do nothing here
-!            compind = tape(t)%hlist(f)%field%zonal_complement
+!            compind = tape(t)%hlist(fld)%field%zonal_complement
 !            compid => tape(t)%hlist(compind)%varid(index)
-!            call write_interpolated(tape(t)%File, compid, varid,              &
-!                 tape(t)%hlist(compind)%hbuf, tape(t)%hlist(f)%hbuf,          &
+!            call write_interpolated(tape(t)%Files(f), compid, varid,              &
+!                 tape(t)%hlist(compind)%hbuf, tape(t)%hlist(fld)%hbuf,          &
 !                 mdimsize, PIO_DOUBLE, fdecomp)
```
nusbaume, line 5380: I know you didn't write this, but it seems like we should be doing nothing here, which means I might vote to remove the commented out code.

src/control/cam_history.F90
```diff
@@ -5486,28 +5638,53 @@ subroutine wshist (rgnht_in)
...
           !
-          do f = 1, ptapes
-            if (masterproc.and. trim(fname) == trim(nhfil(f)) )then
-              write(iulog,*)'WSHIST: New filename same as old file = ', trim(fname)
-              write(iulog,*)'Is there an error in your filename specifiers?'
-              write(iulog,*)'hfilename_spec(', t, ') = ', hfilename_spec(t)
-              if ( t /= f )then
-                write(iulog,*)'hfilename_spec(', f, ') = ', hfilename_spec(f)
+          duplicate = .false.
+          do f = 1, t
```
nusbaume, line 5650: Should we only be looping to `t` here, or should we be looping through all of the tapes (`ptapes`), which is what it looks like it was doing in the original version?

src/control/cam_history.F90
```diff
@@ -5486,28 +5638,53 @@ subroutine wshist (rgnht_in)
...
+                 write(iulog,*)'WSHIST: New instantaneous filename same as old file = ', trim(fname_inst)
+                 duplicate = .true.
+              end if
+              if (duplicate) then
+                 write(iulog,*)'Is there an error in your filename specifiers?'
+                 write(iulog,*)'hfilename_spec(', t, ') = ', trim(hfilename_spec(t))
+                 if ( t /= f )then
+                   write(iulog,*)'hfilename_spec(', f, ') = ', trim(hfilename_spec(f))
+                 end if
+                 call endrun
```
nusbaume, line 5668: I would add some sort of error message to this `endrun` call (even it just says to look at the atm log file).

src/control/cam_history.F90
```diff
@@ -6273,7 +6518,14 @@ subroutine wrapup (rstwr, nlend)
...
           if (.not.nlend .and. .not.lfill(t)) then
-            call cam_PIO_openfile (tape(t)%File, nhfil(t), PIO_WRITE)
+            if (hfile_accum(t) .and. hfile_inst(t)) then
+               call cam_PIO_openfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), PIO_WRITE)
+               call cam_PIO_openfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), PIO_WRITE)
+            else if (hfile_accum(t)) then
+               call cam_PIO_openfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), PIO_WRITE)
+            else if (hfile_inst(t)) then
+               call cam_PIO_openfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), PIO_WRITE)
+            end if
```
nusbaume, line 6528: I don't think you need the `.and.` option here:

```suggestion
            if (hfile_accum(t)) then
               call cam_PIO_openfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), PIO_WRITE)
            end if
            if (hfile_inst(t)) then
               call cam_PIO_openfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), PIO_WRITE)
            end if
```

nusbaume (changes requested):
Great work @peverwhee!  I mostly just have some minor clean-up suggestions.

src/control/cam_history.F90
brian-eaton: @gold2718, I'm not understanding your dislike of referring to a file as "instantaneous".  That file contains instantaneous time samples, i.e., the values of the state variables at the end of discrete time steps.  The suggestion of "last value" is an implementation detail related to the use of history buffers which doesn't clearly describe the file contents.  What am I missing?

gold2718: The issue in (possibly only) my mind is that there are a (potentially large) number of "instantaneous" samples that are capture by `outfld` calls. However, none of these are written to a history file, only the last captured sample is written out. Perhaps "last_sample" is more accurate. This is not a huge deal but is something that has always bothered me since I have run into folks who didn't understand that it works like that (capturing then overwriting all the samples except the last one before a write).

brian-eaton: @gold2718, I think what we're trying to convey in the filename is something about the temporal characteristics of the data.  Focusing on the implementation of collecting the time samples (e.g., by overwritting the history buffer each time a sample is collected until it's time to write the buffer) ignores that other implementations are possible, like not putting anything in the buffer until the desired timestep is reached.  The sample collection implementation doesn't describe the temporal characteristics of the data.  Instantaneous does.

gold2718: In the new instantaneous files, does the time variable always match the time step of the sample?

brian-eaton: Yes.

It has always been the case that the time coordinate matches the end of the time step producing the data for an instantaneous field.  From the code I've looked at so far the time coordinate in the hXi file is the same as it's always been for instantaneous fields.  I am running tests as I review this PR to check these details.

src/utils/cam_grid_support.F90
```diff
@@ -32,8 +37,8 @@ module cam_grid_support
     integer(iMap),    pointer :: map(:) => NULL()  ! map (dof) for dist. coord
     logical                   :: latitude          ! .false. means longitude
     real(r8),         pointer :: bnds(:,:) => NULL() ! bounds, if present
-    type(var_desc_t), pointer :: vardesc => NULL() ! If we are to write coord
-    type(var_desc_t), pointer :: bndsvdesc => NULL() ! If we are to write bounds
+    type(vardesc_ptr_t)       :: vardesc(2)        ! If we are to write coord
```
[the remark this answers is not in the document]
peverwhee: an excellent point. changed.

src/control/cam_history.F90
```diff
@@ -778,10 +799,15 @@ subroutine history_readnl(nlfile)
...
+            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t%f.%y-%m.nc'
           else
-            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t.%y-%m-%d-%s.nc'
+            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t%f.%y-%m-%d-%s.nc'
           end if
+        else
+           ! Append file type - instantaneous or accumulated - to filename
+           ! specifier provided
+           temp_spec = trim(hfilename_spec(t)) // '%f'
+           hfilename_spec(t) = temp_spec
```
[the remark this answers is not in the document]
peverwhee: I had convinced myself that I needed it when I got an apparently-unrelated error. You're right. Temp variable removed.

src/control/cam_history.F90
```diff
@@ -1742,12 +1773,12 @@ subroutine write_restart_history ( File, &
...
-      start(2) = f
-      ierr = pio_put_var(File, mdimname_desc, start, hist_coord_name(f))
+    do fld = 1, registeredmdims
+      start(2) = fld
+      ierr = pio_put_var(File, mdimname_desc, start, hist_coord_name(fld))
     end do
 
-    deallocate(xyfill, allmdims)
+    deallocate(xyfill, allmdims, is_subcol, interp_output, restarthistory_tape)
     return
```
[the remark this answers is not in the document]
peverwhee: removed

src/control/cam_history.F90
```diff
@@ -2237,90 +2274,90 @@ subroutine read_restart_history (File)
...
+                     fdims(1:nfdims), dimlens(1:ndims), tape(t)%hlist(fld)%sbuf(:,1,:), vdesc)
              end if
           endif
 
-          ierr = pio_inq_varid(tape(t)%File, trim(fname_tmp)//'_nacs', vdesc)
-          call cam_pio_var_info(tape(t)%File, vdesc, nacsdimcnt, dimids, dimlens)
-
+          ierr = pio_inq_varid(tape(t)%Files(restart_file_index), trim(fname_tmp)//'_nacs', vdesc)
+          call cam_pio_var_info(tape(t)%Files(restart_file_index), vdesc, nacsdimcnt, dimids, dimlens)
+
```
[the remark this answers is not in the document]
peverwhee: good call! spaces removed

src/control/cam_history.F90
```diff
@@ -2342,27 +2378,45 @@ subroutine read_restart_history (File)
...
+              call cam_pio_openfile(tape(t)%Files(accumulated_file_index), locfn, PIO_WRITE)
+              call getfil (cpath(t,instantaneous_file_index), locfn)
+              call cam_pio_openfile(tape(t)%Files(instantaneous_file_index), locfn, PIO_WRITE)
+           else if (hfile_accum(t)) then
+              call getfil (cpath(t,accumulated_file_index), locfn)
+              call cam_pio_openfile(tape(t)%Files(accumulated_file_index), locfn, PIO_WRITE)
+           else if (hfile_inst(t)) then
+              call getfil (cpath(t,instantaneous_file_index), locfn)
+              call cam_pio_openfile(tape(t)%Files(instantaneous_file_index), locfn, PIO_WRITE)
+           end if
```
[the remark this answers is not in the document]
peverwhee: indeed! simiplified

src/control/cam_history.F90
```diff
@@ -2342,27 +2378,45 @@ subroutine read_restart_history (File)
...
+            call sat_hist_define(tape(t)%Files(sat_file_index))
           end if
         end if
         !
         ! If the history file is full, close the current unit
         !
         if (nfils(t) >= mfilt(t)) then
           if (masterproc) then
-            write(iulog,*)'READ_RESTART_HISTORY: nf_close(',t,')=',nhfil(t), mfilt(t)
+            do f = 1, size(tape(t)%Files)
```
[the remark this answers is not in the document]
peverwhee: I had waffled on that. I used the size in case some person someday decided to set it to some random variable/integer. But that would cause other problems anyway, so i've switched all the size(tape(t)%Files) to maxsplitfiles

src/control/cam_history.F90
```diff
@@ -4234,34 +4307,52 @@ subroutine h_define (t, restart)
...
-      call cam_pio_createfile (tape(t)%File, nhfil(t), amode)
+      ! figure out how many history files to generate for this tape
+      if (hfile_accum(t) .and. hfile_inst(t)) then
+         call cam_pio_createfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), amode)
+         call cam_pio_createfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), amode)
+      else if (hfile_accum(t)) then
+         call cam_pio_createfile (tape(t)%Files(accumulated_file_index), nhfil(t,accumulated_file_index), amode)
+      else if (hfile_inst(t)) then
+         call cam_pio_createfile (tape(t)%Files(instantaneous_file_index), nhfil(t,instantaneous_file_index), amode)
+      end if
```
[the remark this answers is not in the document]
peverwhee: simplified

src/control/cam_history.F90
```diff
@@ -4380,412 +4426,503 @@ subroutine h_define (t, restart)
...
 
-      ierr=pio_def_var (tape(t)%File,'mdt',PIO_INT,tape(t)%mdtid)
-      ierr=pio_put_att (tape(t)%File, tape(t)%mdtid, 'long_name', 'timestep')
-      ierr=pio_put_att (tape(t)%File, tape(t)%mdtid, 'units', 's')
+! Put these back in when they are filled properly
+!       ierr=pio_put_att (tape(t)%Files(f), PIO_GLOBAL, 'title',ctitle)
+!       ierr= pio_put_att (tape(t)%Files(f), PIO_GLOBAL, 'Version', &
+!            '$Name$')
+!       ierr= pio_put_att (tape(t)%Files(f), PIO_GLOBAL, 'revision_Id', &
+!            '$Id$')
```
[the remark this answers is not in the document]
peverwhee: yep. removed

src/control/cam_history.F90
```diff
@@ -4380,412 +4426,503 @@ subroutine h_define (t, restart)
...
+           ierr=pio_put_att (tape(t)%Files(f), tape(t)%f12vmrid, 'long_name', trim(str))
+
+           ierr=pio_def_var (tape(t)%Files(f),'sol_tsi ',pio_double,(/timdim/),tape(t)%sol_tsiid)
+           str = 'total solar irradiance'
+           ierr=pio_put_att (tape(t)%Files(f), tape(t)%sol_tsiid, 'long_name', trim(str))
+           str = 'W/m2'
+           ierr=pio_put_att (tape(t)%Files(f), tape(t)%sol_tsiid, 'units', trim(str))
+
+           if (solar_parms_on) then
+             ! solar / geomagetic activity indices...
```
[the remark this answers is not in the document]
peverwhee: typo fixed

src/control/cam_history.F90
```diff
@@ -4380,412 +4426,503 @@ subroutine h_define (t, restart)
...
+           if (tape(t)%hlist(fld)%avgflag == 'I') then
+              str = 'instantaneous'
+           else
+              str = tape(t)%hlist(fld)%time_op
+           end if
+           if (len_trim(cell_methods) > 0) then
+              cell_methods = trim(cell_methods)//' '//'time: '//str
+           else
+              cell_methods = trim(cell_methods)//'time: '//str
+           end if
```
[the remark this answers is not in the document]
peverwhee: simplified

src/control/cam_history.F90
```diff
@@ -4380,412 +4426,503 @@ subroutine h_define (t, restart)
...
+                endif
+                call cam_pio_def_var(tape(t)%Files(f), trim(fname_tmp), pio_double,      &
+                     dimids_tmp(1:fdims), tape(t)%hlist(fld)%sbuf_varid)
+             endif
+           end if
+         end do ! Loop over output patches
+       end do   ! Loop over fields
+       !
+       deallocate(mdimids)
+       ret = pio_enddef(tape(t)%Files(f))
```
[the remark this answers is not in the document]
peverwhee: added ret check and endrun call.

src/control/cam_history.F90
```diff
@@ -5076,56 +5221,57 @@ subroutine h_field_op (f, t)
...
+  subroutine dump_field (fld, t, f, restart)
     use cam_history_support, only: history_patch_t, dim_index_2d, dim_index_3d
     use cam_grid_support,    only: cam_grid_write_dist_array, cam_grid_dimensions
     use interp_mod,       only : write_interpolated
 
     ! Dummy arguments
-    integer,     intent(in)    :: f
+    integer,     intent(in)    :: fld
     integer,     intent(in)    :: t
+    integer,     intent(in)    :: f
```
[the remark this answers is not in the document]
peverwhee: comments added

src/control/cam_history.F90
peverwhee: removed and logic amended

src/control/cam_history.F90
peverwhee: I changed it to t since this is a loop nested within a loop over ptapes, but now I'm questioning how that will work with threading, so i've changed it back to ptapes

src/control/cam_history.F90
peverwhee: message added

src/control/cam_history.F90
peverwhee: simplified

src/control/cam_history.F90
brian-eaton: Another option (which gets my vote) is to remove dead code.  This function isn't referenced anywhere.  Same comment for `get_hist_restart_filepath` and `get_ptapes`.

src/control/cam_history.F90
gold2718: I do not think you answered my question. You are saying that the time that shows up in the file is the same as the time step active when the write occurs. I'm asking if the time that shows up in the file matches the time step that was active when the `outfld` buffer was captured for each and every field in the file. For instance, what if a particular diagnostic is only called every other time step and the write step is one where it was not captured.

src/control/cam_history.F90
gold2718: I seriously doubt `wshist` (if I got the right subroutine) can ever be threaded.

src/control/cam_history.F90
brian-eaton: I'll jump in to add that the reason `cell_methods` wasn't used for instantaneous data is that instantaneous is the default value.  To be consistent with the CF Metadata Conventions this should be added as `time: point`, not `time: instantaneous`.  CF uses `point` to indicate both time and space points.  I think it would be better to leave it out than to use a non-conforming value.

gold2718: I think `point` is a great term to use to replace `instantaneous`. Closer to the truth and CF compliant!

src/control/cam_history.F90
brian-eaton: Take 2.  Yes, the time variable always matches the time step of an instantaneous sample assuming that outfld was called during that time step.

You are correct to point out that outfld isn't called every time step for all fields that are output.  Radiation fields are a prime example. To analyze fields like that one needs to look at the details of when the outfield calls are made, and at what happens during the time step(s) when outfld is not called.  In the case of radiation, the heating rates that are applied during time steps when radiation isn't run (and outfld calls are not made) are from the most recent calculation (they are saved in the physics buffer).  So even though the outfld call for the instantaneous field was not made at the time indicated by the time coordinate, the field is in fact valid at that time.  I'm not claiming this to be the case for all fields that fall into this category.  But I do claim that all fields in this category are instantaneous, even if there is a mismatch in the associated time coordinate.

src/control/cam_history.F90
nusbaume: I am also happy if we use `time: point` as the metadata attribute to label the quantity as different from the accumulated files.  I just want _something_ there, because the actual file names are easily changed, and part of the reason we are doing this entire exercise is because the post-processing folks basically didn't like our metadata (so I don't want to give them more ammunition).

src/control/cam_history.F90
nusbaume: Ahh, I think I understand now!  I am fine then with either `ptapes` or `t` (and your original `t` method is very likely more performant, at least for the average CAM configuration).

nusbaume (approved):
Thanks for dealing with my concerns!  I am happy with this PR now.

src/control/cam_history.F90
```diff
@@ -5580,64 +5762,92 @@ subroutine wshist (rgnht_in)
...
+                cycle
+             end if
+             ! We have two files - one for accumulated and one for instantaneous fields
+             if (f == accumulated_file_index .and. .not. restart .and. .not. is_initfile(t) .and. .not. is_satfile(t)) then
+                ! accumulated tape - time is midpoint of time_bounds
+                ierr=pio_put_var (tape(t)%Files(f), tape(t)%timeid, (/start/),(/count1/),(/(tdata(1) + tdata(2)) / 2._r8/))
+             else
+                ! not an accumulated history tape - time is current time
+                ierr=pio_put_var (tape(t)%Files(f), tape(t)%timeid, (/start/),(/count1/),(/time/))
+             end if
```
brian-eaton, line 5776: When the output frequency is specified to be every time step, i.e., `nhtfrq==1`, this is instantaneous output.  But if instantaneous output has not been explicitly requested (for most variables the default output is averaged) then it should be recognized that this is a special case of instantaneous data and the metadata should be adjusted accordingly.  This code is instead outputting the time coordinate as the average value of the beginning and end of the time step.  That's not correct.  The data produced during a single time step should have the time coordinate from the end of the time step, which is what happens when instantaneous output is explicitly specified.  Here is a suggested change to the conditional.  Note that the condition `.not. is_satfile(t)` is already satisfied, so can just replace that by `nhtfrq(t) /= 1`

`
 if (f == accumulated_file_index .and. .not. restart .and. .not. is_initfile(t) .and. nhtfrq(t) /= 1) then
`

gold2718: I think I missed something. For an `'I'` file, what difference do you expect in the NetCDF file for `nhtfrq == 1` and `nhtfrq /= 1`?

brian-eaton: For an `'I'` file the time coordinate will always be at the end of the time step which is being written, regardless of `nhtfrq`.

The issue I'm addressing is when the averaging flag of a variable is not `'I'`.  Suppose `avgflag_pertape='A'` and suppose `nhtfrq=1`.  Then the output is instantaneous even though `avgflag='A'`.  The code I referenced is then setting the time coordinate to be at the midpoint of the time step being written which is not correct for instantaneous output.  The suggested modification is recognizing that `nhtfrq=1` is a special case of instantaneous output and setting the time coordinate accordingly.  Make sense?

peverwhee: ooh this is a excellent catch! I guess I similarly need to update the logic for other spots where we're conditionally writing accumulated vs instantaneous data

src/control/cam_history.F90
peverwhee: I am also happy with `time: point`

src/control/cam_history.F90
gold2718: Thanks; I understand although I think the time for this case should be set wherever the post-processing community wants it set (since, as you pointed out many many months ago, the time bounds are always the best way to know what is actually in the file).

I'm still not happy with the term, instantaneous, which is a concept from physics that is explicitly not measurable while CAM diagnostics files are all about what _is_ measurable. Of course, engineers abuse the term all the time so . . . :smirk:

brian-eaton: The post-processing community has never asked that the time coordinate of an instantaneous time sample be placed at the midpoint of the time step that produced it.  Also note that for instantaneous samples the time bounds associated with the time coordinate are not relevant.  That's why our current history files are CF compliant even though they mix instantaneous and non-instantaneous output.

I would argue that the term instantaneous in this discussion is a mathematical concept, not a physical one.  We're solving a discretized system of equations and the solutions are valid at points in time.

src/control/cam_history.F90
```diff
@@ -4381,412 +4417,495 @@ subroutine h_define (t, restart)
...
+             else
+               cell_methods = trim(cell_methods)//trim(tape(t)%hlist(fld)%field%cell_methods)
+             end if
+           end if
+           ! Time cell methods is after field method because time averaging is
+           ! applied later (just before output) than field method which is applied
+           ! before outfld call.
+           str = tape(t)%hlist(fld)%time_op
+           if (tape(t)%hlist(fld)%avgflag == 'I') then
+              str = 'instantaneous'
```
brian-eaton, line 4828: Another place to deal with the special case of `nhtfrq==1`:
```suggestion
           if (tape(t)%hlist(fld)%avgflag == 'I' .or. nhtfrq(t) == 1) then
              str = 'point'
```

src/control/cam_history.F90
brian-eaton: @peverwhee, In addition to the comment I just added for line 4827 of `cam_history.F90`, at line 2520 the `time_op` value should change from `' '` to `'point'`.  I couldn't figure out how to make this comment at that point in the code since there are no changes there.

src/control/cam_history.F90
```diff
@@ -778,10 +798,14 @@ subroutine history_readnl(nlfile)
...
-            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t.%y-%m.nc'
+            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t%f.%y-%m.nc'
           else
-            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t.%y-%m-%d-%s.nc'
+            hfilename_spec(t) = '%c.cam' // trim(inst_suffix) // '.h%t%f.%y-%m-%d-%s.nc'
           end if
+        else
+           ! Append file type - instantaneous or accumulated - to filename
+           ! specifier provided
+           hfilename_spec(t) = trim(hfilename_spec(t)) // '%f'
```
brian-eaton, line 808: This code doesn't recognize that the `.nc` file extension is part of the template.
```suggestion
           ! Append file type - instantaneous or accumulated - to filename
           ! specifier provided (in front of the .nc extension).
           tlen = len_trim(hfilename_spec(t))
           hfilename_spec(t) = hfilename_spec(t)(:tlen-3)  // '%f.nc'
```
and provide the declaration for `integer :: tlen`.

src/control/cam_history.F90
gold2718: > I would argue that the term instantaneous in this discussion is a mathematical concept, not a physical one. We're solving a discretized system of equations and the solutions are valid at points in time.

Oh no, math plus engineering vs. physics! That's like rock and paper ganging up on scissors :grimacing:
Good thing we are only doing math and not trying to describe physical reality :smirk:

src/control/cam_history.F90
peverwhee: changed back to t

src/control/cam_history.F90
peverwhee: changed!

src/control/cam_history.F90
peverwhee: logic updated to treat nhtfrq == 1 as instantaneous; time_op value changed to 'point'

src/control/cam_history.F90
peverwhee: changed!

src/control/cam_history.F90
peverwhee: good catch - code updated

src/control/cam_history.F90
brian-eaton: I'm afraid commit 7a0e5282 is a regression.  My testing revealed that if the `h0` stream contains averaged and instantaneous fields (as declared in their `avgflag` values) and `nhtfrq==1`, only the `h0i` file is produced and it only contains the fields that have `avgflag='I'`.  The fields with `avgflag='A'` are not output, even though they should be output with instantaneous values.

The changes I suggested (and tested) didn't attempt to change the `avgflag` settings for the case `nhtfrq==1`.  They only make sure the file contains the metadata and time coordinates which are correct for instantaneous fields even though they appear in an `hXa` file due to their `avgflag` values.

If we want the `nhtfrq==1` case to have filenames consistent with our new convention for instantaneous data, I think the best approach is to modify the field `avgflag` setting to be `'I'`.  This would eliminate the need to add an `'.or. nhtfrq(t)==1'` condition in other code locations.  I suggest the following code be added at line 792 in `cam_history.F90`:
```
      ! If nhtfrq == 1, then the output is instantaneous.  Enforce this by setting
      ! the per-file averaging flag.

      do t = 1, ptapes
         if (nhtfrq(t) == 1) then
            avgflag_pertape(t) = 'I'
         end if
      end do
```

src/utils/cam_grid_support.F90
```diff
@@ -156,7 +161,7 @@ module cam_grid_support
     type(horiz_coord_t), pointer       :: lon_coord => NULL() ! Longitude coord
     logical                            :: unstructured  ! Is this needed?
     logical                            :: block_indexed ! .false. for lon/lat
-    logical                            :: attrs_defined = .false.
+    logical                            :: attrs_defined(2) = .false.
```
brian-eaton, line 164: Should this `2` be `maxsplitfiles`?  If not a parameter should be used that indicates what the array length should be.

src/control/cam_history.F90
```diff
@@ -2597,11 +2651,11 @@ subroutine fldlst ()
       ! are also present in the fincl array.
 
       ! The first empty slot in the current fincl array is index f from loop above.
```
brian-eaton, line 2653: `index f` should be `index fld`

src/control/cam_history.F90
```diff
@@ -4235,34 +4302,50 @@ subroutine h_define (t, restart)
       if(masterproc) write(iulog,*)'Opening netcdf history restart file ', trim(hrestpath(t))
     else
       tape => history_tape
-      if(masterproc) write(iulog,*)'Opening netcdf history file ', trim(nhfil(t))
+      if(masterproc) then
+         if (hfile_accum(t) .and. nhtfrq(t) /= 1 .and. hfile_inst(t)) then
+            write(iulog,*)'Opening netcdf history files ', trim(nhfil(t,accumulated_file_index)), &
+                  trim(nhfil(t,instantaneous_file_index))
```
brian-eaton, line 4308: Add space between filenames
```suggestion
            write(iulog,*)'Opening netcdf history files ', trim(nhfil(t,accumulated_file_index)), &
                  '  ', trim(nhfil(t,instantaneous_file_index))
```

src/control/cam_history.F90
peverwhee: Updated to enforce instantaneous output for nhtfrq = 1. Also added code at line ~486 to override any flags specified in fincl to 'I' when nhtfrq = 1

src/control/cam_history.F90
peverwhee: good call - added

src/control/cam_history.F90
peverwhee: updated!

cacraigucar (approved):
I left my one conversation unresolved as I'm not sure if the side discussion was finished or not.  But my concern was addressed to my satisfaction

brian-eaton (approved):
Looks good.  Nice work!

ekluzek:
@peverwhee nice to see this in. We are going to study it in CTSM and figure out what we need to do using this as strong inspiration. Thanks so much.

billsacks:
Thank you very much for your work on this, @peverwhee ! It's exciting to see this completed!!!

PR #908, opened 2023-10-25
cam6_3_137: Update mpas external to use v8.0.1

### LINKED ISSUE #861: Update MPAS-A within CAM to v8.0.1
### Issue Type

Externals Update

### Issue Description

The current version of MPAS-A used in CAM points to a specific commit from sometime after the v7.3 tag was made but before the v8 release was made. Since MPAS-A v8.0.1 has released, I propose to update the Externals_CAM.cfg and supporting infrastructure to use this tag.

### Will this change answers?

I Don't Know

Yes

gdicker1:
Changes include updating the mpas external hash to a few commits after the 'v8.0.1' tag, updating the mpas/Makefile, adding the mpas_cam_damping_levels namelist option, updating some subroutine calls to use new arguments, and setting up the halo_groups MPAS uses.

Closes #861


gdicker1:
Tests involving `mpasa*` resolutions will have namelist differences due to adding the mpas_cam_damping_levels option.

Externals_CAM.cfg
```diff
@@ -73,7 +73,7 @@ local_path = src/dynamics/mpas/dycore
 protocol = git
 repo_url = https://github.com/MPAS-Dev/MPAS-Model.git
 sparse = ../.mpas_sparse_checkout
-hash = ff76a231
+hash = b8c33daa
```
gdicker1, line 76: Link to [b8c33daa](https://github.com/MPAS-Dev/MPAS-Model/commit/b8c33daa4b124f116b0ba1b1386968c4c15b42e4) on MPAS-Dev/MPAS-Model:develop branch. It's about 6 PRs after the 'v8.0.1' tag was created. I'll update this hash to a tag as the CESM3 release approaches.

bld/namelist_files/namelist_definition.xml
```diff
@@ -8416,10 +8416,20 @@ A value of 1.0 will result in layered diffusion similar to CAM-SE and gives damp
 coefficients of 0.2216E7, 0.6482E6, 0.1927E6 in the top-most three layers on the dynamics
 variables u, w, and theta. The top 3 damping coefficients scale linearly with
 mpas_cam_coef. 0.0 disables SE like 2nd-order diffusion in the absorbing layer and is the
```
jtruesdal, line 8418: With the addition of mpas_cam_damping_levels I think the comment here should reflect that the mpas_cam_coef and mpas_cam_damping_levels work together to determine the scaling and number of levels over which the 2nd-order horizontal diffusion is applied.  Additionally for the sentence speaking about diffusion coeff similar to CAM-SE you would also have to specify the number of mpas_cam_damping_levels needed for this case which I am assuming is 3 not the default of 4.  Also it seems the coefficients also depend on  the type of horizontal mixing chosen (2d_fixed or 2d_smagorinsky).  If that is the case the comment might be better stated as: To provide a set of coefficients giving a layered diffusion similar to CAM-SE a user would configure for xxx horizontal mixing as well as setting mpas_cam_damping_levels to 3 and using mpas_cam_coeff of 1.0

doc/ChangeLog
```diff
@@ -1,5 +1,93 @@
...
+Github PR URL: https://https://github.com/ESCOMP/CAM/pull/908
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+ - Update MPAS-A within CAM to v8.0.1: https://github.com/ESCOMP/CAM/issues/861
+
+Describe any changes made to build system: N/A
+
+Describe any changes made to the namelist: N/A
+
+List any changes to the defaults for the boundary datasets: N/A
```
jtruesdal, line 16: I know this is just a template but please be sure to fill in the new namelist variable and newly added files.  I was also expecting the update from V7 to V8 to be answer changing.  If that is  not the case it would make this update easier. (ie no need to validate climate etc.)

bld/namelist_files/namelist_definition.xml
```diff
@@ -8416,10 +8416,20 @@ A value of 1.0 will result in layered diffusion similar to CAM-SE and gives damp
...
 </entry>
 
+<entry id="mpas_cam_damping_levels" type="integer" category="mpas"
+       group="damping" valid_values="">
+Number layers from the top in which to apply cam 2nd-order horizontal
+diffusion. Viscocity linearly ramps to zero by layer number from the top.
+mpas_cam_damping_levels and mpas_cam_coef must both be greater than 0 for the
+diffusion to be enabled. 
+Default: 4
+</entry>
```
jtruesdal, line 8431: Can you say something about this default of 4 levels in the comment.  It seems reasonable and given the higher number of levels that CAM is operating at necessary for stability.

jtruesdal (changes requested):
This looks good and if it is really bfb then not the major update I was thinking it was.  Just a few small requests and it should be good to go to full review.

doc/ChangeLog
gdicker1: Thanks for the reminder. Let me know if anything else needs to change after [010f084d](https://github.com/ESCOMP/CAM/pull/908/commits/010f084d9d2d5b0bef73322413d58ce9e8d982b3) (besides filling in date, reviewers, etc)

bld/namelist_files/namelist_definition.xml
gdicker1: I'm not sure how to resolve this. I'm partial to just deleting the mention of the specific coefficients since I don't think we can get the same results without setting `mpas_cam_coef` outside its 0.0 to 1.0 range.

What do you think of this description for `mpas_cam_coef`?
```
Coefficient for scaling the 2nd-order horizontal diffusion in the mpas_cam absorbing layer.
The number of layers this diffusion is applied to is controlled by mpas_cam_damping_levels.
0.0 disables this 2nd-order diffusion in the absorbing layer and is the current default.
Setting mpas_cam_damping_levels to 0 also disables this diffusion. Sponge layer absorption
can also be provided by Rayleigh damping.
```

bld/namelist_files/namelist_definition.xml
gdicker1: I wouldn't know what to say about it. 4 is the default value that was given in [MPAS-Model PR #1040](https://github.com/MPAS-Dev/MPAS-Model/pull/1040) when this change was introduced.

bld/namelist_files/namelist_definition.xml
```diff
@@ -8416,10 +8416,20 @@ A value of 1.0 will result in layered diffusion similar to CAM-SE and gives damp
...
+Sponge layer absorption can also be provided by Rayleigh damping.
 Default: 0.0
 </entry>
 
+<entry id="mpas_cam_damping_levels" type="integer" category="mpas"
+       group="damping" valid_values="">
+Number layers from the top in which to apply cam 2nd-order horizontal
+diffusion. Viscocity linearly ramps to zero by layer number from the top.
+mpas_cam_damping_levels and mpas_cam_coef must both be greater than 0 for the
+diffusion to be enabled.
```
cacraigucar, line 8429: Should there be a check that if one is non-zero and the other is zero, that an error occurs and namelist generation is halted?  i.e. should both of these variables match in being either zero, or non-zero and if they are not, then an error occurs?

doc/ChangeLog
```diff
@@ -1,5 +1,96 @@
...
+  DAE_Vnuopc.f45_f45_mg37.FHS94.izumi_nag.cam-dae (Overall: FAIL) details:
+    FAIL DAE_Vnuopc.f45_f45_mg37.FHS94.izumi_nag.cam-dae RUN time=10
+ - pre-existing failure
+
+  ERC_D_Ln9_Vnuopc.mpasa480z32_mpasa480.FHS94.izumi_nag.cam-outfrq3s_usecase (Overall: NLFAIL) details:
+    FAIL ERC_D_Ln9_Vnuopc.mpasa480z32_mpasa480.FHS94.izumi_nag.cam-outfrq3s_usecase NLCOMP
+      found extra variable: 'mpas_cam_damping_levels'
+ - expected failure, added option mpas_cam_damping_levels to default namelist file
+
+izumi/gnu/aux_cam: All PASS
```
cacraigucar, line 87: This section needs to be filled out when the official CAM tagging regression tests are run and reflects those actual test results.

cacraigucar (changes requested):

bld/namelist_files/namelist_definition.xml
gdicker1: It's not something we check in MPAS-A. A warning would probably be good, but I don't think it should halt things. I consider these "high-level" options, maybe my perspective is off on that.

bld/namelist_files/namelist_definition.xml
PeterHjortLauritzen: In the current version of MPAS the damping is hardcoded to the top 3 levels:

```
            do iCell = cellStart,cellEnd                                                                                                                                      
               !                                                                                                                                                              
               ! 2nd-order filter for top absorbing layer as in CAM-SE :  WCS January 2021                                                                                    
               ! CAM6 SE damping layer coefficients from Peter Lauritzen with config_mpas_cam_coef = 1.0                                                                      
               !                                                                                                                                                              
               kdiff(nVertLevels,iCell) = max(kdiff(nVertLevels,iCell), 18.333*config_len_disp*config_mpas_cam_coef)                                                          
               kdiff(nVertLevels-1,iCell) = max(kdiff(nVertLevels-1,iCell), 5.4*config_len_disp*config_mpas_cam_coef)                                                         
               kdiff(nVertLevels-2,iCell) = max(kdiff(nVertLevels-2,iCell), 1.6*config_len_disp*config_mpas_cam_coef)                                                         
            end do  
```

and has now changed to a more general formulation

```
               do k = nVertLevels-config_number_cam_damping_levels + 1, nVertLevels
                  visc2cam = 4.0*2.0833*config_len_disp*config_mpas_cam_coef
                  visc2cam = visc2cam*(1.0-real(nVertLevels-k)/real(config_number_cam_damping_levels))
                  kdiff(k  ,iCell) = max(kdiff(nVertLevels  ,iCell),visc2cam)
               end do
```

I would suggest changing the text to:

```
Coefficient for scaling the 2nd-order horizontal diffusion in the mpas_cam absorbing layer. The absorbing layer depth is controlled with mpas_cam_damping_levels. The damping coefficients scale linearly with
mpas_cam_coef. A value of 0.0 (or mpas_cam_damping_levels=0) disables the 2nd-order diffusion in the absorbing layer. Sponge layer absorption can also be provided by Rayleigh damping.

E.g. a value of 1.0 with mpas_cam_damping_levels=3 will result in damping coefficients of 2E6 m^2/s, 6E5, 2E5 in the top-most three layers on the dynamics variables u, w, and theta. 
Default: 0.0
```

FYI: In CAM-SE the damping coefficients (at the time of writing) use coefficients 6E5, 2E5, 1E5, 4E4, 2E4 m^2/s since we changed to a ```tanh``` formulation (from CAM-FV).

The mpas_horiz_mixin (options 2d_fixed or 2d_smagorinsky) are different from the sponge layer damping. These are applied throughout the MPAS domain.

PeterHjortLauritzen: Regarding my question on regression test failures: since mpas_cam_coef=0 by default this code is not invoked. That is probably why you were not seeing any differences. If we introduce or do have an AMIP regression test then I believe we should set 

mpas_cam_coef=1.0
mpas_cam_damping_levels=3

bld/namelist_files/namelist_definition.xml
PeterHjortLauritzen: If we should match the current MPAS code then mpas_cam_damping_levels default value should be 3. (again since mpas_cam_coef=0.0 this does not really matter) ...

This implementation was simply mimicking what was used in CAM-SE a while ago so I think there is not much to say here ...

PeterHjortLauritzen (approved):
Looks good to me. 

Had one suggestion for changing the description of mpas_cam_coef and changing the default for mpas_cam_damping_levels to 3.

bld/namelist_files/namelist_definition.xml
gdicker1: This is resolved. After some discussion outside this PR, it was decided to turn off the damping by default and make `mpas_cam_coef` and `mpas_cam_damping_levels` both consistent at zero.

bld/namelist_files/namelist_definition.xml
gdicker1: After discussing this outside the PR, I don't think this change is needed. There will likely be a future PR that deals more with this damping, testing of it, and will address inconsistent settings of `mpas_cam_coef` and `mpas_cam_damping_levels` within CAM's `build-namelist` code.

bld/namelist_files/namelist_definition.xml
gdicker1: The root of this (the description in namelist_definition.xml) is resolved by [commit 809e76f](https://github.com/ESCOMP/CAM/pull/908/commits/809e76f274ccf340cbe86aa7b8e3bc1f6c096c91) in this branch. I used Peter's suggested description.

Hoping to resolve the testing question soon.

bld/namelist_files/namelist_definition.xml
cacraigucar: @gdicker1 - Could you open an issue that contains this proposed change so that it doesn't get lost?

cacraigucar (approved):

bld/namelist_files/namelist_definition.xml
gdicker1: Testing question was also resolved: it was decided to add turn on this damping in the F2000climo aux_cam test using the mpasa480_mpasa480 grid. See commit [268f40b3](https://github.com/ESCOMP/CAM/pull/908/commits/268f40b350d44ea3ba550b8217be633d00df55ed)

cacraigucar:
It is quite possible that you haven't pushed back your ChangeLog updates yet.  I did want to give you a reminder to add an entry for test/system/archive_baseline.sh

PR #929, opened 2023-11-27
cam6_3_141: Improve handling of 0th / initial time in history output

### LINKED ISSUE #459: Improve handling of 0th / initial time in history output
From @phillips-ad

> An instantaneous time step is included for subdaily variables at the beginning of hybrid runs in CAM, even when all output is averaged. This is inconsistent, and is another hoop to jump through when combining runs across scenarios (historical->ssp370 for instance).

> The 0th time step is a complication that doesn't need to be there in my mind, but I"d be curious as to if there is a logical reason for its presence. If someone were to concatenate the CESM2-LENS historical and ssp370 runs together (as is commonly done), the first timestep of the ssp370 scenario of a CAM daily output file is that extra 0th instantaneous timestep, which I would guess is a surprise to most folks. You can see this by examining/converting the time_bnds array for the following file (on CGD systems): 
/project/mojave/cesm2/LENS/atm/day_1/PSL/b.e21.BSSP370smbb.f09_g17.LE2-1301.018.cam.h1.PSL.20150101-20241231.nc

I know about this issue mainly in the context of CTSM (see https://github.com/ESCOMP/CTSM/issues/925 ): for example, if you set a run to start on Jan 1, 2010, CTSM will run an initial, 0th time step from 23:30 - 00:00 on Dec 31, 2009 before running the time step at the intended model start date. For CTSM, it's unclear to us if we really need to run this 0th time step at all, or if that is just a historical artifact that can be removed. Does anyone here know the answer to that? But even if we maintain running the 0th time step, the inclusion of it in history files causes off-by-one issues that are confusing to users, so I feel we should separate this 0th time step from the normal history output. From @phillips-ad 's comments above, it sounds like there are similar confusions for CAM output.

I can't tell how much overlap this issue has with #361

brian-eaton:
Issue #459 - Improve handling of 0th / initial time in history output

-  Add namelist variable, write_nstep0, to control output of nstep==0 time sample.  This output is now turned OFF by  default.

resolves #459


cacraigucar (approved):

peverwhee (approved):
looks good to me!

billsacks:
Thank you for your work on this, @brian-eaton ! It's exciting to see this coming in!!

PR #933, opened 2023-11-30
part of cam_cesm2_2_rel_09: remove cism from all f compsets on CESM2_2 branch

jedwards4b:
Also adds a pelayout for <grid name="a%ne0np4CONUS" > on derecho.

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -1460,7 +1460,7 @@
   </grid>
 
   <grid name="a%ne0np4CONUS" >
-    <mach name="any" >
+    <mach name="cheyenne" >
       <pes pesize="any" compset="any">
         <comment>none</comment>
 	      <ntasks>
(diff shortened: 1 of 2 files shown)


cime_config/config_pes.xml
```diff
@@ -1460,7 +1460,7 @@
   </grid>
 
   <grid name="a%ne0np4CONUS" >
-    <mach name="any" >
+    <mach name="cheyenne" >
```
adamrher, line 1463: Derecho?

adamrher (approved):
approve. just one comment double checking the machine

adamrher:
@jedwards4b can you remind me the derecho issue with GLC component in the cesm2.2 tag? I'm fine with this soln of removing the GLC component from F-cases (this isn't climate changing), but @Katetc wanted to understand further and wondered if its related to pe layout issues she has since resolved?

jedwards4b:
It was because cism only supports a NO_LEAP calendar and WACCM needs a GREGORIAN.   CISM doesn't really do anything in cesm2.2 so the easiest solution was to simply remove it.

adamrher:
Thanks. There will be answer changes in switching from CISM2%NOEVOLVE to SGLC, as the elevation classes are defined differently in CLM, but both methods are scientifically valid. @Katetc did Jim answer your question?

Katetc:
Yes, that was helpful. I agree that CISM doesn't exactly "do" anything here, but it will change 2.2 answers for several non-WACCM compsets. Maybe the prudent thing to do would be only change to SGLC in WACCM compsets? I don't want to add work for Cheryl here, and I'm not super invested in keeping CISM in F compsets, I just agreed with Adam that we may not want to make too many answer changing updates for what is essentially just a machine updating release.

fvitt:
I am not so sure there was a need to remove CISM from F compsets.  The GREGORIAN calendar is needed in WACCMX and SD compsets.  Those compsets did not previously include CISM.

PR #992, opened 2024-03-13
cam6_3_158: ZM clean up round 2 for CAM

### LINKED ISSUE #965: change zm directory to be zhang_mcfarlane

@swrneale requested that zm be renamed zhang_mcfarlane.

_No response_

No

Yes

### LINKED ISSUE #978: Reimplement writing within ZM and remove pflx variable
### Issue Type

Other (please describe below)

### Issue Description

When CCPP'izing ZM, it was unclear how to handle internal writes.  We will now pass in iulog and masterproc so that writes to iulog can be restored.

Also we can remove the `pflx` variable from `tphysbc`, as it doesn't appear to be used anywhere.

### Will this change answers?

No

Yes

### LINKED ISSUE #984: ZM cleanup

Separate all ZM variables in zm_conv_momtran.F90 which have ncnst dimension (which is actually 2 for north and east winds) into two separate variables

Make jctop and jcbot be integer values since they are indices

@swrneale @nusbaume 

No

Yes

### LINKED ISSUE #1004: GEOS-Chem compsets will fail due to bugs in CAM and CIME
### What happened?

I updated to the latest cam_development to test my draft https://github.com/ESCOMP/CAM/pull/996 which updates GEOS-Chem to version 14.3.1. In testing I discovered two problems getting the GEOS-Chem configuration files to the case run directory.

1. There was a change in the CIME code to call `cam.case_setup.py` which causes the files never to be copied from the source code. I created a PR to fix this. See https://github.com/ESMCI/cime/pull/4604.
2. There was a change in `buildnml` code to copy the GEOS-Chem configuration files from the case directory to the run directory. A typo prevents the files from ever getting copied. The change happened in commit https://github.com/ESCOMP/CAM/commit/2a08713a6c82a4babbf0390ae48742fc9d869c7d. A new conditional that checks if the species database yaml file is present was added. However, there is a typo in the extension name of the filename string. I left a comment on the commit page showing where this is. It should be "species_database.yml"

Create a GEOS-Chem compset case using `cam_development` branch. Upon case setup there should be several GEOS-Chem configuration files in the case directory. Due to bug #1 they are missing. Merging in my PR to cime will fix this. After setting up the case with the fix you will see configuration files for GEOS-Chem in the case, such as `geoschem_config.yml`. Building and submitting should result in copying these files to the run directory. Due to bug #2 they will not be there. Fixing the typo in `buildnml` will fix this and resubmitting will copy them over.

cam6_3_150 (plus my GEOS-Chem updates in PR linked to above)

CISL machine (e.g. cheyenne)

Intel

/glade/u/home/elundgren/code.cesm_gc_14.3/components/cam/cime/scripts/case.FCHIST_GC_14.3

Any CAM SE can do this

### Extra info

_No response_

### LINKED ISSUE #1012: Tag cam6_3_157 missing updated .gitignore
### Issue Type

Code Clean-up

### Issue Description

cam6_3_157 is missing an updated .gitignore file. That tag included changes to move the FV3 cam interface to an external that  requires an update to .gitignore.  Without the update the new FV3 interface directory is reported as an untracked file after the interface code is pulled in via checkout_externals.  This error doesn't affect the running of CAM or its answers but makes development work with that checked out code a little more cumbersome as the developer will need to clear these untracked files before checking out a different tag with the local repository.  Below is the git diff of the one line mod

diff --git a/.gitignore b/.gitignore
index cafb205f..e24b6fa5 100644
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ src/physics/pumas
 src/physics/pumas-frozen
 src/physics/rrtmgp/data
 src/physics/rrtmgp/ext
-src/dynamics/fv3/atmos_cubed_sphere
+src/dynamics/fv3
 libraries/FMS
 libraries/mct
 libraries/parallelio

### Will this change answers?

No

Any CAM SE can do this

cacraigucar:
Various additional cleanup for ZM moving to CCPP and atmospheric_physics repo

closes #965 
closes #978 
closes #984
closes #1012 
closes #1004


src/physics/cam/zm_conv_intr.F90
```diff
@@ -607,6 +608,13 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
    lengath = count(ideep > 0)
    if (lengath > ncol) lengath = ncol  ! should not happen, but force it to not be larger than ncol for safety sake
 
+   jctop(:) = pver
+   jcbot(:) = 1
+   do i = 1,lengath
+      jctop(ideep(i)) = jt(i)
+      jcbot(ideep(i)) = maxg(i)
+   end do
```
nusbaume, line 616: Would it be worth making the data type conversion here explicit?  For example:

```suggestion
   jctop(:) = real(pver, r8)
   jcbot(:) = 1._r8
   do i = 1,lengath
      jctop(ideep(i)) = real(jt(i), r8)
      jcbot(ideep(i)) = real(maxg(i), r8)
   end do
```

src/physics/cam/zm_conv_intr.F90
```diff
@@ -746,33 +754,34 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
...
 
-     ptend_loc%u(:ncol,:pver) = wind_tends(:ncol,:pver,1)
-     ptend_loc%v(:ncol,:pver) = wind_tends(:ncol,:pver,2)
      ptend_loc%s(:ncol,:pver) = seten(:ncol,:pver)
 
      call physics_ptend_sum(ptend_loc,ptend_all, ncol)
 
+     ! Output ptend variables before they are set to zero with physics_update
+     call outfld('ZMMTU', ptend_loc%u(1,1), pcols, lchnk)
+     call outfld('ZMMTV', ptend_loc%v(1,1), pcols, lchnk)
```
nusbaume, line 783: Is this `outfld` call correct?  I understand that you just copied it from below, but isn't this, and the other wind-based `outfld` calls below with a dimension of `(1,1)`, grabbing just a single grid cell?

nusbaume (commented):
Looks good, but did have a couple questions that I hope to get answered before approval.  Thanks!

src/physics/cam/zm_conv_intr.F90
gold2718: Ack! This is ancient usage and the `(1,1)` should be removed. This call 'works' because of the structure of `outfld` which uses the third argument to determine the shape of the incoming array but it should definitely be cleaned up so a 'better' `outfld` (someday) will work correctly.

swrneale: There's a few more than just these outflds in ZM where it happens also.

cacraigucar: Fixed in ZM

src/physics/cam/zm_conv_intr.F90
cacraigucar: Added explicit type conversions

nusbaume (approved):
Thanks for dealing with my questions!  Everything looks good to me now.

Issue #991, opened 2024-03-11
Remove gustiness term from U10/add new variable with gustiness


megandevlan:
### What is the feature/what would you like to discuss?

Following the addition of gustiness in CMEPS, we modified CAM to also output UGUST, but U10 was also changed - it became not just the large-scale resolved wind, but the resolved wind _plus_ the gustiness term. We propose that U10 should go back to being the large scale resolved wind, and a new variable be added to output the sum of resolved+gusts ("U10WITHGUSTS"). 

The necessary mods in CMEPS are very briefly described here https://github.com/ESCOMP/CMEPS/issues/438

I'm testing out the modification now and will open a PR soon. 

### Will this change (regression test) answers?

I Don't Know

### Will you be implementing this enhancement yourself?

Yes

cacraigucar:
Closing - in cam6_3_154

PR #994, opened 2024-03-14
cam6_3_154: Update U10 to be resolved wind; add variable for U10+gusts

megandevlan:
Updates to U10 are made in CMEPS so that it's only the large-scale resolved U10 again (as before gustiness was added). This particular PR then adds an option to output a new variable (U10WITHGUSTS) that is the merged sum of large-scale and gustiness contributions to U10. 

Closes: https://github.com/ESCOMP/CAM/issues/991 

Also needs CMEPS PR https://github.com/ESCOMP/CMEPS/pull/440 to send the new variable from CMEPS to CAM.


cacraigucar:
@cecilehannay @adamrher @PeterHjortLauritzen 

Is this change needed for the CESM beta tag that Cecile is waiting for?

PR #998, opened 2024-03-18
add pointers in cslam for intel performance (and update cime external?) for cam_cesm2_2_rel branch

### LINKED ISSUE #955: Update CSLAM Derecho performance issues in 2.2 release tag
### Issue Type

Other (please describe below)

### Issue Description

Derecho performance issues for CSLAM were resolved on the trunk https://github.com/ESCOMP/CAM/pull/845? but not on the 2.2 release tag. This is needed for running CSLAM with the 2.2 release tag.

There is also an issue with i/o via the mpich compiler, which will be resolved with a new 2.2 cime tag @fvitt. Ideally that tag wold be made first and would update the externals in this issue as well.

### Will this change answers?

I Don't Know

Any CAM SE can do this

adamrher:
Fixes #955 

For the new cesm2.2 tag, we will need to update the cime external as in https://github.com/ESMCI/cime/pull/4559 i/o performance issues with VR grids. Should we update the external in the cam tag?

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -9,8 +9,7 @@ local_path = components/cice
 required = True
 
 [cime]
-#tag = cime5.8.32.7
-tag = maint-5.8_5.8.32
+tag = 5.8.32.10
 protocol = git
 repo_url = https://github.com/ESMCI/cime
 local_path = cime
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -9,7 +9,8 @@ local_path = components/cice
 required = True
 
 [cime]
-tag = cime5.8.32.7
+#tag = cime5.8.32.7
+tag = maint-5.8_5.8.32
 protocol = git
 repo_url = https://github.com/ESMCI/cime
 local_path = cime
--- a/src/dynamics/se/dycore/fvm_consistent_se_cslam.F90
+++ b/src/dynamics/se/dycore/fvm_consistent_se_cslam.F90
@@ -44,7 +44,7 @@ contains
     use thread_mod            , only: vert_num_threads, omp_set_nested
     implicit none
     type (element_t)      , intent(inout) :: elem(:)
-    type (fvm_struct)     , intent(inout) :: fvm(:)
+    type (fvm_struct), target     , intent(inout) :: fvm(:)
     type (hybrid_t)       , intent(in)    :: hybrid   ! distributed parallel structure (shared)
     type (TimeLevel_t)    , intent(in)    :: tl              ! time level struct
     type (hvcoord_t)      , intent(in)    :: hvcoord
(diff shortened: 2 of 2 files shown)


adamrher:
Updated cime exterrnal and ran a cslam test case to completion. The cime tag has a weird name ```maint-5.8_5.8.32``` but I think it's the correct tag. Provided it is, I think regressions testing can start.

Externals.cfg
```diff
@@ -9,7 +9,8 @@ local_path = components/cice
 required = True
 
 [cime]
-tag = cime5.8.32.7
+#tag = cime5.8.32.7
+tag = maint-5.8_5.8.32
```
cacraigucar, line 13: I have requested a tag to be made for the head of this branch.  When it is made, we will use that tag.

src/dynamics/se/dycore/fvm_consistent_se_cslam.F90
```diff
@@ -152,22 +154,26 @@ subroutine run_consistent_se_cslam(elem,fvm,hybrid,dt_fvm,tl,nets,nete,hvcoord,&
...
              fvm(ie)%vertex_recons_weights,fvm(ie)%vtx_cart,&
              irecons_tracer_lev(k))
-        !call t_stopf('fvm:tracers_reconstruct')
-        !call t_startf('fvm:swept_flux')
-        call swept_flux(elem(ie),fvm(ie),k,ctracer,irecons_tracer_lev(k),gsweights,gspts)
-        !call t_stopf('fvm:swept_flux')
+         !call t_stopf('FVM:tracers_reconstruct')
+         !call t_startf('fvm:swept_flux')
+         call swept_flux(elem(ie),fvm(ie),k,ctracer,irecons_tracer_lev(k),gsweights,gspts)
+         !call t_stopf('fvm:swept_flux')
```
cacraigucar, line 176: I know you didn't make this change, but I'm wondering why these timers were commented out.  I would suggest it makes sense to uncomment them, so they are available if folks wanted to time these routines.

cacraigucar (changes requested):
If I have access to your fork, and you are okay with it, I can put in the updated tag when it is made.  I will also need to make a ChangeLog for this commit, and can add it when it is ready.

Externals.cfg
cacraigucar: The tag has been made and it is:  5.8.32.10

cacraigucar (commented):
Still need to squash bugs, but the code looks fine

cacraigucar (approved):

cacraigucar:
It was decided to not continue supporting the cesm2_2 branch, so this PR is being closed without being implemented.

PR #1003, opened 2024-03-20
Fix for multiple monthly average history files.

### LINKED ISSUE #1000: Output of more than 1 monthly average history file is broken.
### What happened?

If monthly average files are requested after the h0 file, the first write occurs at the end of the first day rather than the end of the month.

Any compset will exhibit this behavior.

cam6_3_140 and later

Other (please explain below)

Other (please specify below)

_No response_

Yes

### Extra info

The problem does not depend on machine or compiler.

The bug was introduced with the commit to break the history output into separate files for accumulated and instantaneous output.

brian-eaton:
resolves #1000

### DIFF
--- a/src/control/cam_history.F90
+++ b/src/control/cam_history.F90
@@ -5516,6 +5516,7 @@ end subroutine print_active_fldlst
 #endif
 
     integer :: yr, mon, day      ! year, month, and day components of a date
+    integer :: yr_mid, mon_mid, day_mid ! year, month, and day components of midpoint date
     integer :: nstep             ! current timestep number
     integer :: ncdate(maxsplitfiles) ! current (or midpoint) date in integer format [yyyymmdd]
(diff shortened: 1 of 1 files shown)


cacraigucar (approved):

peverwhee (approved):
thanks for finding this, @brian-eaton !

I'll be interested to see if any regression tests aren't b4b (indicating that I missed an answer change). I'm confused as to why the other accumulated files were "working" (but perhaps they weren't...)

brian-eaton:
> thanks for finding this, @brian-eaton !
> 
> I'll be interested to see if any regression tests aren't b4b (indicating that I missed an answer change). I'm confused as to why the other accumulated files were "working" (but perhaps they weren't...)

The regression tests are b4b.  We don't have any tests that include output to multiple monthly output files.  The bug is in code that is specific to checking whether it's time to write history buffers to a monthly output file, i.e., `nhtfrq==0`.  No other output frequencies were impacted.

Issue #1015, opened 2024-04-26
SILHS subcolumns output as all zeros


bstephens82:
### What happened?

When running SILHS with either -phys cam6 or -phys cam_dev (although for cam_dev see the necessary code changes associated with issue #844), SILHS subcolumn fields are being output to netcdf as all zeros, despite being nonzero right before the outfld call and even within the outfld call.  I have also tried setting the entire SILHS subcolumn field `WM_lh_out = 1.0_r8` immediately before the [outfld call in subcol_SILHS.F90](https://github.com/ESCOMP/CAM/blob/776400d2a56d1c95551c9782810e4120ecb4f111/src/physics/cam/subcol_SILHS.F90#L1631), just to make sure the outfld call is not receiving anything crazy, and it still outputs all zeros in the netcdf history file.

### What are the steps to reproduce the bug?

- Using a recent tag (I've tested up to cam6_3_155), set up a CAM test case with "-silhs -psubcols 4" included in CAM_CONFIG_OPTS.  You can use either -phys cam6 or -phys cam_dev (although cam_dev will require some code changes to cam_dev/micro_pumas_cam.F90 associated with separate issue #844 to work).  Ask @bstephens82 for the tentative code changes (or see PR #1016).
- Then add SILHS output fields to fincl1 in user_nl_cam, for example fincl1 = 'SILHS_WM_SCOL' or 'SILHS_THLM_SCOL'.  
- You will also need to add `use_gw_convect_dp=.false.` to your namelist since turning on SILHS will automatically set `deep_scheme='off'` and for some reason that won't run with the `gw` flag set to T.  This seems like a separate bug to me, but in any case this is necessary when `deep_scheme='off'`.
- I believe you also need to SourceMod in cam/subcol_SILHS.F90 where you change `l_outfld_subcol=.true.` (this is [hard-coded to false by default in that module](https://github.com/ESCOMP/CAM/blob/776400d2a56d1c95551c9782810e4120ecb4f111/src/physics/cam/subcol_SILHS.F90#L842)...I am not sure why). 

### What CAM tag were you using?

cam6_3_155

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Path to a case directory, if applicable

/glade/derecho/scratch/stepheba/cam155_0/

### Will you be addressing this bug yourself?

Any CAM SE can do this

### Extra info

[...]

PR #1016, opened 2024-04-26
part of cam6_4_075: cam_dev silhs bug fix

### LINKED ISSUE #844: Crash when running cam_dev with SILHS
### What happened?

CAM crashes when running -silhs with -phys cam_dev.  The immediate error is coming from subcol.F90, in the subcol_field_avg interface, from the [subcol_field_avg_2dr subroutine](https://github.com/ESCOMP/CAM/blob/ea6d88db42902a85bf6e126dc22f9c42ee6edd15/src/physics/cam/subcol.F90#L392-L394).  Traceback indicates that the problem originates from cam_dev/micro_pumas_cam.F90, specifically where we are trying to call subcol_field_avg on the [proc_rates derived type](https://github.com/ESCOMP/CAM/blob/ea6d88db42902a85bf6e126dc22f9c42ee6edd15/src/physics/cam_dev/micro_pumas_cam.F90#L2519).  This derived type is not present in cam/micro_pumas_cam.F90, which I think explains why this isn't a problem in -phys cam6 runs.  I *think* the problem is that when proc_rates is allocated, [it's allocated with ncol in the first slot](https://github.com/ESCOMP/CAM/blob/ea6d88db42902a85bf6e126dc22f9c42ee6edd15/src/physics/cam_dev/micro_pumas_cam.F90#L1852), but the averaging subroutine requires that the first slot have dimension pcols*psubcols.  Not being familiar with the code, it's not obvious to me what the right solution to this is.  I tried simply allocating proc_rates with psetcols, but this created a bunch of NaNs in CLUBB and the model crashed.

Set up a new case, and set CAM_CONFIG_OPTS="-phys cam_dev -silhs" I think is all it should take.

cam6_3_114

CISL machine (e.g. cheyenne)

Intel

/glade/scratch/stepheba/silhstest_camdev

Any CAM SE can do this

### Extra info

_No response_

bstephens82:
Resolves #844.

This additional code in cam_dev/micro_pumas_cam.F90 will allow a run using cam_dev and SILHS, i.e. something like
~~~
CAM_CONFIG_OPTS="-phys cam_dev -silhs -psubcols 4"
~~~
to run without hitting the [pcols*psubcols abort error](https://github.com/ESCOMP/CAM/blob/776400d2a56d1c95551c9782810e4120ecb4f111/src/physics/cam/subcol.F90#L393) in subcol.F90.  This code is not super familiar to me, so my proposed fix (worked out with advice from @cacraigucar) may not be optimal yet, however I've tested it and it does work to get past the error and the model appears to be running normally afterward although I have not performed a months-long test yet.  

Furthermore, this will **not** resolve a separate problem which seems more general to any subcolumns, where subcolumns print as zeros in the netcdf history files.  See issue #1015 for more info about that.

### DIFF
--- a/src/physics/cam_dev/micro_pumas_cam.F90
+++ b/src/physics/cam_dev/micro_pumas_cam.F90
@@ -2744,9 +2744,10 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
       call subcol_field_avg(qrsedten_sc,  ngrdcol, lchnk, qrsedtenout_grid(:,top_lev:))
       qssedten_sc(:ncol,:) = proc_rates%qssedten(:ncol,1:nlev)
       call subcol_field_avg(qssedten_sc,  ngrdcol, lchnk, qssedtenout_grid(:,top_lev:))
-
-      call subcol_field_avg(proc_rates%umr,       ngrdcol, lchnk, umrout_grid(:,top_lev:))
-      call subcol_field_avg(proc_rates%ums,       ngrdcol, lchnk, umsout_grid(:,top_lev:))
+      umr_sc(:ncol,:) = proc_rates%umr(:ncol,1:nlev)
+      call subcol_field_avg(umr_sc,       ngrdcol, lchnk, umrout_grid(:,top_lev:))
+      ums_sc(:ncol,:) = proc_rates%ums(:ncol,1:nlev)
+      call subcol_field_avg(ums_sc,       ngrdcol, lchnk, umsout_grid(:,top_lev:))
 
       if (micro_mg_version > 2) then
             call subcol_field_avg(state_loc%q(:,:,ixgraupel),    ngrdcol, lchnk, qg_grid)
--- a/src/physics/cam_dev/micro_pumas_cam.F90
+++ b/src/physics/cam_dev/micro_pumas_cam.F90
@@ -1524,6 +1524,47 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    real(r8), pointer :: degrau(:,:)       ! Graupel effective diameter (m)
    real(r8), pointer :: bergstot(:,:)     ! Conversion of cloud water to snow from bergeron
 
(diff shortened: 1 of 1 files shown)


src/physics/cam_dev/micro_pumas_cam.F90
```diff
@@ -1524,6 +1524,47 @@ subroutine micro_pumas_cam_tend(state, ptend, dtime, pbuf)
    real(r8), pointer :: degrau(:,:)       ! Graupel effective diameter (m)
    real(r8), pointer :: bergstot(:,:)     ! Conversion of cloud water to snow from bergeron
 
+   real(r8) :: evapsnow_sc(state%psetcols,pver-top_lev+1)
```
Katetc, line 1527: Would it be possible to add a comment block before all of these explaining that you are duplicating the variables within the proc_rates DDT to get the averaging code to work for diagnostics and no other reason? Thanks!

Katetc (approved):
I am kind of sad about this change because we created the DDT to reduce the amount of code related to declaring all of these variables, and here it is again. I think it would be better to be able to do this within the DDT - ie, have some accessor that gives back the variable in the appropriate dimensions for subcolumn averaging. But, that would take a good bit more work than anybody has time for here, and this fixes the problem, so I'm ok with it.

src/physics/cam_dev/micro_pumas_cam.F90
cacraigucar: @bstephens82 - @nusbaume will be bringing this into his misc tag that he will be making in a day or two.  Do you want to add this comment, or should Jesse just add it when he brings it into his misc PR?

bstephens82: I assume the easiest thing would be for Jesse to just add it? It's been a good while since I've looked at this.

Katetc: That's perfect, thanks so much!

PR #1023, opened 2024-05-02
cam6_4_004: Corrections for WACCMX

### LINKED ISSUE #987: Goemetric height at high altitudes (WACCM-X)
### Issue Type

Other (please describe below)

### Issue Description

From Hanli Liu:

I went through the derivation again and here are the correct formulations:

(1) Z_gm = 1/(1 - (1+Zs/r) * Z_gp/r) * (Zs + (1+Zs/r) * Z_gp)

Z_gm: geometric height
Zs: Surface height
Z_gp: model calculated geopotential height (zm and zi in the model)

If we assume Zs/r << 1, then
(2) Z_gm  ~= 1/(1 - Z_gp/r) * (Zs + Z_gp)

The worst violation is over Mt Everest is 8.8/6370 = 1.4e-3, so this is not a bad assumption. The formulation in ionosphere_interface.F90 Francis sent yesterday is:
(3) Z_gm = (1 + Z_gp/r) * (Zs + Z_gp).

To the first order 1/(1-Z_gp/r) ~ 1+Z_gp/r, the error is about 10% at 600 km (1.104 and 1.094 respectively)

I would suggest using either (1) or (2) in the model for calculating Z3GM.

### Will this change answers?

I Don't Know

Yes

### LINKED ISSUE #1013: Corrections to thermosphere heating diagnostics
### Issue Type

Other (please describe below)

### Issue Description

From Hanli Liu:

Suggested changes to QEI/QEN/QIN history fields In `src/physics/waccmx/ion_electron_temp.F90`:

1. qei/qen/qin are all with unit ev/g/s. We may want to add a comment about their unit at the beginning of the subroutine when they are declared. 
2. Add cpairv to 
    use physconst, only : rairv, mbarv  ! Constituent dependent rair and mbar                                                 

3.  Before adding qei/qen/qin to history, convert them to K/s to be consistent with other heating/cooling terms.
qei_out = qei/sToQConv/cpairv
qen_out = qen/sToQConv/cpairv
qin_out = qin/sToQConv/cpairv

4. Change/add unit for QEN/QEI/QIN to K/s.

### Will this change answers?

No

Yes

fvitt:
closes #987
closes #1013


cacraigucar (approved):

src/physics/waccmx/ion_electron_temp.F90
```diff
@@ -34,6 +34,7 @@ module ion_electron_temp
   use spmd_utils,       only : masterproc
   use cam_logfile,      only : iulog ! Output unit
   use ionos_state_mod,  only : ionos_state
+  use air_composition,only : cpairv
```
nusbaume, line 37: Total nit-pick, but would it help to line up the `only` text to match the other use statments above?

src/physics/waccmx/ion_electron_temp.F90
```diff
@@ -135,16 +136,16 @@ subroutine ion_electron_temp_init(pbuf2d)
...
     call addfld ('TElec'        ,(/ 'lev' /), 'I', 'K',      'Electron Temperature')
     call addfld ('TIon'         ,(/ 'lev' /), 'I', 'K',      'Ion Temperature')
     call addfld ('ElecColDens'  ,horiz_only , 'I', 'TECU',   'Electron Column Density')
     if (.not.steady_state_ion_elec_temp) then
-       call addfld ('QIN'          ,(/ 'lev' /), 'I', 'J/kg/s', 'Ion-neutral Heating')
-       call addfld ('QEN'          ,(/ 'lev' /), 'I', ' ',      'Electron-neutral Heating')
-       call addfld ('QEI'          ,(/ 'lev' /), 'I', ' ',      'Electron-ion Heating')
+       call addfld ('QIN'          ,(/ 'lev' /), 'I', 'K sec-1','Ion-neutral Heating Rate')
+       call addfld ('QEN'          ,(/ 'lev' /), 'I', 'K sec-1','Electron-neutral Heating Rate')
+       call addfld ('QEI'          ,(/ 'lev' /), 'I', 'K sec-1','Electron-ion Heating Rate')
```
nusbaume, line 148: Just curious why the units are using `sec-1` instead of `s-1`?  It probably doesn't actually matter, but CF-conventions and CCPP standard name rules all specify the use of `s-1` when listing variable units, so if there are no strong opinions on your end I might vote to convert `sec` to `s`.

nusbaume (approved):
Looks good!  I just had two (optional) change requests.

PR #1026, opened 2024-05-06
cam6_4_021: CCPP'ize dadadj

jtruesdal:
Refactor dadadj parameterization for inclusion in the CCPP framework.  On the CAM side this entails moving the core parameterization over to atmos_phys and updating the dadadj_cam interface file to call the new library routine.  There are additional mods to cime and ccs_config that are needed for the CAM-SIMA dadadj CCPP validation.


Externals_CAM.cfg
```diff
@@ -50,9 +50,9 @@ tag = ALI_ARMS_v1.0.1
 required = True
 
 [atmos_phys]
-tag = atmos_phys0_02_006
+branch = dadadj
```
nusbaume, line 53: Just adding the usual reminder to point to a tag once ESCOMP/atmospheric_physics#91 has been approved and merged.

src/physics/cam/dadadj_cam.F90
```diff
@@ -98,39 +103,46 @@ subroutine dadadj_tend(dt, state, ptend)
...
+   ! dadadj_run returns t tend, we are passing the ptend%s array to receive the t tendency and will convert it to s
+   ! before it is returned to CAM..
+   call dadadj_run( &
+        ncol, dt, state%pmid(:ncol,:), state%pint(:ncol,:), state%pdel(:ncol,:), state%t(:ncol,:), state%q(:ncol,:,1), cappav(:ncol,:,lchnk), &
+        ptend%s(:ncol,:), ptend%q(:ncol,:,1), dadpdf(:ncol,:), scheme_name, errmsg, errflg)
+
+   if (errflg /= 0) then
+      ! error exit
+      write(errstring, *) errmsg,' at lat,lon:', &
+           state%lat(errflg)*180._r8/pi, state%lon(errflg)*180._r8/pi
```
nusbaume, line 138: I believe this error message could fail if the `errflg` is being returned via a bad `allocate` call in `dadadj_run`.  Maybe it would be worth checking the `errmsg` string (say via the ["index" function](https://gcc.gnu.org/onlinedocs/gfortran/INDEX-intrinsic.html)) for the word "Allocate", and then only if that word is absent use the `errflg` as the physics column index?

src/physics/cam/dadadj_cam.F90
```diff
@@ -98,39 +103,46 @@ subroutine dadadj_tend(dt, state, ptend)
...
+
+   ! dadadj_run returns t tend, we are passing the ptend%s array to receive the t tendency and will convert it to s
+   ! before it is returned to CAM..
+   call dadadj_run( &
+        ncol, dt, state%pmid(:ncol,:), state%pint(:ncol,:), state%pdel(:ncol,:), state%t(:ncol,:), state%q(:ncol,:,1), cappav(:ncol,:,lchnk), &
+        ptend%s(:ncol,:), ptend%q(:ncol,:,1), dadpdf(:ncol,:), scheme_name, errmsg, errflg)
+
+   if (errflg /= 0) then
+      ! error exit
+      write(errstring, *) errmsg,' at lat,lon:', &
```
nusbaume, line 137: Should `errmsg` be trimmed here?  Otherwise I think it is technically the same length as `errstring`, and so it could cause problems when the additional text is being added to `errstring` via this write statement.

nusbaume (changes requested):
Looks good!  Just had a couple of concerns related to the new error message generation.

src/physics/cam/dadadj_cam.F90
jtruesdal: I'll change the dadadj error to read  'Convergence failure at column i, zeps > 1.e-4 (errflg set to failing column index)'.  Any other users of this scheme could then use that value to back out location info.  I will also add a check for the word 'Convergence' and appended the additional error info on lat/lon if found.

src/physics/cam/dadadj_cam.F90
```diff
@@ -98,39 +103,51 @@ subroutine dadadj_tend(dt, state, ptend)
...
+      else
+         errstring=trim(errmsg)
+      end if
+      call endrun('Error dadadj_tend:'//trim(errstring))
+   end if
+
+   call outfld('DADADJ_PD',  dadpdf(:ncol,:),  ncol, lchnk)
+
+   ! convert the t tendency to an s tendency for cam
+   ptend%s(:ncol,:) = ptend%s(:ncol,:) * cpairv(:ncol,:,lchnk)
```
cacraigucar, line 150: @nusbaume - Just wanting to confirm our definition of `tendency_of_air_temperature`?  Is it without out the cpairv adjustment?  If so, we may want to augment the conversion document to indicate the ptend%s special adjustment?

cacraigucar (commented):
I found it easier to review this along with the atmospheric_physics PR.

src/physics/cam/dadadj_cam.F90
jtruesdal: Trimmed.

src/physics/cam/dadadj_cam.F90
jtruesdal: Changed.

src/physics/cam/dadadj_cam.F90
nusbaume: Yes, `tendency_of_air_temperature` is basically the change in `state%t` (or air temperature), while `ptend%s` is technically the `tendency_of_dry_air_enthalpy_at_constant_pressure`, which in CAM we use to calculate the change `state%t` in `physics_update` by dividing by `cpairv`.  So in other words:

`ptend%t = ptend%s / cpairv`

or in standard name lingo:

```
tendency_of_air_temperature = tendency_of_dry_air_enthalpy_at_constant_pressure / composition_dependent_specific_heat_of_dry_air_at_constant_pressure
```

Hopefully that makes sense?  If not just let me know.  Regardless I can say that I am pretty sure this new CAM code is correct as-is.

.gitmodules
```diff
@@ -35,8 +35,8 @@
 
 [submodule "atmos_phys"]
 	path = src/atmos_phys
-	url = https://github.com/ESCOMP/atmospheric_physics
-        fxtag = atmos_phys0_02_006
+        url = https://github.com/jtruesdal/atmospheric_physics
+        fxtag = dadadj
```
cacraigucar, line 39: Once this is updated, then we can schedule this to go into CAM

cacraigucar (changes requested):

cacraigucar (approved):

PR #1028, opened 2024-05-07
cam6_4_001: Change name of physics package 'cam_dev' to 'cam7'

### LINKED ISSUE #813: Introduce "-phys cam7" and remove "-phys cam_dev"

Remove the "-phys cam_dev" setting and implement "-phys cam7"

This change will include:
- Replace all references to cam_dev with cam7
- Move cam_dev directory to a directory named cam7
- Bump the cam numbering to cam6_4_001 with this change

_No response_

No

Any CAM SE can do this

### LINKED ISSUE #1033: Change DART test to use 128 instead of 108 processors
### Issue Type

Other (please describe below)

### Issue Description

@fischer-ncar reports that  SMS_C80_P108x1_Lh1.f09_f09_mg17.FHIST_DARTC6.derecho_intel.cam-dartcambigens is failing

Changing to 128x1 solves the problem

### Will this change answers?

No

Any CAM SE can do this

### LINKED ISSUE #1038: Replace ne16np4 grid for WACCM HIST test with ne16np4.pg3
### What happened?

The test SMS_D_Ln9.ne16_ne16_mg17.FX2000.derecho_intel.cam-outfrq9 fails because of a missing transient dataset for ne16np4, while there is one for ne16np5.pg3.

Run SMS_D_Ln9.ne16_ne16_mg17.FX2000.derecho_intel.cam-outfrq9

cam6_3_161

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

See the CTSM issue on this:

https://github.com/ESCOMP/CTSM/issues/2547

There it was decided that CAM should address this and change the test to ne30np4.pg3 rather than adding a new dataset in CTSM.

### LINKED ISSUE #1039: Change transient ne30np4 cam tests to ne30np4.pg3
### Issue Type

Other (please describe below)

### Issue Description

In CESM3, we are discontinuing transient cases (which require a landuse_timeseries file from CTSM) for "vanilla" se. Some of our existing cam tests are vanilla transient cases, causing test failures https://github.com/ESCOMP/CTSM/issues/2549. I propose we replace all transient vanilla tests with pg3 tests to resolve these failures.

@PeterHjortLauritzen I would like your feedback on whether this is an acceptable resolution to the current test failures.

@fvitt @ekluzek @cacraigucar

### Will this change answers?

No

No

brian-eaton:
#813 - Introduce "-phys cam7" and remove "-phys cam_dev".  
Related changes:
- The compset tokens CAM%DEV are replaced by CAM70
- The src/physics/cam_dev/ directory is renamed src/physics/cam7
- No compset names were changed.

N.B.: This PR depends on a CMEPS tag to make the corresponding change of CAM%DEV to CAM70 in buildnml

. resolves #813

Additional misc issue and issues resolved to address regression test failures:
. resolves #1033
. resolves #1038
. resolves #1039


bld/configure
```diff
@@ -1300,6 +1300,20 @@ EOF
...
+#              independently of the number of model layers.
+
+# Set default
+my $model_top = 'none';
+$cfg_ref->set('model_top', $model_top);
+
+# user override
+if (defined $opts{'model_top'}) {
+    $cfg_ref->set('model_top', $opts{'model_top'});
+}
```
cacraigucar, line 1314: Should there be a check that a user is using cam7 if they specify model_top?  I could see a naïve user wanting to use 'lt' or 'mt' with cam5 or cam6, especially for comparison purposes.  Do we want to prohibit this or would things work okay?

doc/ChangeLog
```diff
@@ -1,4 +1,127 @@
+===============================================================
+
+Tag name:
```
cacraigucar, line 3: This will be cam6_4_001 whenever it comes in

cacraigucar (changes requested):

doc/ChangeLog
gold2718: Out of curiosity, if you are introducing CAM7 to the world, why isn't this cam7_0_001?

cacraigucar: I am reserving cam7_0_000 for when we make our release branch, which is still quite a ways off.  I found it useful with the last release that cam6_0_001 was the first tag which wasn't in the release.

bld/build-namelist
```diff
@@ -3525,13 +3525,6 @@ if ($clubb_sgs  =~ /$TRUE/io) {
    add_default($nl, 'do_hb_above_clubb');
 }
 
-# Force exit if running cam_dev and CLUBB is off
-if ($phys eq 'cam_dev') {
-    if ($clubb_sgs  =~ /$FALSE/io) {
-        die "$ProgName - ERROR: If running cam_dev physics, do_clubb_sgs must be .true.\n";
-    }
-}
-
```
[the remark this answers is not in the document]
brian-eaton: The variable `do_clubb_sgs` is not meant to be user settable since turning on clubb is a build-time decision.  This variable is similar to, for example, `cam_physpkg` which is communicating build-time information to the run-time environment.  `build-namelist` has checks that the user doesn't try to set
`cam_physpkg` and similar variables.  I will add a similar check for `do_clubb_sgs` and update the documentation in the definition file.

bld/configure
brian-eaton: I agree that we should restrict this option to cam7.  Currently the only use of the `lt`/`mt` `model_top` settings are as attributes in the `namelist_defaults_cam.xml file`.  They are being used to distinguish cam7 settings from previous cam versions.  For example, cam6 is a low top model, but `model_top='lt'` is used to distinguish cam6 from cam7 settings for some GWD variables and an SE variable.

I will add the check in `configure` that you suggested.  I will also modify the compset matching in `config_component.xml` by changing `_CAM.*%LT` to `_CAM70%LT` and similarly for `%MT`.

cacraigucar (approved):

nusbaume (approved):

cacraigucar:
Please update the CICE external to be cesm_cice6_5_0_10.  This will eliminate you from seeing components/cice being "modified".  This tag was made after I started up my regression runs for cam6_3_162.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1533,32 +1531,12 @@
   </test>
   <test compset="F2000dev" grid="f09_f09_mg17" name="ERP_Ln9" testmods="cam/outfrq9s_mg3" supported="false">
     <machines>
-      <machine name="derecho" compiler="intel" category="aux_cam"/>
       <machine name="derecho" compiler="intel" category="aux_pumas"/>
-      <machine name="derecho" compiler="intel" category="prealpha"/>
```
cacraigucar, line 1538: @PeterHjortLauritzen and @adamrher  - I don't believe we are dropping support for running CAM7 with the FV dycore, correct?  Since I also believe we are not supporting 32 levels in CAM7, do we need to create initial condition files for at least the LT case for the various grids?  The reason I am asking is that it appears that with Brian's changes to move to CAM7, we no longer have any FV tests with CAM7.

PeterHjortLauritzen: No we are not supporting FV with CAM7. FV should only be supported for CAM4,5,6 configurations.

nusbaume (approved):
Looks good!  I did want to note though that the `resolves #XXX` text should be added to the main PR  description box on Github for all of the additional issues that were fixed so that they can be closed automatically once the PR has been merged.  Thanks!

Issue #1036, opened 2024-05-17
New CLUBB external with fixes to support GPU testing


Katetc:
### Issue Type

Externals Update

### Issue Description

Some issues have been found with the GPU code and the CLUBB externals. Te UWM guys are working on a new tag now to update the CLUBB external so the GPU test suite can move forward correctly.

Unsure of answer changes at this point. If there are any, they should be very minimal.

### Will this change answers?

Yes

### Will you be implementing this yourself?

Yes

Katetc:
Making this issue to let @sjsprecious see that we are prioritizing this!

jedwards4b:
I tried update cam6_4_016 to use clubb_4ncar_20240605_73d60f6 in order to get the fix for mt95 in the nvhpc compiler - however this leads to a host of clubb interface errors - is there a way to get the fix only for mt95 so that I can pass it on to the nvhpc compiler developers?

Katetc:
This was addressed in pr #1086 and cam tag cam6_4_019. Not sure why it wasn't closed. Closing now.

PR #1048, opened 2024-05-29
part of cam6_4_019: Add GPU regression test suite

### LINKED ISSUE #1047: Missing GPU tests in the regular CAM regression test suite

Currently all the CAM regression tests run on the Derecho's CPUs only. This will fail to capture any potential bugs that only break the GPU code (e.g., PUMAS) and any compiler bugs (mostly the NVIDIA compiler) related to the GPU code.

_No response_

No

sjsprecious:
This PR updates the Casper GPU tests in CAM to Derecho GPU tests and shortened the tests to two, whose details as listed below:

- F2000dev compset
- ne30pg3 resolution
- PUMAS with MG3 option, default namelist settings
- one GPU test with a single GPU node on Derecho and another GPU test with two GPU nodes and different PCOLS values

Note that there is a GPU code bug in CLUBB (#1036) and a potential nvhpc compiler bug (I didn't compile the CLUBB GPU code but the compiler still tried to run the CLUBB code on GPU, which caused a runtime error). Thus I manually removed all the GPU directives in the CLUBB code in order to get the tests working for the PUMAS GPU code so far.

Currently I name the GPU tests as a separate category `derecho_gpu`. Once the CLUBB GPU code bug is fixed and verified for correctness, I can run both PUMAS and CLUBB on the GPU, and add the tests to the `aux_cam` category to be run regularly in the future.  

fix #1047


cime_config/testdefs/testlist_cam.xml
```diff
@@ -1485,6 +1485,22 @@
...
+    </options>
+  </test>   
+  <test compset="F2000dev" grid="ne30pg3_ne30pg3_mg17" name="ERS_Ln9_G4-a100-openacc" testmods="cam/outfrq9s_mg3_pcols760">
+    <machines>
+      <machine name="derecho" compiler="nvhpc" category="derecho_gpu"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:30:00</option>
+    </options>
+  </test>
```
cacraigucar, line 1503: Do you want both of these tests to be routinely tested?  I would suggest making one test `aux_cam` (which is tested every time a CAM tag is made) and the other `prealpha `(which is tested every time a CESM tag is made).  You can keep `derecho_gpu` as a second test in both tests so that you have a convenient way to run all the GPU tests.

cacraigucar (changes requested):

cime_config/testdefs/testlist_cam.xml
sjsprecious: Thanks @cacraigucar for your comments. Eventually I want these tests to be routinely tested and what you suggested makes sense to me. Currently I am waiting for a new CLUBB tag from #1036, which hopefully fixes the CLUBB GPU bug so that I do not need the manual revision here any more. Before that issue is fixed, I think we should not add these two tests to either `aux_cam` or `prealpha` because they will always fail at runtime. This is why I make a separate category `derecho_gpu` here.

Issue #1053, opened 2024-06-04
Bug in CLUBB interface code


PeterHjortLauritzen:
### What happened?

@klindsay28 found a bug in the CLUBB interface code (clubb_intr.F90) with this call (commited in tag cam6_3_094):
```
2553     call set_wet_to_dry(state1)
```
Our conversion from wet to dry mixing ratios should have been:

```
  do m = 1,pcnst
     if (cnst_type(m).eq.'wet') then
        state1%q(:ncol,:,m) = state1%q(:ncol,:,m)*state1%pdel(:ncol,:)/state1%pdeldry(:ncol,:)
     endif
  end do
```
but the set_wet_to_dry subroutine call executes:

```
  do m = 1,pcnst
     if (cnst_type(m).eq.'dry') then
        state1%q(:ncol,:,m) = state%q(:ncol,:,m)*state1%pdel(:ncol,:)/state1%pdeldry(:ncol,:)
     endif
  end do
```
So, as Keith states, it makes dry tracers "double dry" and does not convert wet tracers to dry (as the name seems to imply!). That also means that the scaling of the tendencies of wet tracers is incorrect; the tracer tendencies (which are already using moist mixing ratios) are converted to moist again."

### What are the steps to reproduce the bug?

run any configuration of CAM that uses CLUBB!

### What CAM tag were you using?

 cam6_3_161

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Yes

### Extra info

Working on PR ...

brian-eaton:
Hi @PeterHjortLauritzen, @klindsay28, @adamrher.

The name of the subroutine `set_wet_to_dry` is misleading.  But the subroutine, which has been in CAM since cam3_1 (2005) is correct.  The implicit assumption here is that the physics package operates on constituents which are dry except for the water constituents (both mass and number concentrations).  As far as I know this is still true; the moist adjustment schemes assume the values of water species are wet mixing ratios.  The purpose of `set_dry_to_wet` and `set_wet_to_dry` is to convert the 'dry' species mixing ratios, which are known by the public `cnst_type` array, from dry to wet, and then back from wet to dry.  Hence those subroutine are designed to provide conversions for **only** the dry species.  The `cnst_type` array, which is used to distinquish the water (wet) from the non-water (dry) species, is set by the `add_cnst` call for each constituent during the register phase of initialization.

[...]

PR #1054, opened 2024-06-04
cam6_4_006: fix CLUBB interface bug

PeterHjortLauritzen:
fixes https://github.com/ESCOMP/CAM/issues/1053


src/physics/cam/clubb_intr.F90
gold2718, line 1: I'm curious why make this fix in place here instead of fixing `set_wet_to_dry` in physics_types.F90?
The "broken" `set_wet_to_dry` is still called in `vertical_diffusion_tend` which is called in `tphysac`.

PeterHjortLauritzen: What the subroutine `set_wet_to_dry` does would be clearer if it was named `undo_set_dry_to_wet` or `revert_set_dry_to_wet`

Explanation:

subroutine `set_dry_to_wet` does as the subroutine name suggests: it converts mixing ratios that are dry (```(cnst_type(m).eq.'dry')```) to wet. 

subroutine `set_wet_to_dry` undoes this operation by converting the tracers that are supposed to be dry back to dry (from wet).

set_wet_to_dry does NOT convert wet tracers to dry as the name seems to imply: here is the code from set_wet_to_dry

```
  do m = 1,pcnst
     if (cnst_type(m).eq.'dry') then
        state%q(:ncol,:,m) = state%q(:ncol,:,m)*state%pdel(:ncol,:)/state%pdeldry(:ncol,:)
     endif
  end do
```

Hence the operation in `vertical_diffusion_tend` is correct as it converts dry tracers to wet and reverts the operation at the end.

Converting all of CAM physics to a dry basis would highly likely make sure these kinds of bugs are avoided ...

gold2718: Ah, thanks. I keep forgetting about that subtlety (probably because of the broken name and also getting rusty in that part of the code). Also, there is no tag on a constituent's current state which has caused plenty of other problems in the past. I'm still hoping this can get fixed in CAM-SIMA but I don't know if that has been fully fleshed out yet.

sjsprecious: @PeterHjortLauritzen @gold2718 I have wondered this question for a while: why do we need to use dry basis and wet basis in different CAM parts?

PeterHjortLauritzen: Traditionally, moist physics developers have always used moist mixing ratios for water species. For example, specific humidity (which equals the moist mixing ratio of water vapor) is a directly measurable and observable quantity, making it a logical choice for moist physics. However, chemists usually use dry mixing ratios; for them, a moist basis is problematic because it includes a water vapor signal embedded in the mixing ratio. Numerical schemes are generally designed to conserve a constant mixing ratio. Specifically, some chemical species remain constant (on a dry basis) in most of the atmosphere. If one uses a moist basis, the numerical schemes will likely introduce a water vapor signal into the mixing ratio that was supposed to remain constant.

We wrote a detailed argument for using a dry basis, which you can find here:

https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2017MS001257

(see Introduction)

sjsprecious: Thanks @PeterHjortLauritzen so much for your detailed explanation and sharing of your paper. That is very helpful.

brian-eaton (changes requested):
I agree with Steve's initial reaction: why don't we fix the utility code in physics_types.F90?  By doing so we have the opportunity to add some clarity the complexity of having different types of constituents and different parameterizations that expect wet or dry mixing ratios based on the type of the constituent (water specie or not).  I have made an attempt at this below and invite comments for further clarification.  I left the subroutine names the same since the functionality of converting between wet and dry mixing ratios is fairly clear in those names.  What's not clear is what constituent types the conversion is applied to.  Adding an argument will make that clearer.  I considered making the argument optional with the default to convert the dry species as it currently does.  But there aren't many calls to these subroutines so updating them all is easy, and will make existing code more clear.
```
subroutine set_wet_to_dry(state, convert_cnst_type)

  ! Convert mixing ratios from a wet to dry basis for constituents of type
  ! convert_cnst_type.  Constituents are given a type when they are added
  ! to the constituent array by a call to cnst_add during the register
  ! phase of initialization.  There are two constituent types: 'wet' for
  ! water species and 'dry' for non-water species.

  use constituents,  only: pcnst, cnst_type

  type(physics_state),  intent(inout) :: state
  character(len=*),     intent(in)    :: convert_cnst_type
  
  ! local variables
  integer :: m, ncol
  character(len=*), parameter :: sub = 'set_wet_to_dry'
  !-----------------------------------------------------------------------------

  ! check input
  if (.not.(convert_cnst_type == 'wet' .or. convert_cnst_type == 'dry')) then
    write(iulog,*) sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type
    call endrun(sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type)
  end if

  ncol = state%ncol

  do m = 1,pcnst
     if (cnst_type(m) .eq. convert_cnst_type) then
        state%q(:ncol,:,m) = state%q(:ncol,:,m)*state%pdel(:ncol,:)/state%pdeldry(:ncol,:)
     endif
  end do

end subroutine set_wet_to_dry
```
Similar changes can be made to `set_dry_to_wet`

PeterHjortLauritzen:
Thanks for looking into this. I am fine with the proposed change.

brian-eaton (approved):
I have implemented the suggested changes.

src/physics/cam/physics_types.F90
```diff
@@ -1481,40 +1481,72 @@ end subroutine set_state_pdry
...
 
+  ! local variables
   integer m, ncol
+  character(len=*), parameter :: sub = 'set_wet_to_dry'
+  !-----------------------------------------------------------------------------
+
+  ! check input
+  if (.not.(convert_cnst_type == 'wet' .or. convert_cnst_type == 'dry')) then
+    write(iulog,*) sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type
+    call endrun(sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type)
```
nusbaume, line 1505: Couple minor typos:

```suggestion
    write(iulog,*) sub//': FATAL: convert_cnst_type not recognized: '//convert_cnst_type
    call endrun(sub//': FATAL: convert_cnst_type not recognized: '//convert_cnst_type)
```

src/physics/cam/physics_types.F90
```diff
@@ -1481,40 +1481,72 @@ end subroutine set_state_pdry
...
+  ! local variables
   integer m, ncol
+  character(len=*), parameter :: sub = 'set_dry_to_wet'
+  !-----------------------------------------------------------------------------
+
+  ! check input
+  if (.not.(convert_cnst_type == 'wet' .or. convert_cnst_type == 'dry')) then
+    write(iulog,*) sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type
+    call endrun(sub//': FATAL: convert_cnst_type not recognozed: '//convert_cnst_type)
+  end if
```
nusbaume, line 1542: Typos:

```suggestion
  if (.not.(convert_cnst_type == 'wet' .or. convert_cnst_type == 'dry')) then
    write(iulog,*) sub//': FATAL: convert_cnst_type not recognized: '//convert_cnst_type
    call endrun(sub//': FATAL: convert_cnst_type not recognized: '//convert_cnst_type)
  end if
```

nusbaume (approved):
Looks good to me!  Only issue I found was few minor typos (so I don't feel the need to re-review).  Thanks!

src/physics/cam/physics_types.F90
brian-eaton: Thanks for catching that.  Fixed.

PR #1081, opened 2024-07-08
Switch log-gamma call to a Fortran 2008 intrinsic

briandobbins:
We can't use the old method of computing log-gamma; this removes that code, and replaces it with a Fortran 2008 intrinsic.

This may be answer changing, but should be mathematically identical.  I will run the full suite of WACCMX tests once there is a sign-off by Francis.


fvitt:
@hanli-liu, Are we okay with replacing the log-gamma routine in the Weimer model?

@briandobbins 
Most of the waccmx tests do not test the weimer potential model. 
However, this will test weimer:
```
SMS_Ln3.f09_f09_mg17.FXHIST.derecho_intel.cam-waccmx_weimer
```

briandobbins:
Got it - I'll try that test.  I do expect we'll see numerical differences, but they should be small.

fvitt:
The differences in the Weimer high-latitude electric potential (PHIHM) due to this change is quite small (~1.e-6 relative difference).  So I think we are good to go with this change.

fvitt (approved):

briandobbins:
@fvitt FYI, the FXHIST test unexpectedly (to me) passed.  I'm guessing this simply means we aren't testing any fields this log-gamma change would've impacted, but I wanted to ping you in case you want to update the test, or take a deeper look.

If not, let me know, and I'll proceed with this PR today.

fvitt:
> @fvitt FYI, the FXHIST test unexpectedly (to me) passed. I'm guessing this simply means we aren't testing any fields this log-gamma change would've impacted, but I wanted to ping you in case you want to update the test, or take a deeper look.
> 
> If not, let me know, and I'll proceed with this PR today.

Not surprising.   None of the aux_cam tests use the modified module.  
There are prealpha and waccmx tests  that exercise the module:
```
prealpha          : ERS_Ld3.f19_f19_mg17.FXHIST.derecho_intel.cam-waccmx_weimer                                                         
waccmx            : ERS_Ld3.f19_f19_mg17.FXHIST.derecho_intel.cam-waccmx_weimer                                                         
waccmx            : SMS_D_Ln9_P384x3.f09_f09_mg17.FXHIST.derecho_intel.cam-waccmx_weimer                                                
```
These should show differences in baseline tests.

briandobbins:
Ah, got it, apologies for my misunderstanding -- this makes sense now.  I'm going to go ahead with the PR then, since you've already approved this knowing the differences we'll see, and this clears up the mystery of why I didn't see something in the FXHIST test.

PR #1029, opened 2024-05-13
 cam6_4_033: Add updated ncdata and bnd_topo files for MPASv8

gdicker1:
The new initial condition files contain fields that are expected by routines in cam_dev (cam7) physics. The new topo files were regenerated due to changes in the NCAR/Topo software. These files are only used when `phys="cam_dev"` to ensure previous compsets and/or tests aren't affected.

These files were checked for NaNs and runs were attempted for the `F2000dev` compset before making this PR.

Fixes #995 and #1094

PR #1095, opened 2024-07-18
part of cam6_4_033: Wrap MPAS longitude values to [0,2pi) range

### LINKED ISSUE #1094: Wrap MPAS-A longitudes to [0, 2pi) range

While testing MPAS real-data initial conditions I proposed to add in #1029, F2000dev or F2000climo runs at 480km would fail while FADIAB runs would succeed. The cesm.log\* files for failed runs would include many lines like `imp_sol: Time step   1.1250000000000E+01 failed to converge ...`, the amount of time per timestep would drastically increase, and runs would eventually terminate due to a `CRITICAL ERROR: NaN detected in 'w' field.` message from MPAS-A.

After a fair bit of trial and error, I think this comes down to some part of CAM expecting longitudes to be in the $[0,2\pi)$ range. This is not expected nor required by MPAS-A, and the MPAS-A \*.grid.nc files may have longitudes ranging from $[0,2\pi]$ or $[-\pi, \pi]$ (or possibly even arbitrary radian values/ranges).

This is easily fixed by adding a few lines to put longitudes in the correct range. I propose to add these within `cam_mpas_read_static` subroutine in the cam_mpas_subdriver.F90 file. Since this will affect all longitude values, this change could affect regression tests.

_No response_

Yes

Yes

gdicker1:
This PR addresses issues that were seen when using ncdata with longitudes in the $[-\pi, \pi]$ range in cases with full physics.  This would cause the atmosphere to become unphysical and runs to fail. This change is made just after the longitude values are read to ensure they are correct for the remainder of execution.

Fixes #1094

### DIFF
--- a/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
+++ b/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
@@ -884,15 +884,16 @@ contains
 
        use pio, only : file_desc_t
 
-       use mpas_kind_types, only : StrKIND
+       use mpas_kind_types, only : StrKIND, RKIND
        use mpas_io_streams, only : MPAS_createStream, MPAS_closeStream, MPAS_streamAddField, MPAS_readStream
        use mpas_derived_types, only : MPAS_IO_READ, MPAS_IO_NETCDF, MPAS_Stream_type, MPAS_pool_type, &
                                       field0DReal, field1DReal, field2DReal, field3DReal, field1DInteger, field2DInteger, &
                                       MPAS_STREAM_NOERR
        use mpas_pool_routines, only : MPAS_pool_get_subpool, MPAS_pool_get_field, MPAS_pool_create_pool, MPAS_pool_destroy_pool, &
-                                      MPAS_pool_add_config
+                                      MPAS_pool_add_config, MPAS_pool_get_dimension, MPAS_pool_get_array
        use mpas_dmpar, only : MPAS_dmpar_exch_halo_field
        use mpas_stream_manager, only : postread_reindex
+       use mpas_constants, only : pii
 
        ! Arguments
        type (file_desc_t), pointer :: fh_ini
@@ -929,8 +930,13 @@ contains
 
        type (MPAS_Stream_type) :: mesh_stream
 
+       integer, pointer :: nCells
+       real(kind=RKIND), dimension(:), pointer :: lonCell_arr
+
        nullify(cell_gradient_coef_x)
        nullify(cell_gradient_coef_y)
+       nullify(nCells)
+       nullify(lonCell_arr)
 
        call MPAS_createStream(mesh_stream, domain_ptr % ioContext, 'not_used', MPAS_IO_NETCDF, MPAS_IO_READ, &
                               pio_file_desc=fh_ini, ierr=ierr)
@@ -1170,6 +1176,12 @@ contains
        if (ierr /= MPAS_STREAM_NOERR) then
            call endrun(subname//': FATAL: Failed to close static input stream.')
        end if
+
+       call mpas_pool_get_dimension(meshPool, 'nCells', nCells)
+       call mpas_pool_get_array(meshPool, 'lonCell', lonCell_arr)
+
+       ! Ensure longitudes w/i [0, 2*pi) range
+       lonCell_arr(:) = lonCell_arr(:) - (2.*pii) * floor(lonCell_arr(:) / (2.*pii))
 
        !
        ! Perform halo updates for all decomposed fields (i.e., fields with
--- a/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
+++ b/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
@@ -1178,7 +1178,7 @@ contains
        call mpas_pool_get_array(meshPool, 'lonCell', lonCell_arr)
 
        ! Ensure longitudes w/i [0, 2*pi) range
-       lonCell_arr(:) = lonCell_arr(:) - (2.*pii) * floor(lonCell_arr(:) / (2.*pii))
+       lonCell_arr(:) = lonCell_arr(:) - (2._RKIND*pii) * floor(lonCell_arr(:) / (2._RKIND*pii))
 
        !
        ! Perform halo updates for all decomposed fields (i.e., fields with
--- a/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
+++ b/src/dynamics/mpas/driver/cam_mpas_subdriver.F90
@@ -890,7 +890,7 @@ contains
                                       field0DReal, field1DReal, field2DReal, field3DReal, field1DInteger, field2DInteger, &
                                       MPAS_STREAM_NOERR
        use mpas_pool_routines, only : MPAS_pool_get_subpool, MPAS_pool_get_field, MPAS_pool_create_pool, MPAS_pool_destroy_pool, &
-                                      MPAS_pool_add_config, MPAS_pool_get_dimension, MPAS_pool_get_array
+                                      MPAS_pool_add_config, MPAS_pool_get_array
        use mpas_dmpar, only : MPAS_dmpar_exch_halo_field
        use mpas_stream_manager, only : postread_reindex
        use mpas_constants, only : pii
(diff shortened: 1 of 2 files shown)


gdicker1:
NOTE: This PR branch is based on cam6_3_148 for compatibility with EarthWorks. This can be updated at any time, in whatever way reviewers would prefer.

Changes in this PR should also allow #1029 to proceed.

src/dynamics/mpas/driver/cam_mpas_subdriver.F90
```diff
@@ -1171,6 +1177,12 @@ subroutine cam_mpas_read_static(fh_ini, endrun)
            call endrun(subname//': FATAL: Failed to close static input stream.')
        end if
 
+       call mpas_pool_get_dimension(meshPool, 'nCells', nCells)
```
mgduda, line 1180: Is `nCells` actually needed?

gdicker1: Thanks for the catch, it's not needed now. It was a leftover from when I was iterating on this. Addressed by https://github.com/ESCOMP/CAM/pull/1095/commits/f3dfa5173aea7d1c4a90dafd53f3e3fe2e0660a9

src/dynamics/mpas/driver/cam_mpas_subdriver.F90
```diff
@@ -1171,6 +1175,11 @@ subroutine cam_mpas_read_static(fh_ini, endrun)
            call endrun(subname//': FATAL: Failed to close static input stream.')
        end if
 
+       call mpas_pool_get_array(meshPool, 'lonCell', lonCell_arr)
+
+       ! Ensure longitudes w/i [0, 2*pi) range
+       lonCell_arr(:) = lonCell_arr(:) - (2.*pii) * floor(lonCell_arr(:) / (2.*pii))
```
jtruesdal, line 1181:
```suggestion
       lonCell_arr(:) = lonCell_arr(:) - (2._RKIND*pii) * floor(lonCell_arr(:) / (2._RKIND*pii))
```

jtruesdal (approved):
Looks Good!  I have one small request.  Append the kind to the number literal.  After that it should be good to go.

gdicker1:
@jtruesdal before I push to address your comment, I just noticed that I didn't wrap the commit message for f3dfa5173aea7d1c4a90dafd53f3e3fe2e0660a9 appropriately. In other repos, I would fix the message and force-push the branch.

How should I proceed here? Just ignore it?

jtruesdal:
I'm not sure what you mean by "wrap the commit message".  Everything looks Ok to me when I look at the 3 commits and their associated messages.  We mostly go by the Changelog and the commit messages are useful but not critical.

gdicker1:
Fair enough, I won't worry about it.

> I'm not sure what you mean by "wrap the commit message".
Most resources on git commit messages suggest to break body text at 72 characters and continue on the next line.

src/dynamics/mpas/driver/cam_mpas_subdriver.F90
gdicker1: Thanks for the reminder! Addressed by [288f3ed](https://github.com/ESCOMP/CAM/pull/1095/commits/288f3ed05d3e3e8d475bc39f967ab420f87bb937)

src/dynamics/mpas/driver/cam_mpas_subdriver.F90
```diff
@@ -1171,6 +1175,11 @@ subroutine cam_mpas_read_static(fh_ini, endrun)
            call endrun(subname//': FATAL: Failed to close static input stream.')
        end if
 
+       call mpas_pool_get_array(meshPool, 'lonCell', lonCell_arr)
+
+       ! Ensure longitudes w/i [0, 2*pi) range
+       lonCell_arr(:) = lonCell_arr(:) - (2._RKIND*pii) * floor(lonCell_arr(:) / (2._RKIND*pii))
```
mgduda, line 1181: It might have been me who initially suggested adding this code in `cam_mpas_read_static`, but after a second look, I'm wondering whether this code would better belong in the `setup_time_invariant` routine? We do other checks and processing of fields there after the call to `cam_mpas_read_static`, and moving the modification of the longitude range there would let the focus of `cam_mpas_read_static` be purely on reading fields.

src/dynamics/mpas/driver/cam_mpas_subdriver.F90
```diff
@@ -1171,6 +1175,11 @@ subroutine cam_mpas_read_static(fh_ini, endrun)
            call endrun(subname//': FATAL: Failed to close static input stream.')
        end if
 
+       call mpas_pool_get_array(meshPool, 'lonCell', lonCell_arr)
+
+       ! Ensure longitudes w/i [0, 2*pi) range
```
mgduda, line 1180: It took me a second to realize that `w/i` probably means "within". It probably wouldn't hurt readability to just write this all out, and to also note that this doesn't really matter for the MPAS-A dycore, but it does matter for other parts of CAM that inherit the longitude field from the dycore.

mgduda (changes requested):
Besides giving some consideration to the two comments I've left, I think it would be worth using a loop with a conditional to only modify longitudes that aren't already in the interval [0, 2*Pi). It's ultimately likely to be slower, but this only happens once when the model starts up, and the cost is almost certainly negligible when compared with the cost of filesystem accesses. The benefit would be that, for meshes that provide longitudes in CAM's required range, those values would remain bitwise identical to their values in the input file.

gdicker1:
@mgduda I implemented your suggestions in [fabe37a](https://github.com/ESCOMP/CAM/pull/1095/commits/fabe37a9910838690e43b4a6695079971156e5aa). Thanks for the explanation you gave on the "loop with a conditional".

src/dynamics/mpas/dyn_grid.F90
```diff
@@ -482,6 +485,16 @@ subroutine setup_time_invariant(fh_ini)
                            ') does not match plev ('//int2str(nVertLevelsSolve)//').')
    end if
 
+   ! Ensure longitudes are within the [0,2*pi) range, and only remap values that
+   ! are outside the range.  Some non-simple physics in CAM require this
+   ! longitude range, MPAS may have any (radian) value in lonCell
+   call mpas_pool_get_array(meshPool, 'lonCell', lonCell)
+   do k=1,nCells
+      if (lonCell(k) < 0._r8 .or. lonCell(k) .ge. (2._r8 * pi)) then
```
mgduda, line 493: For consistency (and perhaps even CAM coding standards), I think `.ge.` should be changed to `>=`.

src/dynamics/mpas/dyn_grid.F90
```diff
@@ -482,6 +485,16 @@ subroutine setup_time_invariant(fh_ini)
                            ') does not match plev ('//int2str(nVertLevelsSolve)//').')
    end if
 
+   ! Ensure longitudes are within the [0,2*pi) range, and only remap values that
+   ! are outside the range.  Some non-simple physics in CAM require this
+   ! longitude range, MPAS may have any (radian) value in lonCell
```
mgduda, line 490: The phrasing of this last part doesn't quite give the right sense to me. Perhaps rather than saying "MPAS may have any (radian) value in lonCell" we could say something like "the MPAS-A dycore does not require any specific range for lonCell".

mgduda (changes requested):

src/dynamics/mpas/dyn_grid.F90
gdicker1: Thank you this does correspond to one of the [SHOULD](https://github.com/ESCOMP/CAM/wiki/CAM-Coding-Standards#should-1) standards. Addressed in [005e7f4](https://github.com/ESCOMP/CAM/pull/1095/commits/005e7f45cdb8ec604dd707fc2af56f9351f7d25f)

src/dynamics/mpas/dyn_grid.F90
gdicker1: Addressed in [1325911](https://github.com/ESCOMP/CAM/pull/1095/commits/13259119275a64900e771931df912c5a4ea01982)

mgduda (approved):
Approved by inspection.

jtruesdal (approved):

PR #1096, opened 2024-07-19
cam6_4_025: Aerosol deposition fluxes

### LINKED ISSUE #1060: Update modal dust partitioning for wet and dry deposition

The current dry and wet deposition code crudely maps the modal aerosol dust in MAM onto the 4 bulk size bins. A code by Mark Flanner is using erf functions to partition the lognormal distribution into the 4 sectional bins, which maps of the modal size distributions more precisely to the size bins required by the land and ocean models. 

@tilmes 

Yes

fvitt:
This repartitions the modal dust deposition fluxes into 4 bulk bins for passing to the surface models.  The aerosol fluxes code was refactored in a generalized way which can easily be expanded for other aerosol representations, such as CARMA.

Closes #1060


fvitt:
This has yet to be tested and evaluated.

src/chemistry/aerosol/aero_deposition_cam.F90
```diff
@@ -0,0 +1,314 @@
...
+
+      ! in rare cases, integrated deposition tendency is upward
+      if (cam_out%bcphidry(i) .lt. 0._r8) cam_out%bcphidry(i) = 0._r8
+      if (cam_out%ocphidry(i) .lt. 0._r8) cam_out%ocphidry(i) = 0._r8
+      if (cam_out%bcphodry(i) .lt. 0._r8) cam_out%bcphodry(i) = 0._r8
+      if (cam_out%ocphodry(i) .lt. 0._r8) cam_out%ocphodry(i) = 0._r8
+      if (cam_out%dstdry1(i)  .lt. 0._r8) cam_out%dstdry1(i)  = 0._r8
+      if (cam_out%dstdry2(i)  .lt. 0._r8) cam_out%dstdry2(i)  = 0._r8
+      if (cam_out%dstdry3(i)  .lt. 0._r8) cam_out%dstdry3(i)  = 0._r8
+      if (cam_out%dstdry4(i)  .lt. 0._r8) cam_out%dstdry4(i)  = 0._r8
```
cacraigucar, line 308: CAM guidelines say "should" use `<` instead of `.lt.` .  This is at your discretion since it is a "should" and not a "must"

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -665,4 +675,96 @@ function bin_name(self, list_ndx,  bin_ndx) result(name)
...
+       Mtotal = 0._r8
+       mm = self%indexer(m,0)
+       Ntot = dep_fluxes(mm) ! #/m2
+
+       has_dust = .false.
+
+       do l = 1,self%nspecies(m)
+          mm = self%indexer(m,l)
+          call self%get(m,l, spectype=spectype, density=dns_dst) ! kg/m3
+          if (spectype==bulk_type) then
```
cacraigucar, line 715: I'm not sure if it is doable (an error or not) to capture if no matches occur with `spectype==bulk_type`.  I'm thinking of whether it is worth trying to capture a typo or completely incorrect name for `bulk_type`

cacraigucar (changes requested):

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
fvitt: I could probably implement an error code and error string that get returned by the subroutine if you think it is worth while.  As it is now, if the specified `bulk_type` is not valid then the fluxes returned are zeros.

cacraigucar: I agree that the fluxes will be zero, but I'm not sure if it would be readily apparent to a user that they had a typo or invalid name as zero fluxes might be pretty subtle in a run.  It really is your call - I am just pointing this out for your consideration.

cacraigucar (approved):

src/chemistry/aerosol/aero_deposition_cam.F90
```diff
@@ -0,0 +1,314 @@
...
+
+      ! black carbon fluxes
+      do ispec=1,bcphi_cnt
+         cam_out%bcphidry(i) = cam_out%bcphidry(i) &
+                             + (aerdepdryis(i,bcphi_ndx(ispec))+aerdepdrycw(i,bcphi_ndx(ispec)))
+      enddo
+      do ispec=1,bcpho_cnt
+         cam_out%bcphodry(i) = cam_out%bcphodry(i) &
+                             + (aerdepdryis(i,bcpho_ndx(ispec))+aerdepdrycw(i,bcpho_ndx(ispec)))
+      enddo
```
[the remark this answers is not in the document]
fvitt: Added comments

src/chemistry/aerosol/aero_deposition_cam.F90
fvitt: Done

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
fvitt: Added error checking

PR #1100, opened 2024-07-24
cam6_4_016: Modify RRTMGP interface for MT configurations.

### LINKED ISSUE #1063: Possible modification to RRTMG-P for ~80km top model
### What happened?

See issue presented at CAM7 meeting (slide 18) by @JulioTBacmeister :

https://docs.google.com/presentation/d/1U51QGheey-l6AyDh-YgQ6rkd8ASClmaLv20C2XZBi-A/edit#slide=id.g2e748601ada_0_6

Ask @JulioTBacmeister

cam6_3_160

CISL machine (e.g. cheyenne)

Intel

_No response_

Any CAM SE can do this

### Extra info

Don't know who is going to submit PR at this point in time

### LINKED ISSUE #1097: HEMCO reference in .gitmodules is a branch not a tag.
### What happened?

https://github.com/ESCOMP/CAM/blob/cam_development/.gitmodules#L60
Refers to a branch.  It should refer to the tag hemco-cesm1_2_1_hemco3_6_3_cesm_rme01

NA

cam6_4_014

CISL machine (e.g. cheyenne)

Intel

NA

Any CAM SE can do this

### Extra info

_No response_

brian-eaton:
Modify the RRTMGP interface for the special case when the minimum valid pressure for RRTMGP (1 Pa) falls in the top model layer (as it does in CAM's 93 level MT configuration).  The modification is to use the "extra layer" code path, and add a very thin extra layer just below 1 Pa.  The algorithm to calculate the midpoint pressure in the "extra layer" has changed from the original (which assumed a model top at 0 Pa).  Hence the change affects answers for the low top model configurations (LT and cam6) as well as the MT configuration.

Note that this modification is still being tested for scientific validity in the MT configuration.

Issue #1097 - HEMCO reference in .gitmodules is a branch not a tag.
- Modify .gitmodules to resolve #1097

Closes #1097 
Closes #1063

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,7 +57,7 @@
 [submodule "hemco"]
 	path = src/hemco
 	url = https://github.com/ESCOMP/HEMCO_CESM.git
-        fxtag = hemco-cesm1_2_1_hemco3_6_3_cesm_rme
+        fxtag = hemco-cesm1_2_1_hemco3_6_3_cesm_rme01
         fxrequired = AlwaysRequired
 	fxDONOTUSEurl = https://github.com/ESCOMP/HEMCO_CESM.git
 
--- a/src/physics/rrtmgp/radiation.F90
+++ b/src/physics/rrtmgp/radiation.F90
@@ -453,6 +453,7 @@ subroutine radiation_init(pbuf2d)
    ! pressure interfaces below 1 Pa.  When the entire model atmosphere is
    ! below 1 Pa then an extra layer is added to the top of the model for
    ! the purpose of the radiation calculation.
+
    nlay = count( pref_edge(:) > 1._r8 ) ! pascals (0.01 mbar)
 
    if (nlay == pverp) then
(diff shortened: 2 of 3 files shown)


cacraigucar (approved):
Did you make any changes beyond what @brianpm provided?  If so, do we need a quick review from him?

brian-eaton:
No.  This PR currently has just what @brianpm provided.

brian-eaton:
@brianpm, @PeterHjortLauritzen, @adamrher .  Unfortunately these mods change answers for all RRTMGP runs.  The calculation of the midpoint pressure in the extra layer is different for LT versions than it was before.  The original midpoint pressure calculation in the extra layer ignored the fact that the top interface was placed at 1 Pa (effectively assumed it was 0), so the new calculation is more accurate, but different.  Should we just go with this and say we expect all RRTMGP runs to have answer changes?

brianpm:
Hmmmmm. I didn't think of that. I guess I'm leaning toward going ahead with the change. Interested to hear if @PeterHjortLauritzen or @adamrher have a different thought.

brian-eaton:
Thanks!  I agree that it's best to move ahead with this change.  The new calculation for the midpoint pressure of the "extra layer" is more robust than the original.  Also, this doesn't change answers when WACCM is run with RRTMGP.  Only the low top model configurations are affected.

Issue #651, opened 2022-09-02
Move the dust emission source function and global tuning factor from CAM to CTSM


dmleung:
### Issue Type

Code Clean-up

### Issue Description

Hi this is Danny, a PhD student at UCLA and NCAR working with @jfkok and @dlawrenncar on CTSM dust modeling development. Here we request to move the CAM dust soil erodibility map and global tuning factor for the dust scheme previously read in and manipulated in CAM to CTSM. This issue is also posted on CTSM github https://github.com/ESCOMP/CTSM/issues/1836.

In issue https://github.com/ESCOMP/CTSM/issues/1604 we proposed to add a new dust emission scheme improving the dust mobilization parameterizations in CLM. Although a new scheme will be added as in PR https://github.com/ESCOMP/CTSM/pull/1712, CLM people decided to keep the existing emission scheme by Zender (2003). One issue is that the (time-independent) soil erodibility source function employed by Zender's scheme has always been read in and applied to the CAM emission flux instead of to the CLM emission flux, likely due to historical reasons. However, conceptually, how soil erodibility and land-surface characteristics affect the emission strength should be part of the land processes, and we believe it should be read in and applied in CLM instead of CAM. CLM has a variable for the Zender source function called mbl_bsn_fct, but it has always been set as 1 as a place holder and it is conceptually the same variable as the soil_erod in CAM. Here we request moving the soil erodibility map from CAM to CLM into mbl_bsn_fct, coming as part of either the surface dataset or the stream file. 

This small modification will make the CLM dust emission scheme more conceptually complete, and eliminate conceptually overlapping variables among CLM and CAM. 

Also, there is a global tuning factor (with CAM namelist variable dust_emis_fact) that controls the magnitude of the global total dust emission. Due to the same reason we think it should also be moved to CLM.

In CAM, we would like to modify the atm dust module and the namelist definition/defaults. Several Fortran variables, namelist variables, and output fields are proposed to be removed.

@dmleung proposed the modification. @dlawrenncar, @lkemmons, @L3atm, and software engineers including @ekluzek and @fvitt agreed about the decision.
@dmleung will work on this issue in both CLM and CAM with help from @ekluzek and @fvitt. 

I know CAM people will tune aerosols (including dust) often when they have new CAM versions. They might need to hereafter tune it in CLM instead of CAM. @cecilehannay is informed.

Thank you,
Danny

Files to be changed:
clm/src/biogeochem/DUSTMod.F90
[...]

Issue #141, opened 2020-05-28
MAM5 and updates to dust emis


fvitt:
Add a mode for stratospheric sulfate to the modal aerosol model.
Updates to dust emissions.

Francis, this is my F90 codes and BLD codes to setup MAM5 configuration

/glade/u/home/zke3/scratch/cases/edit8

The F90 codes and chem_mech.in are in CAM, configure codes are in BLD folder, chem_mech.in and namelist files for WACCM are in WACCM folder. The F90 files are used for both WACCM and CAM. Two configurations I added in is trop_mam5 and waccm_tsmlt_mam5. If you need the codes to separate with the convection changes, I also have that older version. That is based on “release-cesm2.1.2” version.

Thanks 

Ziming Ke

Francis, here is our code changes,

/glade/u/home/longei/CAM-MAM5-codes/wet_removal_n02_cam6_2_020

 

and a list of the routines we modified.

DUSTMod.F90

dust_model.F90

modal_aer_opt.F90

Thanks very much.

All the best & Be well,

Longlei

fvitt:
Not required for cesm2.2.  However it is desired get the changes into cam_development soon after code freeze.

fvitt:
Email from Longlei (ll859@cornell.edu):

Below is the directory where the source code is recorded. 

/glade/work/longei/Cheyenne/TESTS/NCAR2dps/CAM6_2dps

 

As said before, I made the old dust emission scheme optional. I defined a new namelist variable named as brift_dps, where brift stands for BRIttle Fragmentation Theory, and dps stands for Dust Production Scheme.

 

By default, the model will automatically use the new dust emission scheme. If one wants to go back the old dust scheme, he/she should set the following in BOTH user_nl_cam and user_nl_clm.

 

brift_dps = .true.

PR #1104, opened 2024-07-24
cam6_4_028: Dust emissions

fvitt:
Mods needed for the new dust scheme 

closes https://github.com/ESCOMP/CAM/issues/651
closes https://github.com/ESCOMP/CAM/issues/141


fvitt:
@ekluzek This is the new CAM PR for the new dust emission scheme coming into CTSM.  

We need to think about how to coordinate namelist settings when the new scheme is turned on...

ekluzek (changes requested):
@fvitt thanks for working on this. I suppose this was just back porting the previous PR that was done before the dust emission namelist was worked out in CMEPS.

The namelist read needs to be redone so that it uses the dust emission settings in the drv_flds_in namelist file and uses the components/cmeps/cesm/nuopc_cap_share/shr_dust_emis_mod.F90 module to determine if CAM needs to modify dust emissions from CLM with the soil erodibility file.

What needs to happen in CAM will be similar to what happens in the CTSM branch being pointed to.

ekluzek (changes requested):
@fvitt thanks for the updates. I ask for a couple changes.

I think you also need to add the new fields to the namelist_defintion file for drv_flds, unless I'm missing something. And you need to add a call for the shr_dust_emis_readnl() (CTSM has this in lnd_import_export.F90).

ekluzek:
@fvitt note, the CTSM tag that allows CAM and CTSM to work together on dust emission settings is complete and it's ctsm5.2.017. 

It doesn't yet allow turning on Leung_2023, but it uses the new XML variable LND_SETS_DUST_EMIS_DRV_FLDS to determine if CTSM or CAM are doing the dust settings and uses the dust_emis_inparm namelist in drv_flds_in to determine the dust settings and when Zender_2003 is on, to determine if CAM or CTSM will be setting the Zender soil erodibility file.

fvitt:
@ekluzek Is there a CTSM tag for the Leung_2023 scheme, or is this still in progress?

ekluzek:
@fvitt the CTSM tag with Leung is still in process, but it's

https://github.com/ESCOMP/CTSM/pull/1897

and will be ctsm5.2.019. It is almost completed. I plan to have it completed over the weekend before Derecho goes down on Monday.

ekluzek:
Ok the needed CTSM tag is done ctsm5.2.019

ekluzek:
@fvitt the thing that is missing here is that when you are running CAM you need CAM to specify the namelist settings and NOT have CLM modify any of them. The problem seems to be having them both trying to mess with them at the same time.

So you need to set the XML variable LND_SETS_DUST_EMIS_DRV_FLDS to false

```
./xmlchange LND_SETS_DUST_EMIS_DRV_FLDS=FALSE
```

I'm not sure where that should go in CAM, but for now it should always be done for CAM compsets that use CLM.

fvitt:
@ekluzek 
These problems occur with `LND_SETS_DUST_EMIS_DRV_FLDS=FALSE`.

ekluzek:
@fvitt thanks for pointing that out. I'll try checking out this branch and reproduce it to figure out what's wrong.

fvitt:
@ekluzek I just pushed up my latest changes.  Try to build namelists for FHIST and FLTHIST compsets.  Thanks.

ekluzek:
@fvitt thanks for pointing this out. I've replicated the problem with your branch. 

I also see what's going on, and why I didn't think of this issue before. I'll have to think about how to fix it though. I'll add an issue on the CTSM side regarding this. We'll include it in an upcoming CTSM tag and I'll update this when we have that.

A workaround that I verified is you can add this to your user_nl_clm for CAM:

``` fortran
use_prigent_roughness = .false.
```

You could add that temporarily while we get a CTSM tag with a fix and then remove it when we get the CTSM in place for CAM to use.

fvitt:
> @fvitt thanks for pointing this out. I've replicated the problem with your branch.
> 
> I also see what's going on, and why I didn't think of this issue before. I'll have to think about how to fix it though. I'll add an issue on the CTSM side regarding this. We'll include it in an upcoming CTSM tag and I'll update this when we have that.
> 
> A workaround that I verified is you can add this to your user_nl_clm for CAM:
> 
> ```fortran
> use_prigent_roughness = .false.
> ```
> 
> You could add that temporarily while we get a CTSM tag with a fix and then remove it when we get the CTSM in place for CAM to use.

The problem with this is I don't see a way to set this namelist in `user_nl_clm` out-of-the-box for all the CAM compsets, which what is needed for the regression tests.

ekluzek:
@fvitt yes it looks like CAM is using test mod directories as completely independent of each other and NOT using the include_user_mods feature to pull in something from a parent test-mod. Maybe that's done on purpose? Or CAM just wasn't aware of this ability?

Using the include ability you could add for example a 

cime_config/testdefs/testmods_dirs/cam/default

directory with the user_nl_clm file with the one modification needed. And then add
include_user_mods
file to each test mod directory that has...

../default

as it's contents. Which is relatively contained and doable, and it would allow you to test before a tag is in place.

That might be more than you want to do here, and you'd definitely want to check in with CAM folks to see if it's even desired to do this. But, it's an option...

ekluzek:
@fvitt I thought of one other workaround that might be easier to try. You could add

"-ignore_warnings"

to the XML variable CLM_BLDNML_OPTS (you'd probably want to do it using the --append option). That should work, but I haven't tried it.

ekluzek:
@fvitt try hash 3f3541b37b7ff9abe37050c98575b61b3702e688 in the CTSM PR https://github.com/ESCOMP/CTSM/pull/2699

on a few of your cases and make sure it's working for you. I successfully ran:

SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s_nochem.

and think that means it should work for everything. But, of course in practice there might be a compset with unexpected problems.

fvitt:
@ekluzek 
With that hash, I am getting this error:
```
 err=ERROR : CLM build-namelist::CLMBuildNamelist::setup_logic_megan() : megan_specifier or megan_factors_file defined, but megan option NOT set
```

The `outfrq9s_nochem` user_mod is not a good choice for testing.  
Try these tests
```
 SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s
 SMS_Ln9.ne30pg3_ne30pg3_mg17.FLTHIST.derecho_intel.cam-outfrq9s
```

ekluzek:
@fvitt thanks for trying that out and giving me some new tests to try. I see the next thing that needs to be done now.

This also shows that the current path here is fragile, but I think it's necessary for now. I think right now it means that we'll have this  LND_SETS_DUST_EMIS_DRV_FLDS  variable and logic around for a while, but we'll want to get it to always FALSE as quickly as we can. And then we'll be able to remove it. The fragility goes away once we can get it to FALSE for coupled to CAM.

bld/build-namelist
```diff
@@ -569,6 +569,14 @@ if ($cfg->get('debug')) {
 my $prescribe_aerosols = $TRUE;
 if ($simple_phys) {$prescribe_aerosols = $FALSE;}
 
+# Dust emissions scheme (all configurations for now)
+add_default($nl, 'dust_emis_method');
+if ( $nl->get_value('dust_emis_method') =~ /Zender/ ) {
+    add_default($nl, 'zender_soil_erod_source');
+}
+my $soil_erod_atm = $FALSE;
+if ($nl->get_value('dust_emis_method') eq 'atm') {$soil_erod_atm = $TRUE;}
```
ekluzek, line 578: This should check zender_soil_erod_source rather than dust_emis_method. And since it's not always set should likely be restructured so that it's set appropriately even if zender_soil_erod_source wasn't being set (because it's Leung_2024 and not Zender_2003).

I'll suggest a potential restructuring...

ekluzek, line 578:
```suggestion
add_default($nl, 'dust_emis_method');
my $soil_erod_atm = $FALSE;
if ( $nl->get_value('dust_emis_method') =~ /Zender/ ) {
    add_default($nl, 'zender_soil_erod_source');
    if ($nl->get_value('zender_soil_erod_source') eq 'atm') {$soil_erod_atm = $TRUE;}
}
```

ekluzek:
@fvitt I have another updated CTSM version for you to try out:

786252bb012172e53ced93b40b9730d11f8d6efe

I ran

SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s_nochem

with it and it passed. Note that Danny noticed a problem with your sandbox that gave problems for him that you likely need to handle before this works here (it's all given in comments above).

fvitt:
> @fvitt I have another updated CTSM version for you to try out:
> 
> 786252bb012172e53ced93b40b9730d11f8d6efe
> 
> I ran
> 
> SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s_nochem
> 
> with it and it passed. Note that Danny noticed a problem with your sandbox that gave problems for him that you likely need to handle before this works here (it's all given in comments above).

I tried this version of CTSM.  This allows CAM to set the namelist settings.  However, with `dust_emis_method = 'Leung_2023'`, the dust emission fluxes seem to be all zeros.

ekluzek:
> > @fvitt I have another updated CTSM version for you to try out:
> > 786252bb012172e53ced93b40b9730d11f8d6efe
> > I ran
> > SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s_nochem
> > with it and it passed. Note that Danny noticed a problem with your sandbox that gave problems for him that you likely need to handle before this works here (it's all given in comments above).
> 
> I tried this version of CTSM. This allows CAM to set the namelist settings. However, with `dust_emis_method = 'Leung_2023'`, the dust emission fluxes seem to be all zeros.

Hmmm. That's not something I can explain by any of the code changes that have happened. Danny has also been doing his own F cases while we've gone along.

Before diving into this -- can you let me know about what period it's all zeros? Is that the first time-step for example? If you run longer does it become non-zero after awhile? Zero is valid, it would just be unusual over the entire globe, but maybe possible, but over time it obviously needs to be non-zero.

fvitt:
> > > @fvitt I have another updated CTSM version for you to try out:
> > > 786252bb012172e53ced93b40b9730d11f8d6efe
> > > I ran
> > > SMS_Ln9.f19_f19_mg17.FHIST.derecho_intel.cam-outfrq9s_nochem
> > > with it and it passed. Note that Danny noticed a problem with your sandbox that gave problems for him that you likely need to handle before this works here (it's all given in comments above).
> > 
> > 
> > I tried this version of CTSM. This allows CAM to set the namelist settings. However, with `dust_emis_method = 'Leung_2023'`, the dust emission fluxes seem to be all zeros.
> 
> Hmmm. That's not something I can explain by any of the code changes that have happened. Danny has also been doing his own F cases while we've gone along.
> 
> Before diving into this -- can you let me know about what period it's all zeros? Is that the first time-step for example? If you run longer does it become non-zero after awhile? Zero is valid, it would just be unusual over the entire globe, but maybe possible, but over time it obviously needs to be non-zero.

I ran for a day and output fluxes each time step.  Fluxes seem to zeros all time steps everywhere.

fvitt:
@ekluzek I fixed a bug on the CAM side.  The Leung dust emissions schemes seems to be working now.

.gitmodules
```diff
@@ -178,8 +178,8 @@ fxDONOTUSEurl = https://github.com/NCAR/ParallelIO
 
 [submodule "clm"]
 path = components/clm
-url = https://github.com/ESCOMP/CTSM
-fxtag = ctsm5.2.009
+url = https://github.com/ekluzek/CTSM
+fxtag = 786252bb0
```
cacraigucar, line 182: Just a reminder (more for me than you) to update this to the official CTSM tag when it is ready

cacraigucar (changes requested):

.gitmodules
ekluzek: This can now be replaced with ctsm5.2.027

bld/namelist_files/namelist_definition.xml
```diff
@@ -7698,6 +7698,21 @@ List of fluxes needed by the CARMA model, from CLM to CAM.
...
+Either the Zender 2003 scheme or the Leung 2023 scheme.
+Default: set by build-namelist.
+</entry>
+
+<entry id="zender_soil_erod_source" type="char*80" category="dust_emissions"
+       group="dust_emis_inparm" valid_values="none,lnd,atm" >
+Option only applying for the Zender_2003 method for whether the soil erodibility
+file is handled in the active LAND model or in the ATM model.
+(only used when dust_emis_method is Zender_2003)
+Default: set by build-namelist.
```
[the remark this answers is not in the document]
ekluzek: @cacraigucar this will be a transition thing that we need to do. Once, the older CAM physics versions that use Zender are vetted for lnd to work correctly for them -- we can change it to lnd. We need to make sure that the difference between lnd and atm is small. It's likely greater than roundoff, but shouldn't be much, unless there's a bug in CTSM.

.gitmodules
fvitt: Thanks @ekluzek 
Updating to from ctsm5.2.009 to ctsm5.2.027 changes answers for the F compsets, even when Zender scheme is used.  Is this expected?  I see several namelist changes in lnd_in.

ekluzek: Yes, this is expected. There's several rounds of answer changes between 009 and 027. I can map out what they are if you like and that would help.

Is it OK for your CAM tag to have CLM answer changes in it?

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2532,6 +2532,11 @@
 <dust_emis_fact hgrid="0.47x0.63" offline_dyn="1" phys="cam6"    ver="chem">0.9D0</dust_emis_fact>
 <dust_emis_fact hgrid="0.47x0.63" offline_dyn="1" phys="cam7"    ver="chem">0.9D0</dust_emis_fact>
 
+<!-- dust emissions method -->
+<dust_emis_method>Zender_2003</dust_emis_method>
+<dust_emis_method phys='cam7'>Leung_2023</dust_emis_method>
+<zender_soil_erod_source>atm</zender_soil_erod_source>
```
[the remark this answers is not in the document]
fvitt: Default is now `Zender_2003` for all configurations.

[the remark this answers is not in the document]
ekluzek: @cacraigucar I assume that means there will be a follow on CAM tag where Leung is turned on for CAM7? Or will we just be in experimental mode for awhile and then change the default much later? Just curious how this is going to go...

.gitmodules
cacraigucar: We typically update answer changing externals for CAM runs in their own tag, so I'd like to pull it in separate.  @fvitt - would you be willing to do this, or do we need to do this on the AMP side?

cacraigucar: The exception to this is if @fvitt has independently already run the regression tests with his changes and they are BFB except for the test(s) which use the dust.  If this has been done, he can document in the ChangeLog that he has independently tested that "the code changes were independently tested and were non-answer changing for all but the dust configurations.  All answer changes to non-dust runs are due to updating CLM".  If these tests haven't been run, then it probably makes sense to just split it into two CAM tags.

bld/namelist_files/namelist_defaults_cam.xml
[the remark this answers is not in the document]
cacraigucar: @ekluzek - When(if) the scientists tell us that they want to change the default setting, we will do so.  This will be part of the tunings that will be happening and I suspect that we'll be adjusting quite a few settings/defaults, not just this one.

cacraigucar (approved):

.gitmodules
ekluzek: Here's the list of things I expected to affect CAM F compsets in a substantial way:

- ctsm5.2.016 -- changes answers for clm6_0 for crop grid cells
- ctsm5.2.020 -- changes answers for all physics options for MEGAN BGVOC's which will affect CAM-Chem simulations
- ctsm5.2.026 -- change answers for clm6_0 over urban grid cells

There's a few others that still might affect CAM F compsets, but I expect them to be much smaller. Go to the bottom of this wiki page to see a full description of CTSM answer changing tags:

https://github.com/ESCOMP/CTSM/wiki/Answer-changing-tags

bld/build-namelist
ekluzek: This was done.

ekluzek: Handled.

src/chemistry/bulk_aero/dust_model.F90
```diff
@@ -69,14 +73,26 @@ subroutine dust_readnl(nlfile)
...
+    end if
+    call mpi_bcast(dust_emis_fact, 1, mpi_real8, masterprocid, mpicom, ierr)
+    if (ierr/=mpi_success) then
+       call endrun(subname//' MPI_BCAST ERROR: dust_emis_fact')
+    end if
+
+    call shr_dust_emis_readnl(mpicom, 'drv_flds_in')
+
+    if (masterproc) then
+       write(iulog,*) subname,': soil_erod_file = ',trim(soil_erod_file)
```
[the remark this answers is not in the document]
ekluzek: Handled.

src/chemistry/bulk_aero/dust_model.F90
```diff
@@ -69,14 +73,26 @@ subroutine dust_readnl(nlfile)
...
+       write(iulog,*) subname,': dust_emis_fact = ',dust_emis_fact
+       write(iulog,*) subname,': is_dust_emis_zender : ',is_dust_emis_zender()
```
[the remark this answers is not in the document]
ekluzek: Nicely handled.

src/chemistry/bulk_aero/dust_model.F90
```diff
@@ -69,14 +73,26 @@ subroutine dust_readnl(nlfile)
...
        close(unitn)
-       call freeunit(unitn)
     end if
 
-#ifdef SPMD
     ! Broadcast namelist variables
-    call mpibcast(dust_emis_fact, 1,                   mpir8,   0, mpicom)
-    call mpibcast(soil_erod_file, len(soil_erod_file), mpichar, 0, mpicom)
-#endif
+    call mpi_bcast(soil_erod_file, len(soil_erod_file), mpi_character, masterprocid, mpicom, ierr)
```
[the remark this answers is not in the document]
ekluzek: Handled.

src/chemistry/modal_aero/dust_model.F90
```diff
@@ -68,14 +71,26 @@ subroutine dust_readnl(nlfile)
...
        close(unitn)
-       call freeunit(unitn)
     end if
 
-#ifdef SPMD
     ! Broadcast namelist variables
-    call mpibcast(dust_emis_fact, 1,                   mpir8,   0, mpicom)
-    call mpibcast(soil_erod_file, len(soil_erod_file), mpichar, 0, mpicom)
-#endif
+    call mpi_bcast(soil_erod_file, len(soil_erod_file), mpi_character, masterprocid, mpicom, ierr)
```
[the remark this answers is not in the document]
ekluzek: Handled.

ekluzek (approved):
All of my suggestions were appropriately handled. So this is good to go from my perspective. Thanks for working on this and getting it done @fvitt!

cime_config/config_compsets.xml
```diff
@@ -906,7 +906,7 @@
 
     <entry id="LND_SETS_DUST_EMIS_DRV_FLDS">
       <values match="first">
-	<value compset="_CAM.*_CLM">FALSE</value>
+	<value compset="_CAM7.*_CLM6">FALSE</value>
```
ekluzek, line 909: @fvitt I was concerned about this at first in case it meant that the CAM Zender soil erodibility files were NOT being used. But, it looks like that is NOT the case. So this should be fine. I'm looking at it more closely. I'm glad you figured this workaround out.

ekluzek: This is also probably fine since longer term we want to always make this TRUE and even remove it entirely. So that just brings us closer to the long term solution.

ekluzek: The only bad thing here is that it means that the way you change dust emission settings will be different for the CAM7 with CAM6 than for anything else. But, it's unlikely that people will be messing with the dust settings for previous CAM physics.

bld/namelist_files/use_cases/sd_waccm_ma_cam4.xml
```diff
@@ -79,7 +79,10 @@
 
 <!-- turn on clm MEGAN VOC emis -->
 <megan_mapped_emisfctrs>.false.</megan_mapped_emisfctrs>
-<megan_specifier>'CH2O = formaldehyde','CO = carbon_monoxide'</megan_specifier>
+<megan_specifier>
```
ekluzek, line 82: I'm surprised you had to do this. But, of course it's a change that's perfectly fine to do. I'll see if I can look into why you had to change it...

PR #1114, opened 2024-07-29
cam6_4_070: Pertlim fix for MPAS

### LINKED ISSUE #1109: MPAS-A Pertlim usage is only implemented for ideal initial states
### What happened?

Adding initial state perturbations is not implemented when reading initial values via ncdata. 

Specify an initial values dataset via ncdata, run 5 day tests with pertlim=0, and with pertlim=1.e-14, and then compare  the results to see that they are equal.

cam6_3_160

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

Correcting this to add perturbations to the initial theta values only requires a few lines of code. The PR to add the changes will be implemented for the most recent cam tag: cam6_4_016

patcal:
Closes  #1109


jtruesdal (approved):
This mods are fine. @cacraigucar I was surprised that there aren't pertlim regression tests since ensembles are an important part of climate analysis.  I'll bring that up in our SE meeting and see if we need an issue to provide testing for all supported dycores.  Since we don't have testing currently in place please run your own test to make sure that the pertlim namelist addition is working correctly and document this in your Changelog.

cacraigucar:
@patcal - Could you please provide a test that you ran with this new pertlim that we can use as a basis of a CAM regression test?

patcal:
The test I ran was for a user-defined MPAS grid and user-generated input
files, so I don't think it is feasible for a standard regression test. I
have not checked
to see if there is a functional implementation of a uniform low-res MPAS
grid in
the current tag. If there is, then I can set up a test using it.

cacraigucar:
@patcal - Have you made any progress on creating a regression test?  This PR is stalled until we get one made

nusbaume:
@cacraigucar I have added the MPAS pertlim regression test that @patcal developed, so I think this PR is ready to be added to the CAM PR queue.  Thanks!

PR #1101, opened 2024-07-24
improved fix for rh write performance

jedwards4b:
Fixes issue #1042.   The original fix in #1043 was not merged - testing showed problems with sathist files.   Those problems have been addressed here.   
```
[jedwards@izumi ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist.20240723_111303_e1nxwq]$ cat TestStatus
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist CREATE_NEWCASE
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist XML
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist SETUP
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist SHAREDLIB_BUILD time=28
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist MODEL_BUILD time=770
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist SUBMIT
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist RUN time=159
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist COMPARE_base_rest
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist MEMLEAK insufficient data for memleak test
PASS ERC_D_Ln9.f10_f10_mg37.QPC6.izumi_nag.cam-outfrq3s_cospsathist SHORT_TERM_ARCHIVER
```

PR #1122, opened 2024-08-09
fix issue #1108, merge PR#1101,...

### LINKED ISSUE #1108: More robust logic in gw_drag.F90 when deep_scheme='off'
### What happened?

When I run a configuration with deep_scheme='off' and use_gw_convect_dp=.true., it errors out as gw_drag.F90 is doing a pbuf_get_index for a ZM pbuf variable that doesn't exist. So I just run with use_gw_convect_dp=.false, which is fine. However, with the new moving mountains gw sources that came in recently, it's using that same pbuf index ```ttend_dp_idx``` to perform a pbuf_get_field call in the moving mountains code block. If use_gw_convect_dp=.false, then that index retains its initialized value of -1 and the model errors out in the pbuf_get_field call.

I propose:

(1) Move the pbuf_get_index outside the ```use_gw_convect_dp``` conditional 
(2) Create logic to only call pbuf_get_index if deep_scheme /= 'off', otherwise ```ttend_dp_idx``` retains its initialized value of -1
(3) Put a kill switch in if ```use_gw_convect_dp```=.true. and ```ttend_dp_idx```= -1
(4) Add logic around the moving mountain code block ```use_gw_movmtn_pbl``` to only perform the pbuf_get_field when  ```ttend_dp_idx```/= -1, else fill an array of the same size as that ZM pbuf variable with zeros, and pass that into the subroutine call ```gw_movmtn_src``` (the ZM array is not actually used by the moving mountains at this time, so making it zero doesn't change answers).

see above.

cam6_4_015

CISL machine (e.g. cheyenne)

Intel

_No response_

Any CAM SE can do this

### Extra info

@JulioTBacmeister

brian-eaton:
Issue #1108 - More robust logic in gw_drag.F90 when deep_scheme='off'
- Modify `build-namelist` to set `use_gw_convect_dp=.false.` when `deep_scheme='off'`.  In `gw_drag::gw_tend` check whether field `TTEND_DP` is in the pbuf.  If so then associate the `ttend_dp` pointer.  If not then allocate the `ttend_dp` pointer and set to zero.

PR #1101 - improved fix for rh write performance
- reorder output fields.  Merge Jim's PR into this one.

fix #1108 
fix #1101


brian-eaton:
Don't know why this PR thinks 34 files have changed instead of only 4.  Closing and starting over.

PR #1131, opened 2024-08-20
cam6_4_031: fix issues #1108, #1106, #1058, #1051, #1050; merge PR#1101

### LINKED ISSUE #1050: Remove CAM3 as a compset or configure option
### Issue Type

Code Clean-up

### Issue Description

After PR #890 CAM development is no longer able to replicate CAM3 physics regardless of settings, and thus CAM3 is no longer considered a supported physics configuration.  Given this, it would be good to officially remove CAM3 from the list of possible compsets, as well as a possible physics package option for CESM3.

This would likely need to be done after all CAM3 regression tests have also been removed, as described in issue #258. 

### Will this change answers?

No

Any CAM SE can do this

### LINKED ISSUE #1051: Bad logic in SE dycore "interpolate_vector" subroutines
### What happened?

In the `interpolate_vector2d` and `interpolate_vector3d` subroutines located in the `src/dynamics/se/dycore/interpolate_mod.F90` source file there are checks that look like this:

```
if (npts==np) then
   interp => interp_p
else if (npts==np) then
   call endrun('interpolate_vector3d: Error in interpolate_vector(): input must be on velocity grid')
endif
```

However, with this setup the `endrun` call will never be made, because the `else if` check is performing the exact same logic as the original if-statement.  

I have no idea if this check is actually necessary or important, but if it is then it would be good to change it so that the `endrun` is actually called when expected.

Look at the code?

cam6_3_161

Other (please explain below)

Other (please specify below)

_No response_

Any CAM SE can do this

### Extra info

No case was run to demonstrate this problem, I simply found it when looking at the code.

### LINKED ISSUE #1058: Remove unused pbuf variable smaw
### Issue Type

Code Clean-up

### Issue Description

It appears that smaw is unused in the code (output from caleddy inside eddy_diff.F90)   Confirm this is true, and if so, remove

### Will this change answers?

No

Any CAM-SE can do this

### LINKED ISSUE #1106: Report an error if a user uses --model_top with anything other than cam7

If a user attempts to set --model_top MT with --phys cam5, they get an undecipherable error.  `configure` should have a check to prevent the user from specifying this combo.

The error:

```
Traceback (most recent call last):
  File "/glade/work/juliob/cam6_4_005_jtb/cases/c64_005_ne30pg3_FMTHIST_C5phys_x02/./case.build", line 267, in <module>
    _main_func(__doc__)
  File "/glade/work/juliob/cam6_4_005_jtb/cases/c64_005_ne30pg3_FMTHIST_C5phys_x02/./case.build", line 251, in _main_func
    success = build.case_build(
              ^^^^^^^^^^^^^^^^^
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/build.py", line 1327, in case_build
    return run_and_log_case_status(functor, cb, caseroot=caseroot)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/utils.py", line 2482, in run_and_log_case_status
    rv = func()
         ^^^^^^
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/build.py", line 1311, in <lambda>
    functor = lambda: _case_build_impl(
                      ^^^^^^^^^^^^^^^^^
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/build.py", line 1184, in _case_build_impl
    sharedpath = _build_checks(
                 ^^^^^^^^^^^^^^
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/build.py", line 721, in _build_checks
    case.create_namelists()
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/case/preview_namelists.py", line 90, in create_namelists
    import_and_run_sub_or_cmd(
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/utils.py", line 668, in import_and_run_sub_or_cmd
    raise e1 from None
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/utils.py", line 664, in import_and_run_sub_or_cmd
    run_sub_or_cmd(
  File "/glade/work/juliob/cam6_4_005_jtb/cime/CIME/utils.py", line 709, in run_sub_or_cmd
    getattr(mod, subname)(*subargs)
  File "/glade/work/juliob/cam6_4_005_jtb/cime_config/buildnml", line 81, in buildnml
    raise RuntimeError("CAM's 'buildcpp' script failed to run properly.")
RuntimeError: CAM's 'buildcpp' script failed to run properly.
`````

_No response_

No

Any CAM SE can do this

### LINKED ISSUE #1108: More robust logic in gw_drag.F90 when deep_scheme='off'

### PR DESCRIPTION
**Pull request by brian-eaton**
Issue #1108 - More robust logic in gw_drag.F90 when deep_scheme='off'
- Modify build-namelist to set use_gw_convect_dp=.false. when
  deep_scheme='off'.  In gw_drag::gw_tend check whether field TTEND_DP is
  in the pbuf.  If so then associate the ttend_dp pointer.  If not then
  allocate the ttend_dp pointer and set to zero.

PR #1101 - improved fix for rh write performance
- reorder output fields.  Merge Jim's PR into this one.

Issue #1106 - Report an error if a user uses --model_top with anything other than cam7
- configure reports this error, but the output in the log file gets
  obscured by a stack traceback issued from buildnml which is not useful.
  The fix implemented in buildnml replaces the "raise RuntimeError" call by
  a warning message in the log file.  If a subsequent check for CAM's
  config_cache.xml file fails, then execution is terminated by a call to
  the CIME.utils "expect()" routine.

Issue #1058 - Remove unused pbuf variable smaw
- Remove both smaw and turbtype from physics buffer.  Neither is used.
  Remove the calculation of smaw entirely.  Calculation of turbtype
  remains.  It is used locally, and may be written to history file
  (UW_turbtype).

Issue #1051 - Bad logic in SE dycore "interpolate_vector" subroutines
- These subroutines are not currently used by CAM as they are restricted to
  interpolating fields on the GLL grid.  Fix the conditional logic and
  update the endrun message.

Issue #1050 - Remove CAM3 as a compset or configure option

fix #1108 
resolve #1101 
fix #1106 
fix #1058 
fix #1051 
fix #1050

### CODE REVIEW

**Review by nusbaume — APPROVED** [review]

File: src/physics/cam/gw_drag.F90 [hunk 1]
```diff
@@ -1783,7 +1783,15 @@ subroutine gw_tend(state, pbuf, dt, ptend, cam_in, flx_heat)
      call outfld('UPWP_CLUBB_GW', upwp_clubb_gw, pcols, lchnk)
      call outfld('VPWP_CLUBB_GW', vpwp_clubb_gw, pcols, lchnk)
 
+     !Deallocae variables that are no longer used:
```
cacraigucar, line 1786: typo - needs a "t" in deallocate

nusbaume: Good catch, thanks @cacraigucar!

cacraigucar (approved):

PR #1137, opened 2024-08-28
cam6_4_026: Neglected to remove the 0.5*timestep call from zm_convr_run - done now

### LINKED ISSUE #1124: Remove half timestep from ZM code

The Zhang McFarlane code currently passes in .5*timestep.  Internally it then multiplies the values by 2 to get it back to a full timestep.  After discussion with @swrneale and @adamrher it was agreed that this should be removed

_No response_

I Don't Know

Yes

cacraigucar:
Actually removes the 0.5 from the call to zm_convr_run

closes #1124 with all the last change which should have been in there


nusbaume (approved):
Well it was a pretty brutal review, but I just made it through.  Thanks @cacraigucar!

PR #1160, opened 2024-10-02
cam6_4_040: Bring in re-organized atmospheric_physics repo

### LINKED ISSUE #1156: Modify CAM build system to handle re-organized atmospheric_physics repo
### Issue Type

Externals Update

### Issue Description

A new version of the atmospheric_physics repo (as described in issue ESCOMP/atmospheric_physics#124) will have a different directory structure then what exists currently, which means the CAM build system will need to be modified so that it still correctly builds and links everything that it needs to.

### Will this change answers?

No

Yes

nusbaume:
This PR brings in an updated atmospheric_physics external with a new, re-organized directory structure.  The associated atmospheric_physics PR can be found here:  ESCOMP/atmospheric_physics#126.  

These directory changes also required modifications to the `configure` script to  point to the new source code locations.

Fixes #1156


bld/configure
```diff
@@ -2360,7 +2360,7 @@ sub write_filepath
     print $fh "$camsrcdir/src/control\n";
     print $fh "$camsrcdir/src/utils\n";
     print $fh "$camsrcdir/src/utils/cam_ccpp\n";
-    print $fh "$camsrcdir/src/atmos_phys/utilities\n";
+    print $fh "$camsrcdir/src/atmos_phys/schemes/utilities\n";
```
jimmielin, line 2363: Can I suggest adding a comment to this section or to `utils/cam_ccpp` and `schemes/utilities`? It looks like `schemes/utilities` is only used for geopotential temp, I was wondering why this is a dozen lines away from the other CCPP-ized schemes.

.gitmodules
```diff
@@ -35,8 +35,8 @@
 
 [submodule "atmos_phys"]
 	path = src/atmos_phys
-	url = https://github.com/ESCOMP/atmospheric_physics
-        fxtag = atmos_phys0_05_000
+	url = https://github.com/nusbaume/atmospheric_physics
+        fxtag = fbef341
```
jimmielin, line 39: Everything looks good! Just a reminder to change this to the ESCOMP repo once the updated directory structure makes it to main.

jimmielin (approved):

bld/configure
nusbaume: I agree that this is confusing, but instead of adding a comment I went ahead and just moved the `utilities` path up to where the other atmospheric_physics directories are being added.  The reason being that while it is only bringing in geopotential now, in the future it could include various other physics utilities, and thus I was worried that a comment could quickly become out-of-date.

Of course if you think it would still benefit from having a comment, or if you would prefer the call in it's original spot, then just let me know.  Thanks!

.gitmodules
nusbaume: Thanks for the reminder!  A tag from the ESCOMP repo should be there now.

bld/configure
jimmielin: Thanks, I agree the comment would go outdated quickly, this looks good!

peverwhee (approved):

PR #1163, opened 2024-10-03
cam6_4_049: History bugfixes

### LINKED ISSUE #1149: cam history always chooses first averaging flag when using duplicate fields
### What happened?

If, in `user_nl_cam` you're trying to do something like:

```
fincl2='Q:I', 'Q:A'
```

in the hopes that you'll get an h1i and h1a file, both with 'Q', it won't work!

You'll get whichever flag you specify first. So, in this case, you'd just get an h1i with 'Q'

Run CAM with the `user_nl_cam` described above. Be sure your nhtfrq /= 1 for the test, since that overrides everything to "I" anyway.

cam6_4_032

CGD machine (e.g. izumi)

GNU

_No response_

Any CAM SE can do this

### Extra info

Not sure exactly, but this seems like the problem area (in fldlst)

```
      listentry => masterlinkedlist
      do while(associated(listentry))
        mastername = listentry%field%name

        call list_index (fwrtpr(1,t), mastername, ffld)
        if (ffld > 0) then
          prec_wrt = getflag(fwrtpr(ffld,t))
        else
          prec_wrt = ' '
        end if

        call list_index (fincl(1,t), mastername, ffld)

        if (ffld > 0) then
          avgflag = getflag (fincl(ffld,t))
          call inifld (t, listentry, avgflag,  prec_wrt)
        else if ((.not. empty_htapes) .or. (is_initfile(file_index=t))) then
          call list_index (fexcl(1,t), mastername, ffld)
          if (ffld == 0 .and. listentry%actflag(t)) then
            call inifld (t, listentry, ' ', prec_wrt)
          else
            listentry%actflag(t) = .false.
          end if
        else
          listentry%actflag(t) = .false.
        end if
        listentry=>listentry%next_entry

      end do
```

### LINKED ISSUE #1150: CAM history erroneously overrides averaging flags for nhtfrq=1
### What happened?

If nhtfrq=1, CAM history currently overrides (since cam6_3_140) the averaging flag to 'I'

```
    !
    ! Initialize history variables
    !
    do t=1,ptapes
      do fld=1,nflds(t)
        if (nhtfrq(t) == 1) then
           ! Override any non-I flags if nhtfrq equals 1
           tape(t)%hlist(fld)%avgflag = 'I'
        end if
        if (tape(t)%hlist(fld)%avgflag .ne. 'I') then
           hfile_accum(t) = .true.
        end if

```

BUT:
as @adamrher pointed out, this neglects subcycling averages.

Run CAM, outputting a variable that exists in a subcycle (like a CLUBB variable in the macmic loop) with the 'A' flag and observe that the results are the same as for the 'I' flag.

cam6_4_032

CGD machine (e.g. izumi)

GNU

_No response_

Any CAM SE can do this

### Extra info

_No response_

### LINKED ISSUE #1166: time_bounds array should not be present in history files containing instantaneous output
### What happened?

The time_bounds array is present in instantaneous history files as seen in output from the CESM3 BMT1850 104 run. Based on [previous discussions on this](https://github.com/ESCOMP/CESM/issues/194#issuecomment-1084734149), the time_bounds array should not be present in instantaneous history files.  

An example of time_bounds being present in a monthly instantaneous history file: /glade/derecho/scratch/hannay/archive/b.e30_beta02.BMT1850.ne30_t232.104/atm/hist/b.e30_beta02.BMT1850.ne30_t232.104.cam.h0i.0010-01.nc

and in a daily instantaneous history file:
/glade/derecho/scratch/hannay/archive/b.e30_beta02.BMT1850.ne30_t232.104/atm/hist/b.e30_beta02.BMT1850.ne30_t232.104.cam.h2i.0104-12-13-00000.nc

cam6_4_014

CISL machine (e.g. cheyenne)

Other (please specify below)

/glade/campaign/cesm/cesmdata/cseg/runs/cesm2_0/b.e30_beta02.BMT1850.ne30_t232.104

No

### Extra info

_No response_

### LINKED ISSUE #1167: Time shown in history file names is incorrect for instantaneous data
### What happened?

The translated time present in instantaneous history files does not match that shown in the actual file name. If the time variable translates to February 1st year 10 at 00Z, the history file name should end in "0010-02.nc". Right now, the file name ends in "0010-01.nc". 

Doing a ncdump on a file shows the issue:
ncdump -t -v time /glade/derecho/scratch/hannay/archive/b.e30_beta02.BMT1850.ne30_t232.104/atm/hist/b.e30_beta02.BMT1850.ne30_t232.104.cam.h0i.0010-01.nc
output: time = "0010-02-01"

cam6_4_014

CISL machine (e.g. cheyenne)

Other (please specify below)

/glade/campaign/cesm/cesmdata/cseg/runs/cesm2_0/b.e30_beta02.BMT1850.ne30_t232.104

No

### Extra info

_No response_

### LINKED ISSUE #1183: Suggested improvement for openfile

In [cam_pio_openfile ](https://github.com/ESCOMP/CAM/blob/cam_development/src/utils/cam_pio_utils.F90#L1141) the file is opened and then a message is written that it was successfully opened.  As a consequence if a file is not opened correctly the message is not printed.  This make it difficult to determine what file was causing the issue.   I would like to request that this message be moved to prior to opening the file:

 ```
if(pio_iotask_rank(pio_subsystem) == 0) then                                                                                                                
       write(iulog,*) 'Opening existing file ', trim(fname), file%fh                                                                                                  
 end if     
```
I think that this will be a great help in debugging issues with inputdata.

@brian-eaton 

No

Any CAM SE can do this

### PR DESCRIPTION
**Pull request by peverwhee**
## Summary
1. Add optional `sampled_on_subcycle` argument to `addfld`. If that's true for a field, we don't override the averaging flag to "I" if nhtfrq = 1 (every timestep). This reenables subcycling averages.
2. Check for incompatible duplicates
    - compatible duplicates (flags match) are ok - just take the first one (to keep existing CAM behavior)
    - incompatible duplicates (flags differ) - endrun
3. Remove time_bounds field from instantaneous history files
4. Update default filename for monthly instantaneous files to include full timestamp
5. Update git-fleximod to 0.9.3
6. Improve logging for cam_pio_openfile
7. Update cime to cime6.1.56

## Mods
*.gitmodules*
- update cime tag

*src/control/cam_history.F90*
- Remove nhtfrq=1 override
  - Add sampled_on_subcycle logic to keep override for subcycle accumulated fields
- modify `list_index` to optionally check for duplicates/compatibility
- endrun with message including all duplicates found (if necessary)
    - Sample error message: `ERROR:  FLDLST: Found duplicate field(s) with different averaging flags. Place in separate fincl lists: "Q:A", "Q:I" (fincl3). "T:A", "T:I" (fincl3). "U:A", "U:I" (fincl4).`
- Remove time_bounds from instantaneous history files
- Override instantaneous file name for monthly files to include full timestep

*src/control/cam_history_support.F90*
- Add `sampled_on_subcycle` to `field_info` object

*src/physics/**
- update subcycle addfld calls to include `sampled_on_subcycle=.true.`

*src/utils/cam_pio_utils.F90*
- Move filename logging to before pio call

*bld/namelist_files/use_cases/sd_waccm_sulfur.xml*
- remove bad XML comment line

*.lib/git-fleximod/**
- update to 0.9.3

closes #1149 
closes #1150 
closes #1166 
closes #1167 
closes #1183

### CODE REVIEW

**Comment by brian-eaton** [comment]
I'll just point out that the reason for setting the averaging flag to `I` when `nhtfrq=1` is to get the correct timestamps.  If the flag is `A` then the timestamps are set to midpoint values which is not correct for a single timestep.

**Comment by peverwhee** [comment]
@adamrher could you confirm that the addfld calls I've amended represent the full set of fields that are sampled on the subcycle?

File: src/physics/cam/check_energy.F90 [hunk 1]
```diff
@@ -201,12 +201,12 @@ subroutine check_energy_init()
...
-    call addfld('TEFIX',  horiz_only,  'A', 'J/m2', 'Total energy after fixer')
-    call addfld('EFIX',   horiz_only,  'A', 'W/m2', 'Effective sensible heat flux due to energy fixer')
-    call addfld('DTCORE', (/ 'lev' /), 'A', 'K/s' , 'T tendency due to dynamical core')
-    call addfld('DQCORE', (/ 'lev' /), 'A', 'kg/kg/s' , 'Water vapor tendency due to dynamical core')
+    call addfld('TEINP',  horiz_only,  'A', 'J/m2', 'Total energy of physics input', sampled_on_subcycle=.true.)
+    call addfld('TEOUT',  horiz_only,  'A', 'J/m2', 'Total energy of physics output', sampled_on_subcycle=.true.)
+    call addfld('TEFIX',  horiz_only,  'A', 'J/m2', 'Total energy after fixer', sampled_on_subcycle=.true.)
+    call addfld('EFIX',   horiz_only,  'A', 'W/m2', 'Effective sensible heat flux due to energy fixer', sampled_on_subcycle=.true.)
+    call addfld('DTCORE', (/ 'lev' /), 'A', 'K/s' , 'T tendency due to dynamical core', sampled_on_subcycle=.true.)
+    call addfld('DQCORE', (/ 'lev' /), 'A', 'kg/kg/s' , 'Water vapor tendency due to dynamical core', sampled_on_subcycle=.true.)
```
adamrher, line 209: I would nix the ```sampled_on_subcycle``` for all these addflds in check_energy.F90 -- their outfld calls are in phypkg.F90, and outside the subcycles.

adamrher (changes requested):
Hi @peverwhee, I found some missing addfld calls. 

micro_pumas_cam.F90, marcop_driver.F90, microp_aero.F90 all look good. There are a 5 addflds you missed, in stats_init_clubb in clubb_intr.F90. I couldn't figure out how to add a review comment that deep into clubb_intr, but they start at line 5116.

microp_aero.F90 has some dependencies that have addfld calls. These are: ndrop.F90, ndrop_bam.F90, nucleate_ice_cam.F90, hetfrz_classnuc_cam.F90. All addflds in those modules need the sample_on_subcycle flag.

Lastly, the drivers in subcol_SILHS.F90 are called inside the macmic loop if use_subcol_microp=.true. Support for subcolumns is a grey area but it shouldn't be too much work (?) to add sample_on_subcycle to all the addflds in subcol_SILHS.F90.

peverwhee:
@adamrher thanks for taking the time to check these! I've added/removed the `sampled_on_subcycle` arguments as you described.

- [x] stats_init_clubb in clubb_intr.F90 (5 fields)
- [x] ndrop.F90 (all addflds)
- [x] ndrop_bam.F90 (all addflds)
- [x] nucleate_ice_cam.F90 (all addflds)
- [x] hetfrz_classnuc_cam.F90 (all addflds)
- [x] subcol_SILHS.F90 (all addflds)
- [x] **REMOVE FROM** check_energy.F90 (all addflds)

src/control/cam_history.F90
```diff
@@ -5688,7 +5730,14 @@ subroutine wshist (rgnht_in)
...
-            fname_inst = interpret_filename_spec( hfilename_spec(t), number=(t-1), &
+            ! If this is a monthly instantaneous file, override to default timestamp
+            !  unless the user input a filename specifier
+            if (default_monthly_filename(t)) then
+               inst_filename_spec = '%c.cam' // trim(inst_suffix) // '.h%t%f.%y-%m-%d-%s.nc'
+            else
+               inst_filename_spec = hfilename_spec(t)
+            end if
+            fname_inst = interpret_filename_spec( inst_filename_spec, number=(t-1), &
                  prev=prev, flag_spec='i' )
```
brian-eaton, line 5741: A 1 month test reveals that the filename of an instantaneous monthly file still does not match the time data in the file.  The problem is that monthly output files have `prev=.true.`, so the timestamp in the filename is 1 timestep earlier than the time data in the file.  Note that the `prev` option was implemented as a hack to get the expected month into the filename for monthly output files.  The time at the end of a monthly period is at 0Z of day 1 of the following month.  By backing up a timestep you get the date to contain the month that corresponds to the time period.  This option should not be on for instantaneous data, but the logic that sets `prev` only takes into account the output frequency and not whether it's an instantaneous or accumulated file.  I think the best solution here is to simply remove the `prev=prev` optional argument from this call to `interpret_filename_spec` since you never want to use a previous timestep for instantaneous data.

peverwhee: good catch, @brian-eaton !

I've removed the `prev` argument from `interpret_filename_spec` for the instantaneous file

brian-eaton (approved):
Looks good.

.lib/git-fleximod/git_fleximod/cli.py
```diff
@@ -2,7 +2,7 @@
 import argparse
 from git_fleximod import utils
 
-__version__ = "0.8.4"
+__version__ = "0.9.2"
```
cacraigucar, line 5: I see that CESM is at 0.9.3 - should we be updating to that while we are at it?

test/system/archive_baseline.sh
```diff
@@ -68,6 +68,7 @@ case $hostname in
     test_file_list="tests_pretag_izumi_${CAM_FC,,}"
     cam_tag=$1_${CAM_FC,,}
     baselinedir="/fs/cgd/csm/models/atm/cam/pretag_bl/$cam_tag"
+    chmod_cmd="chmod -R a+r ${baselinedir}"
```
cacraigucar, line 71: Just curious what this change is doing and why it is needed

cacraigucar (commented):
Curiosity led me to looking into this PR.  This is not intended to be a full review, but rather documents a couple of things that I saw.

.lib/git-fleximod/git_fleximod/cli.py
peverwhee: yep! i will do this tomorrow!

test/system/archive_baseline.sh
peverwhee: some cime changes recently resulted in the baselines on izumi having the wrong permissions. this updates the baselines to be globally readable.

per jesse's suggestion, I will be updating this to also run on derecho since it can't hurt to be globally readable!

PR #1175, opened 2024-10-18
cam6_4_052: clubb_intr GPUization

huebleruwm:
This only modifies clubb_intr.F90 and doesn't require a new verseion of clubb. The purpose of this is the addition of `acc` directives, added in order to offload computations to GPUs. Besides the directives, this mainly consists of replacing vector notation with explicit loops, combining loops with the same bounds where possible, and moving non-gpuized function calls to outside of the GPU section. I also added some new notation for the number of vertical levels (`nzm_clubb` and `nzt_clubb`) that improves readability and will make it easier to merge in with future versions of clubb. I also included some timing statements, similar to the ones added in the [Earthworks ew-develop branch](https://github.com/EarthWorksOrg/CAM/tree/ew-develop), which this version of clubb_intr is also compatible with.

This is BFB on CPUs (tested with intel), runs with intel+debugging, and passes the ECT test when comparing CPU results to GPU results (using cam7). There's some options that I didn't GPUize or test (`do_clubb_mf`, `do_rainturb`, `do_cldcool`, `clubb_do_icesuper`, `single_column` ), so I left the code for them untouched and added some checks to stop the run if they're set when the code is compiled with OpenACC.

If there ends up being something wrong with these changes then [this version](94275055d5275b8ea2f1d6a8ea06242992dd3b66), which is an earlier commit that contains only a new OpenACC data statement and some timer additions, would be nice to get in at least.


nusbaume (approved):
Looks good to me!  I had some questions and change requests but none of them are required, and of course if you have any concerns with any of my requests then just let me know.  Thanks!

src/physics/cam/clubb_intr.F90
```diff
@@ -2754,80 +2739,58 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
-      call init_pdf_params_api( pverp+1-top_lev, ncol, pdf_params_zm_chnk(lchnk) )
+    if (clubb_do_icesuper) then
+      call pbuf_get_field(pbuf, naai_idx, naai)
     end if
 
-     if ( .not. allocated(pdf_implicit_coefs_terms_chnk(lchnk)%coef_wp4_implicit) ) then
-       call init_pdf_implicit_coefs_terms_api( pverp+1-top_lev, ncol, sclr_dim, &
-                                               pdf_implicit_coefs_terms_chnk(lchnk) )
-     end if
+    !  Initialize physics tendency arrays, copy the state to state1 array to use in this routine
```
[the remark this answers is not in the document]
huebleruwm: Added in cfd28249446085affc8674beccfe642a771559a7

src/physics/cam/clubb_intr.F90
```diff
@@ -2754,80 +2739,58 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
 
-      ! update local copy of state with the tendencies
-      ptend_loc%q(:ncol,top_lev:pver,1)=qvtend(:ncol,top_lev:pver)
-      ptend_loc%q(:ncol,top_lev:pver,ixcldice)=qitend(:ncol,top_lev:pver)
-      ptend_loc%q(:ncol,top_lev:pver,ixnumice)=initend(:ncol,top_lev:pver)
-      ptend_loc%s(:ncol,top_lev:pver)=stend(:ncol,top_lev:pver)
+    ! Determine number of vertical levels used in clubb, thermo variables are nzt_clubb
+    ! and momentum variables are nzm_clubb
+    nzt_clubb = pver + 1 - top_lev
+    nzm_clubb = pverp + 1 - top_lev
```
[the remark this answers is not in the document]
huebleruwm: Definitely a good idea. Added here 75f51d1331238b00228c2f0d5ea1086ddb2ebfa8.

src/physics/cam/clubb_intr.F90
```diff
@@ -2867,219 +2830,520 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
-      thlp2(1:ncol,pverp)   = thlp2(1:ncol,pver)
-      rtp2(1:ncol,pverp)    = rtp2(1:ncol,pver)
-      rtpthlp(1:ncol,pverp) = rtpthlp(1:ncol,pver)
-      wpthlp(1:ncol,pverp)  = wpthlp(1:ncol,pver)
-      wprtp(1:ncol,pverp)   = wprtp(1:ncol,pver)
-      wp2(1:ncol,pverp)     = wp2(1:ncol,pver)
-      wp3(1:ncol,pverp)     = wp3(1:ncol,pver)
-      up2(1:ncol,pverp)     = up2(1:ncol,pver)
-      vp2(1:ncol,pverp)     = vp2(1:ncol,pver)
+    ! Initialize the apply_const variable (note special logic is due to eularian backstepping)
```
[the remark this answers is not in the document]
huebleruwm: Fixed here cfd28249446085affc8674beccfe642a771559a7

src/physics/cam/clubb_intr.F90
```diff
@@ -2867,219 +2830,520 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
-
-    call t_startf("clubb_tend_cam_i_loop")
+    !--------------------- Initializations --------------------
 
-    ! Determine Coriolis force at given latitude. This is never used
-    ! when CLUBB is implemented in a host model, therefore just set
-    ! to zero.
-    fcor(:) = 0._r8
+    !  Set the ztodt timestep in pbuf for SILHS
+    ztodtptr(:) = 1.0_r8*hdtime
```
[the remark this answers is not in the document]
huebleruwm: I have no idea why this is, but I was wondering the same thing. The comment implies it's only for SILHS though, which apparently has a dubious future in cam, so I left it untouched.

src/physics/cam/clubb_intr.F90
```diff
@@ -2867,219 +2830,520 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
+    !$acc parallel loop gang vector default(present)
+    do i = 1, ncol
+      rtm(i,pverp)  = rtm(i,pver)
+      um(i,pverp)   = state1%u(i,pver)
+      vm(i,pverp)   = state1%v(i,pver)
+      thlm(i,pverp) = thlm(i,pver)
+
+      !  Compute exner at the surface for converting the sensible heat fluxes
+      !  to a flux of potential temperature for use as clubb's boundary conditions
+      inv_exner_clubb_surf(i) = 1._r8 / ( ( state1%pmid(i,pver) / p0_clubb )**( rairv(i,pver,lchnk) / cpairv(i,pver,lchnk) ) )
```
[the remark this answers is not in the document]
huebleruwm: I also thought this might be bit changing, so I never touched it. But turns out it is BFB, I added this change here 75f51d1331238b00228c2f0d5ea1086ddb2ebfa8.

src/physics/cam/clubb_intr.F90
```diff
@@ -3843,63 +4107,59 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
 
     ! Initialize clubbtop to top_lev, for finding the highlest level CLUBB is
     !  active for informing where to apply the energy fixer.
+    !$acc parallel loop gang vector default(present)
     do i=1, ncol
       clubbtop(i) = top_lev
       do while ((rtp2(i,clubbtop(i)) <= 1.e-15_r8 .and. rcm(i,clubbtop(i))  ==  0._r8) .and. clubbtop(i) <  pver)
-        clubbtop(i) = clubbtop(i) + 1
+       clubbtop(i) = clubbtop(i) + 1
```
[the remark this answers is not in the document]
huebleruwm: An accident, reverted here cfd28249446085affc8674beccfe642a771559a7.

src/physics/cam/clubb_intr.F90
```diff
@@ -2867,219 +2830,520 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
-    ! Define the CLUBB momentum grid (in height, units of m)
-    do k=1, nlev+1
-      do i=1, ncol
-        zi_g(i,k) = state1%zi(i,pverp-k+1)-state1%zi(i,pver+1)
+    call t_stopf('clubb_tend_cam:NAR')
+    call t_startf('clubb_tend_cam:acc_copyin')
+    !$acc data copyin( sclr_idx, clubb_params_single_col, grid_dx, grid_dy, rairv, cpairv, radf_clubb, qrl, &
+    !$acc              pdf_params_chnk(lchnk), pdf_params_zm_chnk(lchnk), &
+    !$acc              state1, state1%q, state1%u, state1%v, state1%t, state1%pmid, state1%s, state1%pint, &
+    !$acc              state1%zm, state1%zi, state1%pdeldry, state1%pdel, state1%omega, state1%phis, &
```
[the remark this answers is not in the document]
huebleruwm: Honestly I don't know either anymore. I remember at one point having trouble with this and finding out that you did have to allocate the derived type itself on the GPU (create/copyin), so I just keep doing it that way, but I discovered recently that it's completely unnecessary. However, I'm not sure if the lack of such need need is something nvidia has implemented, or if it's part of the API standard, or if the trouble I encountered forever ago was just a bug. Common advice from the internet is to copy the user-defined type in itself, but that might be for types that contain scalars or fixed sized arrays, since those values are contained in the memory storing the type variable.

So long story short - just in case we need it.

huebleruwm:
> Looks good to me! I had some questions and change requests but none of them are required, and of course if you have any concerns with any of my requests then just let me know. Thanks!

Good finds, I took you up on every suggestion.

huebleruwm:
I ran a PFS test to check how the performance changed, and initially found that clubb_tend_cam was ~8% slower with these changes (up to [cfd2824](https://github.com/ESCOMP/CAM/pull/1175/commits/cfd28249446085affc8674beccfe642a771559a7)). Which is unacceptable. I took a guess that the biggest performance killer was a massive loop I made to replace a large amount of vector notation, where the code just zeros out a bunch of arrays. That seemed to be the culprit, since afterwards the new code was slightly faster. 

Here's the timing output comparison now. I ran these a couple times and got roughly the same results.

From the cam_dev head I started with ([cam6_4_038](https://github.com/ESCOMP/CAM/releases/tag/cam6_4_038))
```
Region                                         PETs   PEs    Count    Mean (s)    Min (s)     Min PET Max (s)     Max PET
macrop_tend                                    512    512    MULTIPLE 135.3464    124.0819    366     146.0524    42     
  clubb_tend_cam                               512    512    MULTIPLE 132.0139    121.0235    366     142.8016    166    
    clubb_tend_cam_i_loop                      512    512    MULTIPLE 111.1615    102.4363    366     123.0577    166    

```

From the head of this branch ([75f51d1](https://github.com/ESCOMP/CAM/pull/1175/commits/75f51d1331238b00228c2f0d5ea1086ddb2ebfa8))
```
Region                                         PETs   PEs    Count    Mean (s)    Min (s)     Min PET Max (s)     Max PET
macrop_tend                                    512    512    MULTIPLE 132.7466    123.5249    366     144.4218    149    
  clubb_tend_cam                               512    512    MULTIPLE 129.4563    120.4303    366     140.8711    149    
    clubb_tend_cam:ACCR                        512    512    MULTIPLE 107.8117    100.6424    366     117.3765    149    
      clubb_tend_cam:advance_clubb_core_api    512    512    MULTIPLE 91.2902     84.0197     366     100.7022    149    
      clubb_tend_cam:flip-index                512    512    MULTIPLE 11.2684     9.9628      220     13.7559     455    
   clubb_tend_cam:NAR                          512    512    MULTIPLE 21.4333     19.0054     401     24.0401     76     
                      qneg3                    512    512    MULTIPLE 0.5783      0.5268      474     0.6311      66     
   clubb_tend_cam:acc_copyin                   512    512    MULTIPLE 0.0059      0.0045      358     0.0081      260    
   clubb_tend_cam:acc_copyout                  512    512    MULTIPLE 0.0038      0.0025      314     0.0061      149    
```

So it seems these changes made `clubb_tend_cam` faster by ~2%.

src/physics/cam/clubb_intr.F90
```diff
@@ -3213,31 +3482,44 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
-    vpwp_sfc_pert = 0.0_r8
+      !$acc parallel loop gang vector default(present)
+      do i=1,ncol
+        upwp_sfc(i)   = cam_in%wsx(i)/rho_ds_zm(i,1)               ! Surface meridional momentum flux
+        vpwp_sfc(i)   = cam_in%wsy(i)/rho_ds_zm(i,1)               ! Surface zonal momentum flux
+      end do
+
+    endif
+    
+    call t_startf('clubb_tend_cam:flip-index')
```
Katetc, line 3516: So, I might be confused, but it looks like this little else section above isn't included in a timer. Is that what we want? It's probably not a big deal, but I think timers should cover everything.

src/physics/cam/clubb_intr.F90
```diff
@@ -3574,22 +3816,24 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
 
-        do k=1,nlev+1
+        do k=1,nzm_clubb
           do i=1,ncol
             rvm_in(i,k) = rtm_in(i,k) - rcm_inout(i,k)
           end do
         end do
 
-        call update_xp2_mc_api( gr, nlev+1, ncol, dtime, cloud_frac_inout, &
+        call t_startf('clubb_tend_cam:update_xp2_mc_api')
```
Katetc, line 3827: I understand why this timer is just around the call to update_xp2_mc_api, but I kind of think it should include the double loop above that is called right before every time. Just to include all related calculations. Or add a whole "rain_turb" timer around this one.

src/physics/cam/clubb_intr.F90
```diff
@@ -3413,41 +3652,40 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
 
-    ! need to initialize macmic coupling to zero
-    if (macmic_it==1) ttend_clubb_mc(:ncol,:) = 0._r8
-    if (macmic_it==1) upwp_clubb_gw_mc(:ncol,:) = 0._r8
-    if (macmic_it==1) vpwp_clubb_gw_mc(:ncol,:) = 0._r8
-    if (macmic_it==1) thlp2_clubb_gw_mc(:ncol,:) = 0._r8
-    if (macmic_it==1) wpthlp_clubb_gw_mc(:ncol,:) = 0._r8
+    call t_stopf('clubb_tend_cam:flip-index')
 
     do t=1,nadv    ! do needed number of "sub" timesteps for each CAM step
```
Katetc, line 3690: This loop - Its kind of hard to tell but I think it's a larger loop for the cld_mac_mic run. Shouldn't this have an !acc tag?

Katetc (approved):
This is a really nice cleanup and the GPU code is not too obtrusive. I give it five stars!

My comments are mostly questions so I can understand what is going on as best as possible.

I am only about 96% confidant that the vertical dimension changes are correct through the whole file. I think the regression tests will be necessary to make sure we have that right in every case.

src/physics/cam/clubb_intr.F90
```diff
@@ -2339,25 +2346,23 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
+    real(r8), dimension(state%ncol,nzm_clubb,sclr_dim) :: &
       sclrm_forcing,  & ! Passive scalar forcing              [{units vary}/s]
       sclrm,          & ! Passive scalar mean (thermo. levels)          [units vary]
       sclrp2,         & ! sclr'^2 (momentum levels)                     [{units vary}^2]
       sclrp3,         & ! sclr'^3 (thermo. levels)                      [{units vary}^3]
       sclrprtp,       & ! sclr'rt' (momentum levels)          [{units vary} (kg/kg)]
       sclrpthlp,      & ! sclr'thlp' (momentum levels)        [{units vary} (K)]
-      wpsclrp           ! w'sclr' (momentum levels)                     [{units vary} m/s]
+      wpsclrp,        & ! w'sclr' (momentum levels)                     [{units vary} m/s]
+      sclrpthvp_inout   ! sclr'th_v' (momentum levels)                  [{units vary} (K)]
```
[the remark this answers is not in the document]
huebleruwm: You're right that it's dimensioned differently, but that was intentional. I noticed that we never used the whole array, only ever up to `pverp+1-toplev`, which is now `nzm_clubb`, so I set it to that instead.

src/physics/cam/clubb_intr.F90
```diff
@@ -2591,43 +2609,39 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
...
+#ifdef _OPENACC
+    ! These options have not been GPUized
+    if ( do_clubb_mf ) call endrun(subr//': do_clubb_mf=.true. not available when compiling with OpenACC')
+    if ( do_rainturb ) call endrun(subr//': do_rainturb=.true. not available when compiling with OpenACC')
+    if ( do_cldcool )  call endrun(subr//': do_cldcool=.true. not available when compiling with OpenACC')
+    if ( clubb_do_icesuper )  call endrun(subr//': clubb_do_icesuper=.true. not available when compiling with OpenACC')
+    if ( single_column .and. .not. scm_cambfb_mode )  then
+      call endrun(subr//': (single_column && !scm_cambfb_mode)=.true. not available when compiling with OpenACC')
+    end if
+#endif
```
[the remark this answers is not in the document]
huebleruwm: That was my first instinct too. It would take some of `acc update host` and `acc update device` directives to move the data we need back and forth inside of the if-statements, and there would be no issue with doing so. The reason I didn't do that is because I was testing the GPU results with a CPU ensemble baseline that I had run with those options turned off. So to confirm those changes I would've had to run another baseline with those options turned on, which would've taken at least another day, and I was trying to get this in quickly. Also I asked Vince about the options and he indicated they weren't run often or were more so test options that didn't turn out well. So I thought it was safest to leave the code in those sections untouched and just disable the options for GPU runs. I did do a little test to confirm the code ran to completion with them enabled though.

src/physics/cam/clubb_intr.F90
huebleruwm: The whole if statement is inside of a larger timing section labelled "ACCR" (for accelerated region). The `if` part isn't GPUized though, so it ends the "ACCR" timer (accelerated region) and starts the "NAR"  timer (non-accelerated region), then does the opposite at the end. This was needed because the `if` bit contains a call to `calc_ustar`, which I didn't want to GPUize (just so that all changes were contained in clubb_intr), but the `else` region is GPUized, so it's cost is included in the surrounding `ACCR` timer.

src/physics/cam/clubb_intr.F90
huebleruwm: This is the overall sub-stepping loop, so it has to be run sequentially. It also contains the call to `advance_clubb_core` and some other smaller routines, so it wouldn't be possibly to offload to the GPU.

src/physics/cam/clubb_intr.F90
huebleruwm: I think the "rain_turb" timer makes the most sense. I don't know why I didn't do that to begin with, I've changed the timers in this commit https://github.com/ESCOMP/CAM/pull/1175/commits/b1a243ce649a5061d6f7d3c8b1d54092fc02f924

src/physics/cam/clubb_intr.F90
```diff
@@ -58,6 +58,10 @@ module clubb_intr
 
   type (sclr_idx_type) :: &
     sclr_idx
+
+  integer :: &
+    nzm_clubb, &
+    nzt_clubb
```
cacraigucar, line 64: Could a comment be added to describe what each of these new variables are?  (I saw they were documented in the code, but it is somewhat common to document them when they are declared).

cacraigucar (commented):
Since I was curious, I took a cursory look at this PR.  I did find one minor item which I believe would be good to add (since it took me awhile to find the answer).  This is not intended to be a full review.

src/physics/cam/clubb_intr.F90
Katetc: Awesome 👍

nusbaume:
Hi @huebleruwm, apologies for the delay (we had some CESM-wide PRs that had to come in first), but this PR is finally close to the top of our queue and thus we are about to run our full regression testing suite.  Are there any remaining modifications you wanted to make before we test and merge, or is this PR ready to go on your end?

src/physics/cam/clubb_intr.F90
cacraigucar: @nusbaume - Could you please document these variables based on the comment which was further down in the code `! Determine number of vertical levels used in clubb, thermo variables are nzt_clubb
    ! and momentum variables are nzm_clubb`

cacraigucar (approved):

src/physics/cam/clubb_intr.F90
nusbaume: Hi @cacraigucar, I added some comments as requested.  Please let me know if they work for you.  Thanks!

cacraigucar: It looks great to me - thanks!

nusbaume:
@sjsprecious In case you didn't see this I just wanted to let you know that this PR has now been merged into `cam_development`.  It did change answers for the GPU regression test but I believe that was expected.   Anyways, I hope that helps!

sjsprecious:
Thanks @nusbaume. Since Gunther has verified the GPU results through ECT, everything looks good to me so far.

PR #1178, opened 2024-10-22
cam6_4_043: Make RRTMGP default radiation in CAM7

### LINKED ISSUE #1143: turning RRTMG-P on by default in CAM7   + some namelist defaults
### Issue Type

Other (please describe below)

### Issue Description

We would like to turn RRTMG-P on by default in CAM7.

We are also using some namelist changes:

We want to keep the correct radiative property for mam4_mode3 when running with RRTMGP. It was a bug and it was reported already. But if RRTMGP becomes the default, we want to make sure this is fixed:
```
M:mam4_mode3:/glade/campaign/cesm/cesmdata/inputdata/atm/cam/physprops/mam4_mode3_rrtmg_aeronetdust_c141106.nc
```

Other changes:
- We need to have a `ubc_file_path allowing` to full 20th century period.  I saw emails about this from Bill and Francis. 

```
ubc_file_path=  /glade/campaign/cesm/cesmdata/inputdata/atm/cam/chem/ubc/b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensAvg123.cam.h0zm.H2O.1849-2014_c240604.nc
```

- I think we want to change the  seasalt_emis_scale  to:    
```
 seasalt_emis_scale             = 0.75D0 
```

### Will this change answers?

Yes

No

---- 

@PeterHjortLauritzen, @cacraigucar, @tilmes

### LINKED ISSUE #1154: Create at least one CAM7 regression test on izumi
### Issue Type

Other (please describe below)

### Issue Description

Before the big push of regression tests are converted to CAM7 from CAM5, it would be good to get one or two representative SE dycore tests running on CAM7 immediately.

### Will this change answers?

No

No

brian-eaton:
- Resolve #1143 

- Remove some tests that added rrtmgp to the cam7 configuration.  Not needed since rrtmgp is now the default in cam7.

- Remove test of old cam7 development configuration (32 levels) which is no longer needed.

- Remove 13 month F2000climo test.  This was originally created to make sure we didn't make changes that hurt the performance of our production configuration for CMIP6 simulations.  This is no longer needed.

- Resolve #1154
  - add ERP_D_Ln9.ne3pg3_ne3pg3_mg37.FLTHIST.izumi_gnu.cam-outfrq9s


cacraigucar:
@brian-eaton I thought I remembered the 13 month test also being added to help trap a "around the year boundary" error that we had and didn't have a test for at the time.  I could be mistaken though

bld/namelist_files/use_cases/1850_cam_lt.xml
```diff
@@ -14,7 +14,7 @@
 
 <!-- Low top upper boundary conditions -->
 <ubc_specifier>'Q:H2O->UBC_FILE'</ubc_specifier>
-<ubc_file_path>atm/cam/chem/ubc/b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensAvg123.cam.h0zm.H2O.185001-201412_c230509cdf5.nc</ubc_file_path>
+<ubc_file_path>atm/cam/chem/ubc/b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensAvg123.cam.h0zm.H2O.1849-2014_c240604.nc</ubc_file_path>
```
cacraigucar, line 17: This file needs to be added to the svn inputdata repo.  Please confirm that it has the mandatory metadata as described at: https://www2.cesm.ucar.edu/working_groups/Atmosphere/amwg_datasets.html

cacraigucar (approved):

brian-eaton:
> @brian-eaton I thought I remembered the 13 month test also being added to help trap a "around the year boundary" error that we had and didn't have a test for at the time. I could be mistaken though

This is something we routinely checked in the old cam tests (pre-cime) by starting the run an hour before midnight on Dec 31.  Several of our current tests do have this feature.  I have created a set of test mods to do this with the FLTHIST test that I'm adding in this PR.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2548,7 +2548,7 @@
 <seasalt_emis_scale ver="mam7"                                           >1.62D0</seasalt_emis_scale>
 <seasalt_emis_scale ver="strat"                                          >0.90D0</seasalt_emis_scale>
 <seasalt_emis_scale ver="strat" clubb_sgs="1"                            >1.00D0</seasalt_emis_scale>
-<seasalt_emis_scale ver="strat" clubb_sgs="1" phys="cam7"                >1.5D0</seasalt_emis_scale>
+<seasalt_emis_scale ver="strat" clubb_sgs="1" phys="cam7"                >0.75D0</seasalt_emis_scale>
```
peverwhee, line 2551: is this change related to the switch from RRTMG -> RRTMGP? Either way, can you add a note in the changelog about the reason for this change?

peverwhee (approved):
one small request - thanks @brian-eaton !

cime_config/testdefs/testmods_dirs/cam/outfrq9s_mg3_default/shell_commands
```diff
@@ -3,6 +3,6 @@
 ./xmlchange ROOTPE='0'
 ./xmlchange ROF_NCPL=`./xmlquery --value ATM_NCPL`
 ./xmlchange GLC_NCPL=`./xmlquery --value ATM_NCPL`
-./xmlchange CAM_CONFIG_OPTS=' -microphys mg3' --append
+./xmlchange CAM_CONFIG_OPTS=' -microphys mg3 -rad rrtmg' --append
```
sjsprecious, line 6: @brian-eaton Is there a reason not to use `rrtmgp` here?

bld/namelist_files/namelist_defaults_cam.xml
brian-eaton: This is a cam7 tuning mod from issue #1143.  I've added a note to the ChangeLog.

cime_config/testdefs/testmods_dirs/cam/outfrq9s_mg3_default/shell_commands
sjsprecious: Thanks Brian for your update. I see where the problem is. This folder is used by the GPU regression test and the EarthWorks team has identified that there is a problem to run CAM with the default RRTMGP kernels on the GPU. In order to run it correctly, we either need to update the CAM->RRTMGP interface (like this https://github.com/EarthWorksOrg/CAM/pull/25) or update the RRTMGP kernel (like this https://github.com/EarthWorksOrg/rte-rrtmgp/commit/ac0f76e65e772903a541f6bad25750f6a28a9f92). Otherwise, the `rrtmgp_gpu` option is not expected to work properly here. I am surprised that the `rrtmgp` option is not working either as it should only turn on the RRTMGP CPU code. But the GPU tests may turn on some ACC directives accidentally. Do you have an error message for the `rrtmgp` option? The error message posted here seems coming from the `rrtmgp_gpu` option.

brian-eaton: Hi Jian.  I got a similar error message with the `rrtmgp` option (which in this PR is the default radiation for `cam7`).  The test that's failing, and a sample of the error output:
```
ERS_Ln9_G4-a100-openacc.ne30pg3_ne30pg3_mg17.F2000dev.derecho_nvhpc.cam-outfrq9s_mg3_default

deg0011.hsn.de.hpc.ucar.edu 61: Failing in Thread:1
deg0011.hsn.de.hpc.ucar.edu 61: Accelerator Fatal Error: call to cuStreamSynchronize returned error 700 (CUDA_ERROR_ILLEGAL_ADDRESS): Illegal address during kernel execution
deg0011.hsn.de.hpc.ucar.edu 61:  File: /glade/derecho/scratch/eaton/test-src/cam6_4_041_cam7/src/physics/rrtmgp/ext/rte-frontend/mo_rte_lw.F90
deg0011.hsn.de.hpc.ucar.edu 61:  Function: rte_lw:70
deg0011.hsn.de.hpc.ucar.edu 61:  Line: 312
```
I'm running regression tests now for this PR and plan to leave this test as it has been using `rrtmg`.  It would be good if you could open an issue if you want to get this test running with the `rrtmgp_gpu` option.

sjsprecious: Thanks Brian. The file you pointed to also has ACC directives and it means even using the `rrtmgp` option, some code will still be enabled on the GPU by the GPU regression test. I guess it won't work properly if only part of the RRTMGP GPU code is activated. I agree that we can just use `rrtmp` option here and address this problem in a separate issue. I just want to understand what is going on here. Thanks for your help and clarification.

sjsprecious (approved):

bld/namelist_files/use_cases/1850_cam_lt.xml
brian-eaton: Updated `ubc_file_path` file has the required metadata and has been added to the svn inputdata repo.

PR #1186, opened 2024-11-06
cam6_4_046: Update GPU tests with new XML options

### LINKED ISSUE #1165: Update GPU regression tests with new workflow

As discussed in https://github.com/ESMCI/cime/pull/4687, it is better remove the GPU options from the Python workflow in CIME and use XML files instead to configure a GPU test for CESM.

The following tags should be brought into CAM together to make the new GPU workflow function properly:
- cmeps1.0.22
- ccs_config_cesm1.0.8
- cime6.1.33

Once those new tags are merged in, the GPU test definition here (https://github.com/ESCOMP/CAM/blob/cam_development/cime_config/testdefs/testlist_cam.xml#L1493-L1510) needs to be updated accordingly.

_No response_

No

Yes

sjsprecious:
This PR updates the external tags for the new GPU XML options and the setups for the GPU regression test on Derecho.

I have run the `aux_cam` regression tests on Derecho's CPU. Comparing to the baseline `/glade/campaign/cesm/community/amwg/cam_baselines/cam6_4_044_intel`, I see `FIELDLIST field lists differ (otherwise bit-for-bit)`.

The GPU baseline is generated successfully with the new XML options and the results are BFB on Derecho's A100 GPU.

Closes #1165


peverwhee (approved):
thanks @sjsprecious !

PR #1207, opened 2024-12-11
add clim_modal_aero logic around aero prop call in cam7

### LINKED ISSUE #626: -chem none (prescribed bulk aero) broken on cam_dev
### What happened?

Running with `-phys cam_dev -chem none` re-breaks (https://github.com/ESCOMP/CAM/issues/261) prescribed bulk aerosols. The model is trying to use placeholder indices to access buffer vars that are not there...

```
261: ERROR: GET_PBUF1D_FIELD_BY_INDEX: index (-1) out of range
54: ERROR: GET_PBUF1D_FIELD_BY_INDEX: index (-1) out of range
59:Image              PC                Routine            Line        Source
59:cesm.exe           0000000002D86D16  Unknown               Unknown  Unknown
59:cesm.exe           00000000028F7C40  shr_abort_mod_mp_         114  shr_abort_mod.F90
59:cesm.exe           000000000073347E  physics_buffer_mp         848  physics_buffer.F90.in
59:cesm.exe           000000000153A470  modal_aero_calcsi        1475  modal_aero_calcsize.F90
59:cesm.exe           0000000001538717  modal_aero_calcsi        1397  modal_aero_calcsize.F90
59:cesm.exe           0000000000781A73  physpkg_mp_tphysb        2744  physpkg.F90
59:cesm.exe           000000000077C27A  physpkg_mp_phys_r        1108  physpkg.F90
59:cesm.exe           000000000057DBC6  cam_comp_mp_cam_r         258  cam_comp.F90
59:cesm.exe           0000000000567D54  atm_comp_nuopc_mp         877  atm_comp_nuopc.F90
```

While the tracer/chem code is not my "bag," this appears to be because the reg and init aren't called for modal_aero_calcsize when `chem -none` sets the modal flag in physpkg to false and then tphysbc tries to initialize aerosols by running the wet dep routines.

```
      !===================================================
      ! Run wet deposition routines to intialize aerosols
      !===================================================

      call modal_aero_calcsize_diag(state, pbuf)
      call modal_aero_wateruptake_dr(state, pbuf)

      !===================================================
```

So a potential fix might be to have an if statement that calls a different aero init routine within the above block. Just commenting those two lines out as a test resulted in a crash just downstream (NaNs in CLUBB), so something needs to be init, but hopefully someone else knows off the top of their head rather than my feeble brain trying to figure it out.

Build a vanilla F compset after adding `-phys cam_dev -chem none` to the build flags.

cam6_3_064

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

_No response_

adamrher:
Fixes #626

Shouldn't change answers, since there is no regression test for cam7 -chem none.

### DIFF
--- a/src/physics/cam7/physpkg.F90
+++ b/src/physics/cam7/physpkg.F90
@@ -2882,8 +2882,10 @@ contains
       ! Run wet deposition routines to intialize aerosols
       !===================================================
 
-      call modal_aero_calcsize_diag(state, pbuf)
-      call modal_aero_wateruptake_dr(state, pbuf)
+      if (clim_modal_aero) then
+        call modal_aero_calcsize_diag(state, pbuf)
+        call modal_aero_wateruptake_dr(state, pbuf)
+      end if
 
       !===================================================
       ! Radiation computations


fvitt (approved):
I think this makes sense.  Thanks for the fix.

fvitt:
@cacraigucar 
Should we consider adding a regression for this, such as:
```
  SMS_D_Ln9.ne3pg3_ne3pg3_mg37.FLTHIST.izumi_gnu.cam-outfrq9s_nochem
```
?

brian-eaton:
Note that my implementation of the cam7 aquaplanets (PR #1203) fixes this as they use `-chem none`.  They are close to being ready.

adamrher:
Thanks @brian-eaton

PR #1211, opened 2024-12-22
Revert t_sfc limiter in RRTMGP

### LINKED ISSUE #1185: RRTGMP complains that tsfc has values outside range
### What happened?

As reported by @cecilehannay and Jim Edwards, in coupled runs RRTMGP sometimes complains that:

```
radiation_tend: ERROR: kdist_lw%gas_optics: gas_optics(): array tsfc has values
outside range
```

A workaround was applied by @brian-eaton and is in cam6_4_044, but this just gets us through radiation, the surface temperature is still going to be outside the realistic range.

Then range that is checked by RRTMGP comes from its data files, and is 160 - 355K. 

As far as I know, run a coupled case with development version of CESM and RRTMGP (before the workaround).

before cam6_4_044

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

We should discuss this in at least the CAM development meeting to come up with a strategy for how to better diagnose this issue.

### LINKED ISSUE #1188: Revert t_sfc limiter
### Issue Type

Code Clean-up

### Issue Description

Pending a CICE tag that will (hopefully) fix the bad values being passed to RRTMGP, we need to remove the limiter added to `src/physics/rrtmgp/rrtmgp_inputs.F90` in https://github.com/ESCOMP/CAM/pull/1184

```
   t_sfc = merge(t_sfc, tref_min, t_sfc > tref_min)
   t_sfc = merge(t_sfc, tref_max, t_sfc < tref_max)
```

### Will this change answers?

I Don't Know

Any CAM SE can do this

nusbaume:
Reverts the surface temperature limiter used by RRTMGP in CAM, and instead updates the sea ice (CICE) tag to fix the actual cause of the bad surface temperature values in CESM.

Fixes #1185 
Fixes #1188

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -186,7 +186,7 @@ fxDONOTUSEurl = https://github.com/NCAR/ParallelIO
 [submodule "cice"]
 path = components/cice
 url = https://github.com/ESCOMP/CESM_CICE
-fxtag = cesm_cice6_5_0_12
+fxtag = cesm3_cice6_6_0_5
 fxrequired = ToplevelRequired
 fxDONOTUSEurl = https://github.com/ESCOMP/CESM_CICE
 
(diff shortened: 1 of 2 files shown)


brian-eaton (approved):

nusbaume:
Ran the fully-coupled test `SMS_Ld2.ne30pg3_t232.BLT1850.derecho_gnu.allactive-defaultio` with the surface temperature limiter removed and found that the model crashes as described in the relevant issue.  Then ran the same test with CICE tag `cesm3_cice6_6_0_6` and found that the test passed as expected.  Thus I believe these PR changes work as intended.

PR #1214, opened 2024-12-27
cam6_4_075: Clean up (all SE dycore) plus other misc

### LINKED ISSUE #819: dependance of some namelist default on the number of vertical levels (lev)?

This issue should probably remove all nlev references in places where "lt" and "mt" can be used instead of nlev.

In ``namelist_defaults_cam.xml``, there are parameters that are connected to the number of levels. While it makes sense for ``ncdata`` to be connected to nlev, it doesn't make sense for other parameters.
Instead these parameters should connected to the ``lt`` or ``mt`` configuration. This is to avoid getting the wrong parameters when changing nlev. 

For instance, we have  in ``namelist_defaults_cam.xml``
```
<gw_apply_tndmax                        >.true. </gw_apply_tndmax><gw_apply_tndmax chem="ghg_mam4" nlev="93">.false.</gw_apply_tndmax>
<gw_apply_tndmax chem="ghg_mam4" nlev="58">.false.</gw_apply_tndmax>
```

But if we are doing a LT run with a different number of levels than L58, we get ``gw_apply_tndmax=.true.``

I've just experienced this with a LT and L32. I changed it manually in ``user_nl_cam`` but it would be best if it would automatically come out the same for all ``LT`` runs.  

They are other parameters dependent of nlev=58 or 93

Yes

Any CAM SE can do this

### LINKED ISSUE #844: Crash when running cam_dev with SILHS

### LINKED ISSUE #1250: Change the  'parcel_hscale' parameter in the ZM convection to a namelist parameter

It's simple but I'd like to turn the hardwired parameter parcel_hscale Into a namelist parameter.
It's currently in the code here.

! Scaling of PBL height to give parcel mixing length for lparcel_pbl=True

   real(kind_phys), parameter :: parcel_hscale  = 0.5_kind_phys

The variable is in zm_convr.F90 and it should retain the 0.5 value as default to make it bfb in CAM7
It will make it easier for everyone to carry out sensitivity experiments and add visibility to an important parameter.

Nope. it's a pretty simple change.

No

Any CAM SE can do this

### LINKED ISSUE #1254: Incorrect freqsh definition in shallow convection diagnostics
### What happened?

The `freqsh` diagnostic in CAM4 shallow convection code is incorrectly defined.

```fortran
   ! --------------------------------------------------------!
   ! Calculate fractional occurance of shallow convection    !
   ! --------------------------------------------------------!

 ! Modification : I should check whether below computation of freqsh is correct.

   freqsh(:) = 0._r8
   do i = 1, ncol
      if( maxval(cmfmc2(i,:pver)) <= 0._r8 ) then
          freqsh(i) = 1._r8
      end if
   end do
```

From @adamrher:

> That is indeed strange -- a value of 1 should be assigned to the column when there is shallow convection. The comment above it "! Modification : I should check whether below computation of freqsh is correct." suggests we aren't the only one who thinks so (can you git blame this comment?).
> 
> I looked through hk_conv.F90 and verified that the array (cmfmc2, which I presume is cmfmc_sh in your code) cannot be negative. There are explicit limiters on the components that are summed and multiplied to give cmfmc2: beta and eta:
> https://github.com/ESCOMP/CAM/blob/6eb3046284ef57385ca7612165465668c97ec1f5/src/physics/cam/hk_conv.F90#L697.
> 
> And beta is limited using beta(i) = max(0.0_r8,beta(i)) statements anytime it's modified. So I believe this is a bug in this particular diagnostic, and that the logic should be:
> `if (maxval(cmfmc_sh(i,:)) > 0._kind_phys) then`

Observe definition of `freqsh` in a CAM4 model run

cam6_4_055

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

Found this during CCPPization work of Hack shallow convection.

Will likely address this in a separate PR after the CCPPization, in order to maintain bit-for-bit of the eventual CCPP PR.

PeterHjortLauritzen:
During the merge of the SE dycore in CAM-SIMA I discovered nudging variables being allocated and never used (many of these were 3D variables). These have been removed.

Some other minor clean-up as well.

This PR also includes the changes in PRs #1016, #1255, #1266, and #1269.

Resolves #844
Resolves #819  
Resolves #1250
Resolves #1254


cacraigucar:
@PeterHjortLauritzen - was this PR closed intentionally without being merged in?  If so, could you please give a comment on why you are doing so?

PeterHjortLauritzen:
> @PeterHjortLauritzen - was this PR closed intentionally without being merged in? If so, could you please give a comment on why you are doing so?

I renamed branch (apparently one can not do that with an open PR) ...

nusbaume (approved):

Issue #1196, opened 2024-11-25
Update how nitrogen deposition fluxes to surface models are handled


fvitt:
### Issue Type

Other (please describe below)

### Issue Description

When CTSM is updated to always receive nitrogen deposition fluxes from the coupler, as described in https://github.com/ESCOMP/CTSM/issues/1900, the ndep namelist settings in drv_flds_in should be eliminated.  

Another method of determining when chemistry can provide the nitrogen deposition fluxes, such a namelist parameter in atm_in or code internal to cam-chemistry.

We should also consider moving the CAM_STREAM_NDEP_DATA_FILENAME xml variable to a cam namelist parameter in atm_in.

### Will this change answers?

No

### Will you be implementing this yourself?

Any CAM SE can do this

fvitt:
Fixed in tag cam6_4_056

PR #1216, opened 2024-12-30
cam6_4_056: Nitrogen deposition fluxes

fvitt:
Addresses #1196


bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2060,6 +2060,13 @@
 <soil_erod_file phys="cam4"  >atm/cam/dst/dst_source2x2tuned-cam4-06132012.nc</soil_erod_file>
 <soil_erod_file phys="cam4" hgrid="0.9x1.25">atm/cam/dst/dst_source1x1tuned-cam4-06202012.nc</soil_erod_file>
 
+<!-- nitrogen deposition to surface models -->
+<stream_ndep_data_filename>lnd/clm2/ndepdata/fndep_clm_hist_b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensmean_1849-2015_monthly_0.9x1.25_c180926.nc</stream_ndep_data_filename>
+<stream_ndep_mesh_filename>share/meshes/fv0.9x1.25_141008_polemod_ESMFmesh.nc</stream_ndep_mesh_filename>
+<stream_ndep_year_first>2000</stream_ndep_year_first>
+<stream_ndep_year_last>2000</stream_ndep_year_last>
+<stream_ndep_year_align>1</stream_ndep_year_align>
+
```
brian-eaton, line 2069: By making use of the `sim_year` attribute we could define all the options for the `stream_ndep_*` variables right here rather than modifying a bunch of use case files for the non-`2000_` cases.

bld/namelist_files/use_cases/1850_cam_lt.xml
```diff
@@ -61,4 +61,11 @@
 <csw_time_type>CYCLICAL</csw_time_type>
 <csw_cycle_yr>1850 </csw_cycle_yr>
 
+<!-- prescribed nitrogen deposition fluxes -->
+<stream_ndep_data_filename>lnd/clm2/ndepdata/fndep_clm_hist_b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensmean_1849-2015_monthly_0.9x1.25_c180926.nc</stream_ndep_data_filename>
```
brian-eaton, line 65: This value of filename doesn't match what `config_component.xml` set for the `1850_` case.  Another good reason to set all these defaults in one place in `namelist_defaults_cam.xml`.

bld/namelist_files/use_cases/1850_cam_mt.xml
```diff
@@ -56,4 +56,11 @@
 <csw_time_type>CYCLICAL</csw_time_type>
 <csw_cycle_yr>1850 </csw_cycle_yr>
 
+<!-- prescribed nitrogen deposition fluxes -->
+<stream_ndep_data_filename>lnd/clm2/ndepdata/fndep_clm_hist_b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensmean_1849-2015_monthly_0.9x1.25_c180926.nc</stream_ndep_data_filename>
```
brian-eaton, line 60: Wrong filename.

brian-eaton (changes requested):

bld/namelist_files/namelist_defaults_cam.xml
fvitt: @brian-eaton 
I like your suggestion here. However this approach does not work for me.  See mods to my sandbox on derecho here:
```
 /glade/work/fvitt/camdev/nitro_dep_fluxes
```

It seems that the first entries in namelist_defaults_cam.xml are used, regardless of the `sim_year` setting...

brian-eaton: @fvitt, I have just looked at `build-namelist` in the  sandbox you pointed to.  It doesn't contain the mods needed to make this work.  Take a look at how it sets the default for `co2_readflux_fuel` for an example.  Also, since the `add_default` call needs to pass the `sim_year` attribute, you'll need to move the `add_default` calls for the `stream_ndep_*` variables to below where `$sim_year` is set (or move the setting of `$sim_year` up higher).

fvitt: @brian-eaton This has been updated us you suggested.  I think it is working correctly now.
Thank you for the tips.

brian-eaton (commented):
The definition of `sim_year` in `namelist_definition.xml` needs to have `2010` and `1850-2015` added to the list of valid values.

I wasn't sure whether the `sim_year` attribute needed to be added to any of the use case files that don't currently set it.  It depends on whether the use case is for a configuration that prognoses the ndep fluxes or not.  I don't know how to figure that out without running the model.

doc/ChangeLog
```diff
@@ -1,5 +1,162 @@
...
+One-line Summary:
+Github PR URL:
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+
+ Change logical determinations on how to set nitrogen deposition fluxes which are sent to
+ surface models through the NUOPC mediator.  This sets the nitrogen deposition fluxes to
+ prescribed CDEP input stream fluxes if corresponding namelist options are set.  Otherwise,
+ the nitrogen deposition fluxes set to chemistry computed fluxes if the chemistry is capable
+ of providing the fluxes.  Deprecated ndep_list option in drv_flds_in has beed removed.
```
cacraigucar, line 15: `beed` should be `been`

doc/ChangeLog
```diff
@@ -1,5 +1,162 @@
 ===============================================================
 
+Tag name: cam6_4_???
+Originator(s): fvitt
+Date:
+One-line Summary:
+Github PR URL:
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
```
cacraigucar, line 9: I looked for a mention of the removal of the SSP's in config_component and I did not see a mention about it in the ChangeLog.  Or are the SSP's still handled, but not in CAM code anymore?  Regardless, I believe a brief comment might be appropriate.

cacraigucar (changes requested):
Minor comment requests

doc/ChangeLog
fvitt: fixed

doc/ChangeLog
fvitt: I added a blurb about SSP compsets

bld/namelist_files/use_cases/2010_geoschem.xml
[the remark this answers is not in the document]
fvitt: Good catch.  I added sim_year to geoschem use case files.

brian-eaton (approved):

cacraigucar (approved):

Issue #856, opened 2023-07-17
HEMCO restart tests fail


fvitt:
### What happened?

When HEMCO emissions are used restarts are not bit-for-bit the same as continuous runs.
Regression testing give this failure:
    FAIL ERP_Ln9_Vnuopc.f09_f09_mg17.FCSD_HCO.cheyenne_intel.cam-outfrq9s COMPARE_base_rest

Until this is fixed, HEMCO configurations will not be supported.

### What are the steps to reproduce the bug?

ERP_Ln9_Vnuopc.f09_f09_mg17.FCSD_HCO.cheyenne_intel.cam-outfrq9s COMPARE_base_rest

### What CAM tag were you using?

cam6_3_118

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

No

### Extra info

@jimmielin will need to provide hemco external updates to address this reproducibility / restart issue.

lizziel:
Following up about the HEMCO restart issue, I figured out the problem is vertical regridding of 3D emissions using instantaneous surface pressure. The specific problem is that the minimum read frequency for HEMCO is 1 hour on the hour, i.e. 0:00, 1:00, etc., but restarted runs which use a sub-hour timestep, e.g. 30 min, begin off the hour, i.e. 0:00 first run (okay), but then 3:30, 6:30, etc. I am able to eliminate differences by changing the timestep to 1 hour. This results in vertical regridding using identical pressures across all runs since no regridding occurs off the hour in any of the runs.

Solutions within HEMCO could include storage of raw file data within HEMCO to be regridded each timestep, or sub-hour reads. Neither are great from a performance point-of-view. Another option, which is simpler but would impact you, is having restarted runs begin on the hour rather than the first timestep after the hour. This is what we do in GEOS-Chem and hence we have never had this problem before. Note that all of my testing using CAM-chem not GEOS-Chem. Is there a reason you start the subsequent runs at the timestep `after` the time corresponding to the restart file (`cam.r.*`)?

To be more clear about how runs are currently split up in time within CAM, here is a breakdown of my own 30 min timestep runs.

Compset: FCHIST_HCO
Config params: REST_N=3, STOP_N=3, STOP_OPTION=nhours

Run A – seven full timesteps (in GEOS-Chem this would only be six full timesteps)
nstep 0 - 00:00
nstep 1 – 00:30
nstep 2 – 01:00
[...]

PR #1221, opened 2025-01-09
cam6_4_082: Fix HEMCO ERS tests

lizziel:
**Description updated Feb 7, 2025:**
This PR updates submodule HEMCO_CESM to version 2.1.0 and updates the HEMCO configuration file used in CAM-chem to read 3D emissions every timestep. The HEMCO version in the model is unchanged.

The new version of HEMCO_CESM includes a fix when setting the HEMCO clock. Previously the HEMCO clock was incorrectly one timestep off from the model clock. This introduced differences when restarting the model. Together with the update to read 3D emissions every timestep, this PR will make the CAM-chem with HEMCO tests in the CAM test suite pass ERS tests.
Additional improvements in this update are: 
1. Reduced and corrected HEMCO prints. HEMCO prints still go to the CESM log but note that a future update coming soon will move that to the atm log, enabled by a new HEMCO version about to be released.
2. Fleximod capability in HEMCO_CESM

See issue discussion at https://github.com/ESCOMP/CAM/issues/856
See HEMCO_CESM updates related to the ERS test problem at https://github.com/ESCOMP/HEMCO_CESM/pull/41

Note that previously there was discussion (see below) of only being able to read the emissions with temporal frequency down to 1 hour. It turns out HEMCO read is possible via the configuration file. I have not quantified the slow-down of reading and regridding the 3D emissions every timestep but do not observe impact. There is the question of whether we want this to be the long-term handling of all 3D emissions through HEMCO or a different solution would be better. Read through the discussion below for more information on this.


lizziel:
This PR is based on cam6_4_038. Please let me know if you would like me to rebase on a new version. Once the version is settled on I will run the test suite again and create the changelog updates.

cacraigucar:
@lizziel - is the plan to continue working on getting the restarts to work no matter what timestep the model attempts to restart on?  In order to fully function in CESM, this requirement would need to be met.

lizziel:
@cacraigucar - Yes, I am working on a better fix than the one submitted, which is meant as a temporary solution to demonstrate we understand the issue. Since merge of this commit is a little ways off I may be able to come up with a better fix before then. This PR would still be needed since the HEMCO_CESM version update includes a necessary fix regardless, but if I come up with a more complete solution I could remove the new tests I added to the CAM test suite.

There are a few different ways to solve the ERS issue and the one I plan to pursue is expanding HEMCO read capability to allow configuration of reading/regridding an emission every time it is run. Right now you must hard-code frequency in the configuration file, and that hard-coded frequency has a minimum of 1hr. I am not going to expand the hard-coded minimum to number of minutes since then we would have the issue of having to update the config file based on timestep duration used.

The downside of this plan is that reading and regridding 3D emissions every timestep means more computation per timestep. However, other alternatives would take longer to implement and be more complex, such as storing raw file data between timesteps which HEMCO does not currently do. Another option is scaling emissions between timesteps for the vertical regrid rather than do a vertical interpolation from the raw data, but I am not familiar enough with the messy algorithm to know if that is even possible. If it is possible then we would need to save previous timestep pressures (if not already available) in the restart file.

The HEMCO update for this would go into a Z-version of the older HEMCO version 3.9 used by CESM so that we woud not require a GEOS-Chem version update at the same time. I am concurrently developing a branch using the latest GEOS-Chem (14.5) and HEMCO (3.10) but I don't think that needs to be done before the CESM3 freeze date.

Let me know your thoughts on my plan, including if you would like to go in a different direction for a fix.

cacraigucar:
@briandobbins - I think you might be the best person to weigh in on the pros/cons of the options @lizziel is contemplating.  Since HEMCO is slated to be the way emissions are brought into CAM (and this is called during the run phase to step the emissions along, so it is not just an init read), it is important to get it as efficient as possible.  Let us know your thoughts.

gold2718:
> allow configuration of reading/regridding an emission every time it is run

Could this be a variable that CAM could set through its namelist? That way, CAM could decide how often HEMCO needs to read and regrid, say every `N` timesteps. This would allow skipping reads on some timesteps if saving the regridding time is desired. CAM does that with other processes such as radiation and ali_arms.

briandobbins:
I'm woefully unfamiliar with HEMCO, I'm afraid, so let me ask some very naive questions first, and please correct anything I've gotten terribly wrong here.   At a basic level, this changes HEMCO reads from monthly to hourly, which is the highest frequency for HEMCO.  

1) For a run with, say, a 30-minute time-step, this changes file reads from once every ~1400 steps (monthly) to once every 2 steps (hourly), and this is the performance concern?  Do we open _new_ files each time, or read data from an already opened file?  And do we have timing stats on what this typically takes, both for the regridding and the read itself?

2) Is the need for the vertical regridding to pressure levels because of the on-the-hour restart, or general correctness?  

3) How much data do we typically read anyway?  Can we have a setting where we read N time-steps, so we do reads less frequently, albeit at the cost of memory (and, admittedly, development complexity)?

4) For my own curiosity, with runs with sub-hourly steps, do we do interpolation already?

Thanks - if nothing else, I'll learn a lot from this.

lizziel:
@briandobbins questions:

> For a run with, say, a 30-minute time-step, this changes file reads from once every ~1400 steps (monthly) to once every 2 steps (hourly), and this is the performance concern? Do we open new files each time, or read data from an already opened file? And do we have timing stats on what this typically takes, both for the regridding and the read itself?

Correct. HEMCO will read the same file every hour, or every timestep in my proposal, in order to do a vertical interpolation to the current pressure grid. I believe the file is kept open in the buffer but need to verify.

I have only done a timing test for offline GEOS-Chem and found a 6% increase in a 1-day run for a 4x5 run using 8 cores. Further testing is needed to determine scaling with cores and the test should be using CESM.

> Is the need for the vertical regridding to pressure levels because of the on-the-hour restart, or general correctness?

We do vertical regridding to pressure levels simply because that is what is done in the Modular Earth Submodel System (MESSy) algorithm for vertical regridding in HEMCO. CESM uses a vertical pressure grid that is dynamic and so we pass that dynamic pressure grid to HEMCO every timestep. I assume that decision was made for general correctness. We could instead used a fixed pressure grid for the interpolation but that is a judgment call from the CESM scientists.

I am not an expert on this algorithm or the alternatives that could be used, but I am sure that making the pressure ratio used in the algorithm constant rather than dynamic corrects the restart issue. We essentially have this ratio that is different every timestep since PEDGE is coming from CESM:

```
       DO l = 1, HcoState%NZ+1
          sigout(:,:,l) = HcoState%Grid%PEDGE%Val(:,:,l) &
                        / HcoState%Grid%PEDGE%Val(:,:,1)
       ENDDO
```

If we used an approximate grid that is constant then that would solve this issue. Or we might be able to implement a new interpolation from current emissions to new emissions based on change in pressure. That would require storing previous pressures and emissions in the restart file. But I am not sure if that interpolation would be possible without roundoff error or other errors.

> How much data do we typically read anyway? Can we have a setting where we read N time-steps, so we do reads less frequently, albeit at the cost of memory (and, admittedly, development complexity)?

This currently only pertains to 3D emissions which currently consists only of aviation emissions. Setting how frequently to read is easily done in the HEMCO configuration file. But my understanding is we need the tests to pass because the error introduced by not vertically interpolating every timestep is not acceptable for science runs.

> For my own curiosity, with runs with sub-hourly steps, do we do interpolation already?

In the version in CESM right now we only read and interpolate vertically at the start of the month. This PR changes that to hourly. For runs with sub-hourly timesteps there is no vertical interpolation between the hour. This is why the ERS tests which restart after 5 timesteps (2.5 hrs) give failure but my new tests which restart on the hour pass.

lizziel:
@gold2718,

> Could this be a variable that CAM could set through its namelist? That way, CAM could decide how often HEMCO needs to read and regrid, say every N timesteps. This would allow skipping reads on some timesteps if saving the regridding time is desired. CAM does that with other processes such as radiation and ali_arms.

We could set this through the namelist for a custom set of emissions. However, I think it is safer to use the HEMCO configuration file where read settings for all of the emissions are stored. Note that the read frequency is different for different emissions inventories and is not just a blanket value. Also, if users are allowed to change the read frequency to the same cadence as the files, in this case monthly, doesn't that defeat the purpose of the restart tests?

lizziel:
Another thought I have about the design for a fix is that it should be inventory-independent. Currently the problem is only one inventory of emissions simply because aviation emissions happens to be the only 3D emissions. Whatever fix is applied should automatically be applied to additional 3D inventories.

Also, I should clarify my timing results for monthly versus hourly read in GEOS-Chem. The 6% value is for HEMCO only. For a one day run the HEMCO time increased from 31s to 33s. This is negligible compared to total run time.

lizziel:
I figured out how to read emissions through HEMCO every timestep by a simple adjustment in the configuration file. The original ERS tests now pass with this update. I will amend this PR to remove the new 10hr tests since they are not necessary, and rebase on cam_development as well.

I am still investigating why ERP tests fail. It is isolated to CEDS ship emissions and aviation emissions.

lizziel:
Heads up that the ERP test fixes will go into a different PR since they will require HEMCO updates. Updating HEMCO should occur at the same time as updating GEOS-Chem, and that update is in draft PR https://github.com/ESCOMP/CAM/pull/1247.

This PR, to fix ERS tests, is ready to review and merge now. It should be done prior to https://github.com/ESCOMP/CAM/pull/1247.

doc/ChangeLog
```diff
@@ -1,5 +1,58 @@
+Tag name: TBD
+Originator(s): lizziel
+Date: 9 Jan 2025
+One-line Summary: Update to HEMCO_CESM 2.1.0 and introduce new FCSD_HCO tests that pass ERS
+Github PR URL: https://github.com/ESCOMP/CAM/pull/1221
+
+Purpose of changes (include the issue number and title text for each relevant GitHub issue):
+HEMCO_CESM includes fixes to allow ERS tests of CAM-chem with HEMCO to pass. This requires
+a clock time bug fix in HEMCO_CESM released in new version 2.1.0 as well as an update to
+the HEMCO configuration file to read 3D emissions every timestep.
```
fvitt, line 10: Please explicitly document that changing the PE-layout changes answers.  I have confirmed that this fixes reproducibility issues if the PE-layout does not change.  ERP and PEM tests failed in my testing of cam-chem with HEMCO while ERS and REP tests passed.

lizziel: Done

doc/ChangeLog
```diff
@@ -1,5 +1,58 @@
...
+
+Describe any changes made to build system:
```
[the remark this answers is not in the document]
lizziel: Done

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3692,9 +3692,9 @@
 
 <!-- HEMCO: Harmonized Emissions Component -->
 <use_hemco>.false.</use_hemco>
-<hemco_data_root>atm/cam/geoschem/emis/ExtData/HEMCO</hemco_data_root>
-<hemco_config_file>atm/cam/geoschem/emis/HEMCO_Config.CC.CEDS_AEIC19.NEx.c230615.rc</hemco_config_file>
-<hemco_diagn_file>atm/cam/geoschem/emis/HEMCO_Diagn.3_5_0.c230307.rc</hemco_diagn_file>
+<hemco_data_root>atm/cam/geoschem/ExtData/HEMCO</hemco_data_root>
+<hemco_config_file>atm/cam/hemco/HEMCO_Config.CC.CEDS_AEIC19.NEx.c241122.rc</hemco_config_file>
+<hemco_diagn_file>atm/cam/hemco/HEMCO_Diagn.3_5_0.c230307.rc</hemco_diagn_file>
```
cacraigucar, line 3697: These files need to be imported into the SVN inputdata repo.  Please insure they have the required metadata: https://www.cesm.ucar.edu/models/cam/metadata

cacraigucar (changes requested):

bld/namelist_files/namelist_defaults_cam.xml
lizziel: Do you mean the *.rc configuration files? Or all of the emissions data read into the model that are stored at hemco_data_root?

cacraigucar: All files that will be needed to run HEMCO, need to be stored in a place that a user on a different computer system has the data available.  In CESM, this happens automatically when data is stored in the svn inputdata repository.  If HEMCO has these files already available in the HEMCO library which is brought in via git-fleximod, then that is already happening.  I may have missed this nuance, but would like to confirm this.

fvitt (approved):

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: Yes, go ahead and add comment text at the top of the ASCII files to document the "required metadata" information.  As for putting the files into the svn repo, yes this is a task for @fvitt

lizziel: Okay, great. @fvitt, could you replace the files currently in /glade/campaign/cesm/cesmdata/cseg/inputdata/atm/cam/hemco/ with the files of the same name located at /glade/u/home/elundgren/config/add_metadata_to_HCO_files?

Let me know if you have any questions about other data that needs to be added to the SVN inputdata repo.

fvitt: Copied. Also imported to the SVN inputdata repo https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/atm/cam/hemco/

lizziel: Thanks! Could you also make sure the `inputdata/atm/cam/geoschem/ExtData` is up-to-date on the SVN inputdata repo? The link shows it is old, which just an older version of `ChemInputs` subdirectory and no subdirectory for HEMCO. The data is at `/glade/campaign/cesm/cesmdata/cseg/inputdata/atm/cam/geoschem/ExtData`, which is a symlink to my own data directory. If using a symlink is a problem we can try to find a different solution to give me permissions.

cacraigucar: @fvitt - Can you address Lizzie's last request here?

lizziel: @fvitt, let me know if you need me to do anything else for this PR.

fvitt: There is a least ~300GB of data under ExtData, some of which I do not have permissions to read. Do we really want to copy all of that to the SVN inputdata repo?

Could we give @lizziel permissions to do these copies?

cacraigucar: @lizziel - Have you copied over these files?

lizziel: @cacraigucar - I have not copied the files yet but can do this today. Where is the svn repository located on derecho?

lizziel: The data is now copied.

cacraigucar (approved):

Issue #1132, opened 2024-08-20
Extranious output to cesm.log 


jedwards4b:
### What happened?

Too much output of the form: 

```
run/cesm.log.5747942.desched1.240820-082508:dec0992.hsn.de.hpc.ucar.edu 1087:   wetdapa: negative value            4          39 -2.274025570316749E-034
run/cesm.log.5747942.desched1.240820-082508:dec0993.hsn.de.hpc.ucar.edu 1197:   wetdapa: negative value            9          47 -1.270211465045744E-035
run/cesm.log.5747942.desched1.240820-082508:dec0997.hsn.de.hpc.ucar.edu 1530:   wetdapa: negative value            9          42 -6.383863627659505E-036
run/cesm.log.5747942.desched1.240820-082508:dec0984.hsn.de.hpc.ucar.edu 282:   wetdapa: negative value            5          42 -9.754505026958672E-032
run/cesm.log.5747942.desched1.240820-082508:dec0987.hsn.de.hpc.ucar.edu 449:   wetdapa: negative value            7          40 -3.996442881802674E-037
run/cesm.log.5747942.desched1.240820-082508:dec0991.hsn.de.hpc.ucar.edu 966:   wetdapa: negative value           11          47 -3.273374581763710E-035
run/cesm.log.5747942.desched1.240820-082508:dec0998.hsn.de.hpc.ucar.edu 1564:   wetdapa: negative value            1          39 -6.222929777092708E-035
run/cesm.log.5747942.desched1.240820-082508:dec1001.hsn.de.hpc.ucar.edu 1738:   wetdapa: negative value           13          45 -3.237668182500167E-035
```

### What are the steps to reproduce the bug?

In compset BLT1850 ne30pg3_t232.   This output should be reduced to a minimum, only from task 0 or only when in DEBUG mode or both.   The only time this much output should be allowed is when immediatly followed by an abort.

### What CAM tag were you using?

cam6_4_016

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

No

cacraigucar:
Closed by #1228

PR #1228, opened 2025-01-15
cam6_4_066: Negative wet dep messages

fvitt:
This limits the negative aerosol wet deposition messages written to cesm.log 

Addresses #1132


src/chemistry/aerosol/wetdep.F90
```diff
@@ -645,22 +645,20 @@ subroutine wetdepa_v2(                                  &
 
       end do ! End of i = 1, ncol
 
+#ifdef DEBUG
       found = .false.
       do i = 1,ncol
-         if ( dblchek(i) < 0._r8 ) then
+         if (dblchek(i) < -1.e-10_r8) then
```
cacraigucar, line 651: Perhaps a comment on why `-1.3-10_r8` is being used instead of a normal check of `<0` would be appropriate

src/chemistry/aerosol/wetdep.F90
```diff
@@ -645,22 +645,20 @@ subroutine wetdepa_v2(                                  &
 
       end do ! End of i = 1, ncol
 
+#ifdef DEBUG
```
cacraigucar, line 648: Why is this checking only being done for #DEBUG?

cacraigucar (changes requested):

src/chemistry/aerosol/wetdep.F90
fvitt: The check are catching a lot of insignificant small negative values which have been to cesm log an mostly ignored.
With the addition of the `endrun` call, we are only doing the checks in debug mode.

cacraigucar: Perhaps a comment about this would be appropriate?

src/chemistry/aerosol/wetdep.F90
fvitt: added comments

src/chemistry/aerosol/wetdep.F90
fvitt: done

cacraigucar (approved):

PR #1235, opened 2025-01-24
cam6_4_071: PBL_utils atmospheric_physics integration

mwaxmonsky:
Integrated pbl_utils API moved into atmospheric_physics.  Removes current pbl_utils.F90 API except for `compute_radf(...)`.


.gitmodules
```diff
@@ -35,10 +35,10 @@
 
 [submodule "atmos_phys"]
 	path = src/atmos_phys
-	url = https://github.com/ESCOMP/atmospheric_physics
-        fxtag = atmos_phys0_07_001
+	url = https://github.com/mwaxmonsky/atmospheric_physics
+        fxtag = 426f759e1a49d1e0d10e7476e4bf69b02da812a6
```
nusbaume, line 39: Sadly CAM requires a "real" tag, which usually means merging development into main over in atmospheric_physics and then creating a tag of main.  Of course if you are not able to do that for any reason just let me know!

.gitmodules
```diff
@@ -151,7 +151,7 @@ fxDONOTUSEurl = https://github.com/ESMCI/ccs_config_cesm.git
 [submodule "cime"]
 path = cime
 url = https://github.com/ESMCI/cime
-fxtag = cime6.1.56
+fxtag = ef81824baa69947354f807e83924daf756e12937
```
nusbaume, line 154: I think we also need to use a tag here as well.  I believe all CIME PR commits are tagged, so you should be to find a tag that contains this particular hash commit (if we are not past it already).

src/physics/cam/clubb_intr.F90
```diff
@@ -3493,8 +3494,8 @@ subroutine clubb_tend_cam( state,   ptend_all,   pbuf,     hdtime, &
         ubar = sqrt(state1%u(i,pver)**2+state1%v(i,pver)**2)
         if (ubar <  0.25_r8) ubar = 0.25_r8
 
-        call calc_ustar( state1%t(i,pver), state1%pmid(i,pver), cam_in%wsx(i), cam_in%wsy(i), &
-              rrho(i), ustar )
+        rrho(i) = calc_rrho(rair, state1%t(i,pver), state1%pmid(i,pver))
```
nusbaume, line 3497: I think `calc_rrho` is now named `calc_ideal_gas_rrho`.

nusbaume (changes requested):
Thanks for bringing this back into CAM @mwaxmonsky!  It looks like there are some updates that need to happen  now that the atmospheric_physics PR has been merged, but otherwise everything else looks good.

.gitmodules
mwaxmonsky: Removed and using stock cam_dev version!

.gitmodules
mwaxmonsky: Fixed!

mwaxmonsky:
@nusbaume I just kicked off another round of regressions with a clean clone/checkout of the branch so hopefully nothing else breaks during the regressions but if it does, I'll update you.  Otherwise, this should be good to go now!

nusbaume (approved):
Everything looks good to me now.  Thanks again @mwaxmonsky!

PR #1246, opened 2025-02-04
cam6_4_076: Change cam7 testing to use MOM t232 mask.

### LINKED ISSUE #1236: Change our testing from using the POP land/ocean mask to the MOM t232 land/ocean mask

Since POP has been removed in cesm3 -- it's nonsensical that our testing is done with the POP land/ocean mask rather than the MOM t232 land/ocean mask. In. the past it was also seen as important to run our standalone simulations with the same land/ocean mask and or grid used in coupled simulations. I'm concerned that by running all of our testing and standalone simulations with the POP mask that we will miss important problems in the fully coupled simulations that now are ONLY using the MOM t232 land/ocean mask. Anything we can do to make our testing closer to the fully coupled simulations being done is a good idea in my mind.

@cacraigucar @nusbaume @peverwhee 

No

Any CAM SE can do this

brian-eaton:
Resolves #1236 - Change our testing from using the POP land/ocean mask to the MOM t232 land/ocean mask
* The cam7 tests using the ne30pg3_ne30pg3_mg17 grid have been updated to
  use ne30pg3_ne30pg3_mt232.
* One test used the f19_f19 grid which used the gx1v6 mask.  That test is
  updated to use f19_f19_mg17.


cime_config/testdefs/testlist_cam.xml
```diff
@@ -199,7 +199,7 @@
       <option name="comment" >CAM7-LT aquaplanet</option>
     </options>
   </test>
-  <test compset="QPLT" grid="ne30pg3_ne30pg3_mg17" name="ERP_D_Ln9" testmods="cam/outfrq3s_cosp">
+  <test compset="QPLT" grid="ne30pg3_ne30pg3_t232" name="ERP_D_Ln9" testmods="cam/outfrq3s_cosp">
```
adamrher, line 202: I think there should be a "m" in front of the t232 to indicate we are only using the ocean grid for its mask; the actual ocean grid should be ne30pg3.

brian-eaton: Thanks @adamrher.  I made this change in the `ccs_config_cesm` PR and in `testlist_cam.xml`.

adamrher (commented):
Can we (should we?) change all ```ne30pg3_ne30pg3_mg17``` tests to ```ne30pg3_ne30pg3_mt232```, and not just aux_cam / prealpha tests? I'm referring to these carma, aux_pumas, waccm, camchem, zmean_nudging categories in ```testlist_cam.xml```. I think it would make sense to convert these all at once.

adamrher (approved):

brian-eaton:
I'll open a new issue.

cacraigucar (approved):
Approving, but reminding that we need a new issue to replace the one which this PR is closing.

PR #1255, opened 2025-02-11
part of cam6_4_075: remove nlev dependence for some namelist defaults

### LINKED ISSUE #819: dependance of some namelist default on the number of vertical levels (lev)?

brian-eaton:
Resolve #819 - dependance of some namelist default on the number of vertical levels (lev)?
- There are just two variables other than ncdata that have an nlev attribute.  Those entries can be removed because there are no compsets that use them.


fvitt (approved):
Looks okay to me provided we are not breaking any aux_cam baseline tests.

brian-eaton:
Hi @fvitt.  I did run the regression tests with these changes and there were no baseline diffs.

PR #1257, opened 2025-02-14
Update README.md to point to Wiki pages

jtruesdal:
Most of this same information is on the Wiki so I pulled it from here and replaced it with a pointer to the Wiki.


jtruesdal:
Wrong repo.

Issue #997, opened 2024-03-18
RRTMGP not working with GPUs on derecho


brian-eaton:
### What happened?

@sjsprecious reported that CAM w/ RRTMGP is not running with GPUs on derecho.  He is also not able to run the standalone RRTMGP tests with tag v1.7 of the rte-rrtmgp external.  He has contacted Robert Pincus about this problem.

### What are the steps to reproduce the bug?

From Jian:

>  I used the nvhpc/23.7 compiler and cuda/12.2.1 version.
>  I downloaded the input data from git@github.com:earth-system-radiation/rrtmgp-data.git
>  I set "RTE_KERNELS=accel" to enable the GPU code.
>  My compiler flags are "-g -Minfo -Mchkptr -Mstandard -Kieee -Mchkstk
>  -Mallocatable=03 -Mpreprocess -acc -gpu=cc80,lineinfo -Minfo=accel".

### What CAM tag were you using?

cam6_3_148 and later

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

NVHPC

### Will you be addressing this bug yourself?

No

sjsprecious:
Thanks Brian for opening this issue. To provide additional information:

- The failure of running RRTMGP GPU code in CAM is caused by this change (https://github.com/ESCOMP/CAM/blob/cam6_3_148/src/physics/rrtmgp/radiation.F90#L964). In particular, the problem seems to come from changing the type of `fswc` from `ty_fluxes_byband` to `ty_fluxes_broadband`. Switching back to type `ty_fluxes_byband` for `fswc` works for the GPU code. However, based on the discussion with Brian, this does not make sense as type `ty_fluxes_byband` just allocates a few additional arrays than type `ty_fluxes_broadband`, and those arrays are not used by `fswc` calculation anyway.
- According to Robert, ICON model used `ty_fluxes_broadband` type for `fswc` and it worked fine (https://gitlab.dkrz.de/icon/icon-model/-/blob/release-2024.01-public/src/atm_phy_rte_rrtmgp/mo_rte_rrtmgp_interface.f90?ref_type=heads#L643).
[...]

PR #1260, opened 2025-02-14
cam6_4_072: Fix broken RRTMGP GPU tests

sjsprecious:
This PR fixes the broken RRTMGP GPU tests reported in #997 . The changes are borrowed from the EarthWorks code base (https://github.com/EarthWorksOrg/CAM/pull/25, thanks to John and Supreeth).

I used the `cam6_4_064` tag on Derecho and confirmed that running PUMAS+CLUBB+RRTMGP on the NVIDIA A100 GPU now passed the ensemble consistency test.

The CPU tests should be BFB. I updated the GPU tests to enable the RRTMGP GPU code so the results were different from the existing baseline as expected.

### DIFF
--- a/src/physics/rrtmgp/radiation.F90
+++ b/src/physics/rrtmgp/radiation.F90
@@ -1170,9 +1170,13 @@ subroutine radiation_tend( &
 
                   ! Compute the gas optics (stored in atm_optics_sw).
                   ! toa_flux is the reference solar source from RRTMGP data.
+                  !$acc data copyin(kdist_sw,pmid_day,pint_day,t_day,gas_concs_sw) &
+                  !$acc        copy(atm_optics_sw) &
+                  !$acc     copyout(toa_flux)
                   errmsg = kdist_sw%gas_optics( &
                      pmid_day, pint_day, t_day, gas_concs_sw, atm_optics_sw, &
                      toa_flux)
+                  !$acc end data
                   call stop_on_err(errmsg, sub, 'kdist_sw%gas_optics')
 
                   ! Scale the solar source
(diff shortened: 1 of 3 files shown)


sjsprecious:
fix #997

src/physics/rrtmgp/radiation.F90
```diff
@@ -1190,6 +1194,15 @@ subroutine radiation_tend( &
                if (nday > 0) then
 
                   ! Increment the gas optics (in atm_optics_sw) by the aerosol optics in aer_sw.
+                  !$acc data copyin(coszrs_day, toa_flux, alb_dir, alb_dif, &
+                  !$acc             atm_optics_sw, atm_optics_sw%tau, &
+                  !$acc             atm_optics_sw%ssa, atm_optics_sw%g, &
+                  !$acc             aer_sw, aer_sw%tau, &
+                  !$acc             aer_sw%ssa, aer_sw%g, &
```
nusbaume, line 1201: Just curious if it is generally required in OpenACC to pass in both the DDT (e.g. `aer_sw`) as well as the variables contained within the DDT (e.g. `aer_sw%tau`)?  I am mostly asking as it will help us decide when we should wrap up variables in DDTs or objects and when it would be more work then it is worth.  Thanks!

nusbaume (approved):
Looks good to me!  Just had one general question on the explicit passing of DDTs as well as the variables they contain.

src/physics/rrtmgp/radiation.F90
sjsprecious: Generally speaking, yes. This is something called [deep copy](https://www.nersc.gov/assets/Uploads/Talk-NERSC-ApplicationReadinessSeminar-2019-OpenACC.pdf) in OpenACC. Maybe in the future the OpenACC standard or a compiler flag will do it for us automatically. But for now, I would suggest always doing something like I am doing here.

Note that the `nvhpc` compiler does have a [-gpu=deepcopy](https://forums.developer.nvidia.com/t/openacc-derived-data-type-from-host-to-device-deep-copy/183299/2) option but it does not always work well for a complex model like CESM based on my experience.

PR #1266, opened 2025-03-01
part of cam6_4_075: Implement parcel_hscale as a namelist parameter in ZM deep convection scheme

swrneale:
Implement parcel_hscale parameter that is hard coded in the ZM scheme as the namelist parameter zmconv_parcel_hscale. Provide the default value same as that which is currently hard coded (=0.5) so that changes should be bfb.


bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2821,6 +2821,7 @@
 <zmconv_tau                                                           > 3600.0    </zmconv_tau>
 
 <zmconv_parcel_pbl                                                    > .false. </zmconv_parcel_pbl>
+<zmconv_parcel_hscale                                 phys="cam7"     >  0.5    </zmconv_parcel_hscale>
```
cacraigucar, line 2824: What happens if phys is not cam7?  The old logic had this being assigned for every run, so I would suggest removing the `phys="cam7"`

cacraigucar (changes requested):

bld/namelist_files/namelist_defaults_cam.xml
swrneale: If phys /= cam7 then  zmconv_parcel_pbl  = false and zmconv_parcel_hscale is never used. We could simply take out the phys="cam7" right?

cacraigucar: Yes, removing the `phys="cam7"` I believe is the proper fix.  (What happens we go to cam8, this variable would have no value?)

cacraigucar:
This feature has been fixed.  See: https://github.com/orgs/community/discussions/45240#discussioncomment-12380610

bld/namelist_files/namelist_defaults_cam.xml
nusbaume: @cacraigucar I just removed the `phys="cam7"` option.  Please let me know if you want anything else changed!

PR #1283, opened 2025-03-19
cam6_4_084: Nitrogen chemistry updates relevant to the thermosphere

### LINKED ISSUE #1155: WACCMX NO chemistry
### Issue Type

Other (please describe below)

### Issue Description

After the  JPL19 chemistry updates in tag cam6_3_095 it was noted that the thermosphere was much warmer.  The change in thermosphere temperature had been traced to the change in rate parameters for reaction
```
   N + O2  -> NO + O
```

From Hanli's notes:

NO chemistry: Had a discussion (Doug K., Dan, Nick, Hanli) on August 16 2024. I didn’t hear any objections to adopting Fernandez et al (1998) for `N(4S)+O2`, and Duff et al (2003) for `N(2D)+O2`. In addition, we have added the following to WACCM-X (Roble, 1995, included in TIEGCM): 
```
   N2D + NO -> N2  + O
   N2D  -> N4S+hv
   N2D + e  -> N4S + e + 2.38 eV
```
They slightly reduce NO, but also slightly cool the thermosphere (10 K in the upper thermosphere), probably due to decrease of N2D and thus exothermic heating.

### Will this change answers?

Yes

fvitt:
This fixes issues with the nitrogen chemistry which affected the thermosphere in WACCM and WACCMX

Closes #1155


src/chemistry/mozart/mo_usrrxt.F90
```diff
@@ -1092,6 +1098,28 @@ subroutine usrrxt( state, rxt, temp, tempi, tempe, invariants, h2ovmr,  &
           rxt(:,k,usr_CH3COCH3_OH_ndx) = 3.82e-11_r8 * exp_fac(:) + 1.33e-13_r8
        end if
 
+!-----------------------------------------------------------------
+!       ... N + O2 -> NO + O  Abel Fernandez, A. Goumri, and Arthur Fontijn; 1998
```
cacraigucar, line 1102: Are there DOI available for the three references listed in this new code?

cacraigucar (changes requested):

src/chemistry/mozart/mo_usrrxt.F90
fvitt: @hanli-liu, @npedatella
Do you have DOI references for these reaction rate calculations?

npedatella: The DOIs are:

doi:10.1021/jp972365k (Fernandez  et al. 1998)
doi:10.1029/2002GL016720 (Duff et al., 2003)
doi:10.1029/GM087p0001 (Roble 1995)

Also note that the Fernandez paper does not give the rate coefficients. These were calculated from the data in the paper by John Orlando, Doug Kinnison, and Jun Zhang. I am note sure if this should be mentioned or not.

cacraigucar (approved):
Thanks for the DOIs

Issue #1280, opened 2025-03-17
Convert all tests to the mt232 mask.


brian-eaton:
### What is the feature/what would you like to discuss?

The decision has been made to use the MOM t232 mask for all CAM testing. PR #1246 only adopted it for the ne30pg3 grids.  Extend this to all grids.

### Will this change (regression test) answers?

Yes

### Will you be implementing this enhancement yourself?

Yes

Issue #258, opened 2020-10-30
Most converted unit tests are CAM5 - may want to make some CAM6


cacraigucar:
Need to review unit tests and determine which should be using CAM6 instead of CAM5

peverwhee:
Would be helpful to also add a comment (of some standard format) for each test in testlist_cam.xml describing what all is tested. Can be updated when, for example, someone makes a change to an existing test to cover a specific feature. Would make keeping track of what is tested where easier.

gold2718:
I suggest using the `<option name="comment" ></option>` sections (and getting rid of the vestiges of the oooold test system (e.g., `151 tsm, ter).

nusbaume:
List of desired Atmosphere model compsets for CESM3 can be found here:

https://docs.google.com/spreadsheets/d/1h6D4ofy-tmtNDdVZNUC00S7HLgy0EGFPhFgbqMOpuS8/edit?gid=130129697#gid=130129697

cacraigucar:
add a test to use MPAS out-of-the-box with FKESSLER L58 1 degree (mpas120) - can replace a 32 level test if it exists

cacraigucar:
Make sure to keep a minimal set of CAM6 chemistry tests- talk to @fvitt about tests they need to have

cacraigucar:
Don't remove any FV3 tests, but don't add any new ones either

brian-eaton:
The current simple model tests are low res, i.e., ne3pg3 and mpasa480.  Is there a reason to add an mpas120 grid?

cacraigucar:
Testing overhaul is completed

PR #1285, opened 2025-03-21
cam6_4_087: Update CAM testing.

brian-eaton:
Making progress on (but not resolving):

  Issue #258 - Most converted CESM regression tests are CAM5 - may want to make most CAM7
  - simple model tests all use the 58 level LT vertical grid
  - refactor tests for FADIAB, FHS94, FKESSLER, FTJ16, FGRAYRAD

  Issue #1280 - Convert all tests to the mt232 mask.
  - All updated tests use grids with mt232 masks.


cime_config/testdefs/testlist_cam.xml
```diff
@@ -1012,76 +1002,7 @@
       <option name="wallclock">00:20:00</option>
     </options>
   </test>
-<!--              700                         -->
-  <test compset="FADIAB" grid="ne5_ne5_mg37" name="ERI_D_Ln18" testmods="cam/outfrq3s_bwic">
```
cacraigucar, line 1016: Was the test case outfrq3s_bwic testing a feature which should continue to be tested?

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1012,76 +1002,7 @@
...
-    <machines>
-      <machine name="izumi" compiler="nag" category="aux_cam"/>
-    </machines>
-    <options>
-      <option name="comment" >705 teq</option>
-      <option name="wallclock">00:30:00</option>
-    </options>
-  </test>
-<!--              706  tsm teq                      -->
-  <test compset="FADIAB" grid="ne5pg3_ne5pg3_mg37" name="PEM_D_Ln9" testmods="cam/outfrq3s">
```
cacraigucar, line 1066: Do we no longer need to run a PEM test (whatever that does)?

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1326,16 +1247,6 @@
       <option name="wallclock">01:00:00</option>
     </options>
   </test>
-<!--              730                        -->
-  <test compset="FADIAB" grid="ne16_ne16_mg17" name="ERC_D_Ln9" testmods="cam/terminator">
-    <machines>
```
cacraigucar, line 1331: Both terminator tests were eliminated.  Don't we still need to test terminator chemistry?

cime_config/testdefs/testlist_cam.xml
```diff
@@ -1394,40 +1305,62 @@
     </options>
   </test>
 
-  <test compset="FHS94" grid="mpasa480z32_mpasa480" name="ERC_D_Ln9" testmods="cam/outfrq3s_usecase">
+  <test compset="F2000dev" grid="ne30pg3_ne30pg3_mg17" name="MGP_D_Ln9" testmods="cam/outfrq3s">
     <machines>
-      <machine name="derecho" compiler="intel" category="prealpha"/>
```
cacraigucar, line 1399: You removed several prealpha simple model tests.  I always looked to them as being our first line of defense against externals breaking CAM jobs (CESM testing would uncover problems with externals before we bring those externals into CAM).  Did you feel like this functionality is captured with other tests?

cacraigucar (changes requested):

cime_config/testdefs/testlist_cam.xml
brian-eaton: The outfrq3s_bwic testmods set analytic initial conditions to baroclinic wave.  This is now the default for FADIAB and it's set in config_component.xml.

cime_config/testdefs/testlist_cam.xml
brian-eaton: The PEM test runs two smoke tests at different PE counts.  Since FADIAB doesn't run any physics this would only test that the dycore gets answers that are independent of task count.  This functionality is verified in many other tests.  The testing now includes just 1 test for FADIAB (per Peter and Jesse).  It is using test ERC_D_Ln9_P1x1 and is the only test that runs the model serially.

cime_config/testdefs/testlist_cam.xml
brian-eaton: FKESSLER is set up with terminator chemistry.  This is done in config_component.xml.

cime_config/testdefs/testlist_cam.xml
brian-eaton: Exactly.  Since the simple model tests use all stub components except for CAM, the testing only involves the CAM component and the infrastructure components.  The infrastructure components are tested in our other tests.  Simple models are basically CAM component only, so I think they should be tested in aux_cam.

cacraigucar (approved):
Thanks for the details.  I am purposfully leaving these conversations unresolved to facilitate reviewing these comments if needed in the future.

PR #1289, opened 2025-03-25
Fixes for failing restart tests

### LINKED ISSUE #1284: Two restart tests have answer changes due to moving mountain update in cam6_4_078
### What happened?

Restart tests for a couple of regression tests indicate answer changes between a full 9 time step run and a run that is broken in two pieces with a restart to finish the run.  The test failures were missed when making cam6_4_078.

Subsequent investigaion indicates the problem occurs when both the change to gv_convect.F90 and the new setting of effgw_beres_dp=0.15 are implemented together.  The line in gv_convect.F90 is:
`hdepth = max(1000._r8, hdepth*qbo_hdepth_scaling)`

Restarts are bit-for-bit when the line is reverted back to:
`hdepth = hdepth*qbo_hdepth_scaling`

It was noted in issue #1276 :` Change in gw_convect.f90: we introduced a check to make sure that the latent depths (variable called hdepth) do not exceed the range of latent heating depths covered by the lookup table.`

A fix needs to be implemented to get restart tests to be bit-for-bit

Using cam6_4_078:

Run: ERP_D_Ln9.ne30pg3_ne30pg3_mt232.QPC7.derecho_intel.cam-outfrq3s_cosp
This test was tested in the detail described above

Assume the following test which also fails restart is due to the same problem, but it has not been investigated:
ERS_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_WXma.derecho_intel.cam-outfrq9s

cam6_4_078

CISL machine (e.g. cheyenne)

Intel

/glade/derecho/scratch/cacraig/test_cac_intel_20250320150236

### Extra info

@JulioTBacmeister @PeterHjortLauritzen @mbramberger

peverwhee:
Two fixes:

1. As @brian-eaton noted, remove the task-dependent CS_RAINCERT variable from the failing cosp test
    - Fixes ERP_D_Ln9.ne30pg3_ne30pg3_mt232.QPC7.derecho_intel.cam-outfrq3s_cosp
3. Fix in RRTMGP to always set `cam_out%netsw` (previously was only set if shortwave radiation calculation was performed)
    - Fixes ERS_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_WXma.derecho_intel.cam-outfrq9s

closes #1284


PeterHjortLauritzen:
Thanks Courtney!

cacraigucar (approved):

PR #1321, opened 2025-06-04
Update externals to match cesm3_0_beta06/alpha06g

### LINKED ISSUE #1320: Update externals to match alpha06g

Since alpha06g brought in a CICE tag which eliminated excessive output from CICE, we will be making a quick tag to address this problem.

Check to see if any other externals need updating as well

_No response_

Yes

Yes

cacraigucar:
Closes #1320


nusbaume (approved):
Tough review, but I think I made it through!

PR #1338, opened 2025-06-28
cam6_4_106: Add "_mpas" suffix and use "ncol" dimension for MPAS-specific output variables

### LINKED ISSUE #1323: MPAS outputs w, theta, and rho
### Issue Type

Infrastructure Update

### Issue Description

MPAS has three unique output fields - w, theta, and rho. There are some oddities with these:
1. In the native .nc output, these are output with an output dimension nCells. All other outputs are with the output dimension nCols. When using ncremap, the variables that are dimension nCells are ignored and do not get remapped. 
2. All other CAM output variables are all caps, while w, theta, and rho are all lower case. 
3. All other output variables are output during the physics steps. However, w, theta, and rho are output during the dynamics steps. As a result, in principle, they are output at different times. In the .nc file, they are all the same time stamps. 

### Will this change answers?

I Don't Know

No

kuanchihwang:
This PR changes how MPAS-specific output variables are written to *history files*:

1. Add the "_mpas" suffix, making it clear that these variables are specific to MPAS and are different from the standard CAM ones.
2. Change the dimension name from "nCells" to "ncol", improving the compatibility with certain tools like `ncremap` from NetCDF Operator.

*Restart files* are not affected.

Closes #1323


PeterHjortLauritzen (approved):
Looks great ... thanks!

cacraigucar (approved):

Issue #1192, opened 2024-11-14
Conversion of RRTMGP to CCPP


peverwhee:
### Issue Type

Infrastructure Update

### Issue Description

Using the existing NOAA-CCPPized version of RRTMGP (https://github.com/NCAR/ccpp-physics/tree/main/physics/Radiation/RRTMGP), create interstitials in atmospheric_physics to connect CAM-SIMA to RRTMGP

### Will this change answers?

I Don't Know

### Will you be implementing this yourself?

Yes

PR #1339, opened 2025-06-30
cam6_4_108: Bring in file i/o object and update interfaces for CCPPized RRTMGP

peverwhee:
## Description
1. Brings in new atmospheric_physics tag - https://github.com/ESCOMP/atmospheric_physics/pull/258 which incorporates the file i/o object into RRTMGP
2. Updates calling lists to match new interfaces (in rrtmgp/radiation.F90 and rrtmg/radiation.F90)
3. Removes file reading from `src/physics/cam/cloud_rad_props.F90` and instead calls new CCPPized interface.

addresses #1192


.gitmodules
```diff
@@ -35,8 +35,8 @@
 
 [submodule "atmos_phys"]
 	path = src/atmos_phys
-	url = https://github.com/ESCOMP/atmospheric_physics
-        fxtag = atmos_phys0_14_001
+	url = https://github.com/peverwhee/atmospheric_physics
+        fxtag = 74c16d2905c6945ba210103494971fc48aca34f2
```
cacraigucar: A reminder to me to make sure this is updated prior to you making your tag

src/utils/cam_ccpp/pio_reader.F90
```diff
@@ -0,0 +1,2817 @@
+module pio_reader
```
cacraigucar: I am not reviewing this module since @peverwhee said it is a copy of what is in CAM-SIMA (and it was previously reviewed there).  

@nusbaume - this does bring up the unfortunate situation where we have identical copies of code which resides in two repos.  I'm not immediately seeing a way around that, but we do need to be extremely cognizant that we need to keep this version up-to-date with the CAM-SIMA version.

cacraigucar (changes requested):

cacraigucar (approved):

Issue #1195, opened 2024-11-22
Allow nonhydrostatic version of FV3


ta440:
### What is the feature/what would you like to discuss?

Allow FV3 to be run using its nonhydrostatic solver, instead of the default hydrostatic one. As the vertical velocity is now a prognostic variable, this requires the initialisation of a nonzero initial w field.

### Will this change (regression test) answers?

No

### Will you be implementing this enhancement yourself?

Yes

PR #1340, opened 2025-07-01
Allow nonhydrostatic FV3

ta440:
This PR addresses issue #1195 to allow fv3 to be run in nonhydrostatic mode.

Three additional variables of fv3_a_imp, fv3_p_fac, and fv3_use_logp are required for the nonhydrostatic solver.

Analytic ICs are now modified to have a W component, so that an initial vertical velocity can be prescribed for the nonhydrostatic solver. The existing analytical initial conditions in src/dynamics/tests/initial_conditions.F90 have been modified to work with W.

A few modifications are required in the CAM_FV3_interface repository, so this PR is coupled with a PR to that repository.


jtruesdal:
Testing another version of this in stormspeed.  Once all the bugs are ironed out we will create new PRs for cam_dev as well as the fv3 interface external.

PR #1347, opened 2025-07-21
cam6_4_105: update ccs_config_cesm submodule

brian-eaton:
. update the ccs_config_cesm submodule to include the gnu-12 environment on izumi.


cacraigucar (approved):
Nice that we are able to go away from the custom external!

PR #1348, opened 2025-07-21
Used for reviewing cam6_4_121: Fixes for atmospheric chemistry input interpolation, with debugging and tests

jrvb:
DO NOT SUBMIT.

This change is not for submittal, but is for reference to make it easier to compare the new code with the original and to highlight some of the errors in the original interpolation code.


cacraigucar:
Moved this PR to draft as it is only for reference

PR #1353, opened 2025-07-31
cam6_4_107: RRTMGP mods to support MT configurations

### LINKED ISSUE #1279: RRTMGP error with mpas and L93 mid-top configuration
### What happened?

Running the mpas dycore with FMTHIST causes an error in RRTMGP:
```
dec2193.hsn.de.hpc.ucar.edu 654:  radiation_tend: ERROR: kdist_sw%gas_optics: gas_optics(): array play has values
dec2193.hsn.de.hpc.ucar.edu 654:   outside range
```
This occurs with analytic_ic and real world inic, although at different time-steps, and sometimes in the ```kdist_lw%gas_topics``` call. The issue is related to there being more than 1 vertical level with pressures < 1 Pa in some grid columns of the mpas L93 configuration. The L93 RRTMGP fix we previously implemented (https://github.com/ESCOMP/CAM/issues/1063) only considers the case of 1 level < 1 Pa, and it will not work for 2 or more. I've confirmed that in a year long mpasa120 run, there are routinely 2 levels < 1 Pa in the run, and in a few instances there were 3 levels < 1 Pa. There may be even more in mpasa3.75 km runs (?)

In terms of a solution, should we generalize the 1 level fix https://github.com/ESCOMP/CAM/issues/1063 for some arbitrary number of levels < 1 Pa? I think a cleaner way to do this would be to just allow levels < 1 Pa into the RRTMGP pressure arrays in all (some?) instances, and then when doing the calculations in ```kdist_sw%gas_optics``` & ```kdist_lw%gas_optics```, force it to use the lowest pressure in the lookup table when < 1 Pa, and produce a warning instead of erroring out. Although we would probably want to error out if pressures were << 1 Pa.

A problem with the https://github.com/ESCOMP/CAM/issues/1063 approach, at least for MPAS, is it defines the size ```nlay``` of these pressure arrays at init time based on number of levels < 1 Pa in ```pref_edge``` in ref_pres.F90. That is, these are not the time varying pressures and so there could be times and grid columns where there is more than 1 level < 1 Pa and that would get snuck into the RRTMGP calculation and trigger the out of bounds error. That's what happened here -- the ```pref_edge``` for L93 mpas has 1 level < 1 Pa, and so when 2 or more occur during run time it errors out. So in addition to there being more than 1 level < 1 Pa in mpas runs, ```pref_edge``` is not a reliable surrogate for the number levels < 1 Pa at run time.

@briandobbins @brianpm @PeterHjortLauritzen @JulioTBacmeister @fvitt 

run a recent tag with FMTHIST and mpas

cam6_4_055

CISL machine (e.g. cheyenne)

Intel

_No response_

No

### Extra info

_No response_

cacraigucar:
More details to come...

Fixes #1279


src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,169 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
-   ncol = state%ncol
+!-----------------------------------------------------------------------
 
-   call physics_ptend_init(ptend,state%psetcols, 'radheat', ls=.true.)
+    ncol  = state%ncol
+    lchnk = state%lchnk
+    call physics_ptend_init(ptend, state%psetcols, 'radheat', ls=.true.)
+ 
+!++jtb
+! just use RRTMG's
```
cacraigucar: Modify this comment to make it a little more explicit?

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,169 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+    qrs_mrg(:,:) = qrs(:,:)
+
+!+++arh
```
cacraigucar: Remove (or modify) this comment

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,169 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+   icall = 0
+
+! can't find gas name 'O'
+   !call rad_cnst_get_gas(icall,'O    ', state, pbuf, gas_mmr)
+   !xommr(:pcols,:pver) = gas_mmr(:pcols,:pver)
+   !nullify(gas_mmr)
```
cacraigucar: Does this need to just be removed and replace with a comment?

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,169 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+
+   call rad_cnst_get_gas(icall,'O3   ', state, pbuf, gas_mmr)
+   xo3mmr(:pcols,:pver) = gas_mmr(:pcols,:pver)
+   nullify(gas_mmr)
+
+! can't find gas name 'N2'
+   !call rad_cnst_get_gas(icall,'N2   ', state, pbuf, gas_mmr)
+   !xn2mmr(:pcols,:pver) = gas_mmr(:pcols,:pver)
+   !nullify(gas_mmr)
+   !++ jtb set to 0.7547 ('standard' mass mixing ratio of nitrogen gas in atmos)
```
cacraigucar: Remove this and use a modification of the last comment line as the comment?

cacraigucar (commented):
@JulioTBacmeister and @adamrher - a few cleanup questions for you both

src/physics/cam/radheat.F90
peverwhee: I'll chime in to say that O is never in the radiation gas list (which is hard coded to `(/'H2O  ','O3   ', 'O2   ', 'CO2  ', 'N2O  ', 'CH4  ', 'CFC11', 'CFC12'/)`)

I forgot about this when I was originally going to modify rad_cnst_get_gas to not error if the gas isn't found. But because removing another line fixed my issues with CO2 and the gaslist is hardcoded, I opted not to modify this routine.

TLDR; this commented code can be removed in my opinion

src/physics/cam/radheat.F90
JulioTBacmeister: We should say
! Shortwave heating is RRTMG solar heating only

src/physics/cam/radheat.F90
JulioTBacmeister: That sounds good

JulioTBacmeister: I'd like to change this slightly to a more traceable calculation starting with US Standard atmosphere:
N2_VMR  = 0.78084
(from the U.S. Standard Atmosphere and NIST)

Molar mass of N2 =28.0134 g/mol
 
Molar mass of dry air =28.9647 g/mol

Gives 
N2_MMR = N2_VMR * ( N2_mass / Dry_air_mass) ~ 0.7553

JulioTBacmeister: P.S. This will not affect the LT results

JulioTBacmeister (commented):
I think I understand that this code is setting the upper limit for RRTMGP to 10 Pa for all cases except WACCM and WACCM-X (here the only mod is "_r8" to "D0").  If this is true I am in agreement with this.

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+   call rad_cnst_get_gas(icall,'O2   ', state, pbuf, gas_mmr)
+   xo2mmr(:pcols,:pver) = gas_mmr(:pcols,:pver)
+   nullify(gas_mmr)
 
-   call physics_ptend_init(ptend,state%psetcols, 'radheat', ls=.true.)
+   call rad_cnst_get_gas(icall,'O3   ', state, pbuf, gas_mmr)
+   xo3mmr(:pcols,:pver) = gas_mmr(:pcols,:pver)
+   nullify(gas_mmr)
+
+! Using standard US N2_VMR(0.78084) converted to N2_MMR=N2_VMR*(N2_mass/dry_air_mass)
```
peverwhee: would it make more sense to do this calculation in the code? By having parameters for both N2_mass and dry_air_mass? Just so we know for sure what numbers are being used for those?

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
      if (met_rlx(k) < 1._r8 .or. met_srf_feedback) then
-       ptend%s(:ncol,k) = (qrs(:ncol,k) + qrl(:ncol,k))
+       ptend%s(:ncol,k) = (qrs_mrg(:ncol,k) + qrl_mrg(:ncol,k))
      endif
    enddo
-   call calculate_net_heating_run(ncol, ptend%s(:ncol,:), qrl(:ncol,:), qrs(:ncol,:), fsns(:ncol), &
-                fsnt(:ncol), flns(:ncol), flnt(:ncol), .true., net_flx(:ncol), errmsg, errflg)
+   call calculate_net_heating_run(ncol, ptend%s(:ncol,:), qrl_mrg(:ncol,:), qrs_mrg(:ncol,:), .true., &
+           gravit, state%pdel(:ncol,:), fsns(:ncol), fsnt(:ncol), flns(:ncol), flnt(:ncol), net_flx(:ncol), &
+           errmsg, errflg)
```
peverwhee: as i mentioned in the atmospheric_physics review, "gravit" and "state%pdel" can be removed as arguments from these calls.

peverwhee: the "use" statement for "gravit" can also be removed above

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
-   integer :: ncol
-   character(len=512) :: errmsg
-   integer            :: errflg
+    integer  :: i, k , ico2
+    integer  :: ncol                                ! number of atmospheric columns
+    integer  :: lchnk                               ! chunk identifier
+    real(r8) :: qrl_mrg(pcols,pver)                 ! merged LW heating
+    real(r8) :: qrl_mlt(pcols,pver)                 ! M/LT longwave heating rates
+    real(r8) :: qrs_mrg(pcols,pver)                 ! merged SW heating
+    real(r8) :: qout(pcols,pver)                    ! temp for outfld call
```
peverwhee: Do we need this variable? It's set below but never used.

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+    real(r8) :: dcoef(6)                            ! for tidal component of heating
+    real(r8) :: tau_newt                            ! time scale for 'IR' relaxation
```
peverwhee: these variables can be removed

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+    real(r8),            intent(in)  :: flnt(pcols)       ! Net outgoing lw flux at model top
+    real(r8),            intent(in)  :: asdir(pcols)      ! shortwave, direct albedo
+    real(r8),            intent(out) :: net_flx(pcols)
 
 ! Local variables
-   integer :: i, k
-   integer :: ncol
-   character(len=512) :: errmsg
-   integer            :: errflg
+    integer  :: i, k , ico2
```
peverwhee: i believe "i" and "ico2" can be removed

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
+    real(r8) :: o3cool(pcols,pver) ! Fomichev cooling rate ! (K/s)
+    real(r8) :: co2cool(pcols,pver) ! Fomichev cooling rate ! (K/s)
+    real(r8) :: c2scool(pcols,pver) ! Fomichev cooling rate ! (K/s)
+    real(r8) :: xco2mmr(pcols,pver) ! CO2
+    real(r8) :: xo2mmr(pcols,pver)  ! O2
+    real(r8) :: xo3mmr(pcols,pver)  ! O3
+    real(r8) :: xommr(pcols,pver)   ! O
+    real(r8) :: xn2mmr(pcols,pver)  ! N2
+
+!+++arh
```
peverwhee: remove this Adam comment?

src/physics/cam/radheat.F90
```diff
@@ -52,19 +81,132 @@ subroutine radheat_register
...
 
 !================================================================================================
+!================================================================================================
+!================================================================================================
+
+  subroutine radheat_init(pref_mid)
+
+    
+    use nlte_fomichev,    only: nlte_fomichev_init
+    use cam_history,      only: add_default, addfld
```
peverwhee: can be removed

src/physics/cam/radheat.F90
```diff
@@ -52,19 +81,132 @@ subroutine radheat_register
...
+    use phys_control,     only: phys_getopts
+    use physics_buffer,   only : physics_buffer_desc
```
peverwhee: can be removed

src/physics/cam/radheat.F90
```diff
@@ -52,19 +81,132 @@ subroutine radheat_register
...
+    o2_mw  = 32._r8
+    o3_mw  = 48._r8
+    no_mw  = 30._r8
+    n2_mw  = 28._r8
+    nlte_limit_co2 = .true.
+! Initialize Fomichev parameterization
+    call nlte_fomichev_init (co2_mw, n2_mw, o1_mw, o2_mw, o3_mw, no_mw, nlte_limit_co2)
+    
+
+    ! determine upppermost level that is purely solar heating (no MLT chem heationg)
```
peverwhee: typo - 
```suggestion
    ! determine upppermost level that is purely solar heating (no MLT chem heating)
```

src/physics/cam/radheat.F90
```diff
@@ -9,32 +9,61 @@ module radheat
...
-public :: radheat_disable_waccm ! disable waccm heating in the upper atm
+  public :: radheat_disable_waccm ! disable waccm heating in the upper atm
+
+
+! Private variables for merging heating rates
+  real(r8):: qrs_wt(pver)             ! merge weight for cam solar heating
+  real(r8):: qrl_wt(pver)             ! merge weight for cam long wave heating
+
+  logical :: waccm_heating
+  logical :: waccm_heating_on = .true.
```
peverwhee: question for @JulioTBacmeister - do we need these variables at all in this module? Seems like these might be a relic of the old waccm code

src/physics/cam/radheat.F90
```diff
@@ -73,78 +215,165 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
-   if (errflg /= 0) then
-      call endrun('ERROR - failure during calculate_net_heating_run. Message: "'//errmsg//'"')
-   end if
-
-end subroutine radheat_tend
+  end subroutine radheat_tend
 
 !================================================================================================
   subroutine radheat_disable_waccm()
+    waccm_heating_on = .false.
```
peverwhee: if we don't need the waccm heating variables, this can also be removed

src/physics/cam/radheat.F90
```diff
@@ -52,19 +81,132 @@ subroutine radheat_register
...
+       if (qrs_wt(k)==1._r8) ntop_qrs_cam = k
+    enddo
+
+
+    ! WACCM heating if top-most layer is above merge region
+    waccm_heating = (pref_mid(1) .le. min_pressure_sw)
+
+    if (masterproc) then
+       write(iulog,*) 'WACCM Heating is computed (true/false): ',waccm_heating
+    end if
```
peverwhee: if we don't need the waccm heating variables, this chunk can be removed.

peverwhee (changes requested):
some cleanup

src/physics/cam/radheat.F90
peverwhee: the use statement has been removed but I believe (unless github is messing with me) that the interface arguments need to be removed still

src/physics/cam/radheat.F90
```diff
@@ -52,19 +76,122 @@ subroutine radheat_register
...
+
+    delta_merge_sw = max_pressure_sw - min_pressure_sw
+    delta_merge_lw = max_pressure_lw - min_pressure_lw
+
+    midpoint_sw = (max_pressure_sw + min_pressure_sw)/2._r8
+    midpoint_lw = (max_pressure_lw + min_pressure_lw)/2._r8
+
+    do k=1,pver
+
+       ! pressure scale heights for camrt merging (waccm4)
```
peverwhee: maybe change this?
```suggestion
       ! pressure scale heights for camrt merging (cam4)
```

src/physics/cam/radheat.F90
```diff
@@ -73,78 +200,154 @@ subroutine radheat_timestep_init (state, pbuf2d)
...
-   if (errflg /= 0) then
-      call endrun('ERROR - failure during calculate_net_heating_run. Message: "'//errmsg//'"')
-   end if
-
-end subroutine radheat_tend
+  end subroutine radheat_tend
 
 !================================================================================================
-  subroutine radheat_disable_waccm()
-  end subroutine radheat_disable_waccm
```
peverwhee: I think you may need to keep this hook here in case someone tries to use the radiation unit driver?

(I don't really know anything about that code, but it does sometimes call `radheat_disable_waccm`

peverwhee (changes requested):

.gitmodules
```diff
@@ -35,8 +35,8 @@
 
 [submodule "atmos_phys"]
 	path = src/atmos_phys
-	url = https://github.com/ESCOMP/atmospheric_physics
-        fxtag = atmos_phys0_14_001
+	url = https://github.com/cacraigucar/atmospheric_physics
+        fxtag = 205d248
```
peverwhee: comment reminder to add the tag when it's available

peverwhee (approved):

PR #1359, opened 2025-08-14
cam6_4_109: testing updates

brian-eaton:
. Add ERI tests (branch and hybrid runs) for cam5, cam6, and cam7
  F-compsets.  Previously only aquaplanet configurations tested. 

. Add testing of interpolated output for FHISTC_LTso.  Move COSP output to
  a separate history file that uses the native SE grid.  The COSP output of
  fields with more than 1 mdim to an interpolated lat/lon grid doesn't
  work.

. Add check to build-namelist to prevent specifiying COSP output on an
  interpolated history file.

. Updates to cam5 testing: combine several one-off tests into a single test
  and move most QPC5 tests to FHIST_C5.


cacraigucar (approved):
Nice cleanup!

Issue #1314, opened 2025-05-22
Convert diffusion_solver without molecular diffusion to CCPP and move to atmos_phys


jimmielin:
### Issue Type

Infrastructure Update

### Issue Description

This is a subtask of #1205 and a companion task of #1300 

A molecular-diffusion-less version of diffusion_solver.F90::compute_vdiff is being developed and will be CCPPized and moved to atmos_phys for the purpose of servicing CAM4 and CAM7 development in CAM-SIMA.

### Will this change answers?

No

### Will you be implementing this yourself?

Yes

jimmielin:
Completed in https://github.com/ESCOMP/CAM/pull/1361

Issue #1300, opened 2025-04-22
Convert HB (compute_hb_diff; pblintd) to CCPP and move to atmospheric_physics


jimmielin:
### Issue Type

Infrastructure Update

### Issue Description

Convert parts of vertical diffusion and `hb_diff.F90` to operate
* `compute_hb_diff`
* `pblintd` (pbl diagnostics for HB)
with the CCPP framework. Clean up and move the relevant code to atmospheric_physics.

Companion issue in ESCOMP/atmos_phys: https://github.com/ESCOMP/atmospheric_physics/issues/239

### Will this change answers?

No

### Will you be implementing this yourself?

Yes

jimmielin:
Completed in https://github.com/ESCOMP/CAM/pull/1361

Issue #1205, opened 2024-12-09
Convert vertical diffusion to CCPP and move to atmospheric_physics repo


adamrher:
### Issue Type

Other (please describe below)

### Issue Description

This vertical diffusion CCPP work will be broken into several parts due to all its dependencies, and this issue won't be resolved until all parts have been completed and all the relevant modules have been moved over to the atmospheric_physics repo.

The tentative plan is to modularize the various steps in vertical_diffusion.F90 into the following init, ts_init and run sequences:

<img width="897" alt="Screenshot 2024-12-09 at 2 05 11 PM" src="https://github.com/user-attachments/assets/9d393393-133b-449d-ad81-b37b4c891aa8">

Some of the more challenging parts are:

1. Adding radiative constituents flags to the aerosols in the constituents object (init). 
2. Remove host model dependencies from ```pbl_utils.F90``` and move into a share_util/ directory in atmospheric_physics repo. This requires modifying schemes that rely on this util in ESCOMP/CAM. At that stage we can strip out ```compute_hb_free_atm_diff``` from ```hb_diff.F90``` and place it in atmospheric_physics with dependencies on ```pbl_utils.F90```.
3. Moving ```tracer_data.F90``` (incl. pio and interpolation util dependencies) into the CAM-SIMA host model, in order to send data structures to ```upper_bc.F90``` routines, which will also need to be CCPP'ized and moved into the atmospheric_physics repo. 
4. vertical_diffusion_tend is envisioned to only call compute_vdiff, and to compute diagnostics. compute_vdiff will need to be stripped of host model dependencies, and since it is also called by the gravity wave scheme (@jtruesdal can you confirm?), it should be moved into the share_util/ directory, along with it's dependencies ```vdiff_lu_solver.F90``` and its other 3 dependencies currently residing in the to_be_ccppized/ directory (related to this issue: https://github.com/ESCOMP/CAM/issues/1173).

The other steps in this issue are hopefully more straightforward, but we'll see! This will be a collaboration among several SEs (@mwaxmonsky @jimmielin @mattldawson, and perhaps more...).

### Will this change answers?

Yes

### Will you be implementing this yourself?

Yes, but I will need some help

PR #1361, opened 2025-08-15
cam6_4_120: Complete CCPPization of HB PBL scheme and vertical diffusion solver (non-WACCM)

jimmielin:
Companion PR: https://github.com/ESCOMP/atmospheric_physics/pull/270

Purpose of changes (include the issue number and title text for each relevant GitHub issue):
- Completes #1300 by CCPPizing compute_hb_diff, compute_hb_free_atm_diff, pblintd, trbintd
- Cleanup, refactoring, and optimization of HB
- Partially completes #1205; and completes #1314 by CCPPizing partial interstitial logic of vertical_diffusion_tend
- Partially completes #1205; and completes #1314 by CCPPizing compute_vdiff in diffusion_solver.F90
- Correctly update 'AST' field in physics buffer in RK stratiform CAM interface for snapshot consistency (bit-for-bit).
- Adds snapshot points for orographic form drag stress (Beljaars in CAM6+ and TMS in CAM5) for future CCPPization work.

Describe any changes made to build system:
- add holtslag_boville CCPP scheme
- add vertical_diffusion CCPP scheme

Describe any changes made to the namelist: N/A

List any changes to the defaults for the boundary datasets: N/A

Describe any substantial timing or memory changes: N/A

Code reviewed by:

List all files eliminated:
```
D       src/physics/cam/hb_diff.F90
  - moved to atmos_phys as CCPPized holstlag_boville scheme
D       src/physics/cam/vdiff_lu_solver.F90
  - moved to atmos_phys into to_be_ccppized (unchanged)
```

List all files added and what they do: N/A

List all existing files that have been modified, and describe the changes:
```
R053    src/physics/cam/diffusion_solver.F90    src/physics/cam/diffusion_solver_cam.F90
  - this is the "remnant" diffusion solver supporting WACCM with molecular diffusion.
  - partial refactor to avoid excessive code divergence with CCPPized diffusion_solver.
  - renamed to avoid name conflict with CCPPized diffusion_solver.

M       bld/configure
  - add holtslag_boville and diffusion_solver folders in atmos_phys/schemes to build.

M       src/physics/cam/vertical_diffusion.F90
  - use CCPPized subroutines for HB and vertical diffusion solver.

M       src/physics/cam/eddy_diff_cam.F90
  - use CCPPized vertical diffusion solver.

M       src/physics/cam/molec_diff.F90
  - dechunkize and use assumed dimensions for vd_lu_qdecomp.

M       src/physics/cam/clubb_intr.F90
  - use CCPPized HB boundary layer computation.

M       src/physics/cam/ref_pres.F90
M       src/utils/air_composition.F90
  - comment updates for clarification of units/purpose.

M       src/physics/cam/rk_stratiform_cam.F90
  - Correctly update AST pbuf field in RK stratiform CAM interface for snapshot consistency (bit-for-bit).

M       src/physics/cam7/physpkg.F90
M       src/physics/cam/physpkg.F90
M       bld/build-namelist
M       bld/namelist_files/namelist_definition.xml
  - move Beljaars, TMS drag call routines to physpkg level for snapshotting.
  - add orographic_form_drag_stress snapshot point.

M       src/control/cam_snapshot_common.F90
  - expand size of cam_in snapshot output maximum because it needs to fit all constituents.
```

Verified bit-for-bit with Derecho Intel prior to submitting


src/physics/cam/vertical_diffusion.F90
```diff
@@ -490,24 +482,8 @@ subroutine vertical_diffusion_init(pbuf2d)
      call addfld( 'UFLX'        , (/ 'ilev' /)  , 'A', 'W/m2'   , 'Zonal momentum flux'                                )
      call addfld( 'VFLX'        , (/ 'ilev' /)  , 'A', 'W/m2'   , 'Meridional momentm flux'                            )
```
nusbaume: I think (?) the units for these fields should be `N/m2`.  Otherwise they don't match what is in the CCPP-ized SIMA diagnostics.

src/control/cam_snapshot_common.F90
```diff
@@ -84,7 +84,7 @@ module cam_snapshot_common
 type (snapshot_type)    ::  state_snapshot(30)
 type (snapshot_type)    ::  cnst_snapshot(pcnst)
 type (snapshot_type)    ::  tend_snapshot(6)
-type (snapshot_type)    ::  cam_in_snapshot(30)
+type (snapshot_type)    ::  cam_in_snapshot(200)   ! needs to be bigger than pcnst because cam_in is split by constituent.
```
nusbaume: Given that `pcnst` could theoretically change given the CAM compset being used, should we make this dimension size dependent on `pcnst` itself?  For example:

```suggestion
type (snapshot_type)    ::  cam_in_snapshot(pcnst+31)   ! needs to be bigger than pcnst because cam_in is split by constituent.
```

Note that I just based the `31` number off of the original `30` plus one extra for padding, but if a different number works better there then I am happy with that instead.

src/control/cam_snapshot_common.F90
```diff
@@ -1227,28 +1230,40 @@ subroutine cam_pbuf_snapshot_all_outfld(lchnk, file_num, pbuf)
...
-         ndims = count(pbuf_snapshot(i)%dim_name(:) /= '')
+         ! Retrieve the pbuf data. Special handling for certain
+         ! integer-type fields.
+         if( trim(pbuf_snapshot(i)%ddt_string) == 'clubbtop') then
+            call pbuf_get_field(pbuf, pbuf_idx, tmpptr2d_int)
+            ! copy into real
+            allocate(tmpptr2d(size(tmpptr2d_int, 1), size(tmpptr2d_int, 2)))
+            tmpptr2d = real(tmpptr2d_int, r8)
+            call outfld(pbuf_snapshot(i)%standard_name, tmpptr2d, pcols, lchnk)
+            deallocate(tmpptr2d)
```
nusbaume, line 1241: Do you need to allocate an entirely new variable here, or can you just do the following:

```suggestion
            ! convert into real
            call outfld(pbuf_snapshot(i)%standard_name, real(tmpptr2d_int, r8), pcols, lchnk)
```

src/physics/cam/diffusion_solver_cam.F90
```diff
@@ -153,99 +105,82 @@ subroutine compute_vdiff( lchnk           ,
...
+    ! Modification : Ideally, we should diffuse 'liquid-ice static energy' (sl), not the dry static energy.
+    !                Also, vertical diffusion of cloud droplet number concentration and aerosol number
+    !                concentration should be done very carefully in the future version.
+
+    ! Input Arguments
+    integer,  intent(in) :: ncol             ! Number of atmospheric columns
+    integer,  intent(in) :: pver
+    integer,  intent(in) :: pverp
+    integer,  intent(in) :: ncnst            ! # of constituents to diffuse. In eddy_diff, only wv. Others, pcnst.
+    real(r8), intent(in) :: ztodt            ! 2 delta-t [ s ]
```
nusbaume: I might remove the "2" here, as I believe that was only true for the Eulerian dycore (which is no longer in CAM):

```suggestion
    real(r8), intent(in) :: ztodt            ! delta-t [ s ]
```

src/physics/cam/diffusion_solver_cam.F90
```diff
@@ -153,99 +105,82 @@ subroutine compute_vdiff( lchnk           ,
...
-    real(r8), intent(out)   :: tautmsy(pcols)            ! Implicit meridional turbulent mountain surface stress
-                                                         ! [ N/m2 = kg m/s /s/m2 ]
-    real(r8), intent(out)   :: topflx(pcols)             ! Molecular heat flux at the top interface
-    character(128), intent(out) :: errstring             ! Output status
-
-    ! ------------------ !
-    ! Optional Arguments !
-    ! ------------------ !
+    real(r8), intent(in)    :: cpairv(:,:)      ! Specific heat at constant pressure
+    real(r8), intent(in)    :: mbarv(:,:)       ! Composition dependent atmosphere mean mass
```
nusbaume: Science nit-pick, but is `mbarv` the mean mass (kg) or the mean molar mass (kg mol-1)?  Might be good to specify that in the comment here (just in case).

src/physics/cam/eddy_diff_cam.F90
```diff
@@ -200,18 +197,10 @@ subroutine eddy_diff_init(pbuf2d, ntop_eddy_in, nbot_eddy_in)
...
-
-  call handle_errmsg(vdiff_select(fieldlist_wet,'s'), &
-       subname="vdiff_select")
-  call handle_errmsg(vdiff_select(fieldlist_wet,'q',1), &
-       subname="vdiff_select")
-  call handle_errmsg(vdiff_select(fieldlist_wet,'u'), &
-       subname="vdiff_select")
-  call handle_errmsg(vdiff_select(fieldlist_wet,'v'), &
-       subname="vdiff_select")
+  ! Only diffuse constituent 1 and disable molecular diffusion
```
nusbaume: I might call out "water vapor" here specifically:

 ```suggestion
  ! Only diffuse water vapor (constituent 1) and disable molecular diffusion
```

File: src/physics/cam/eddy_diff_cam.F90 [hunk 7]
```diff
@@ -464,7 +454,7 @@ subroutine compute_eddy_diff( pbuf, lchnk  ,
   real(r8), intent(in)    :: taux(pcols)               ! Zonal wind stress at surface [ N/m2 ]
   real(r8), intent(in)    :: tauy(pcols)               ! Meridional wind stress at surface [ N/m2 ]
   real(r8), intent(in)    :: shflx(pcols)              ! Sensible heat flux at surface [ unit ? ]
-  real(r8), intent(in)    :: qflx(pcols)               ! Water vapor flux at surface [ unit ? ]
+  real(r8), intent(in)    :: qflx(pcols,1)             ! Water vapor flux at surface [ unit ? ]
```
nusbaume: If this is coming from `cflx` then I believe the units are `kg/m2/s`:

```suggestion
  real(r8), intent(in)    :: qflx(pcols,1)             ! Water vapor flux at surface [ kg/m2/s]
```

src/physics/cam/molec_diff.F90
```diff
@@ -234,27 +234,26 @@ function vd_lu_qdecomp( &
...
     real(r8), intent(in)    :: mw                    ! Molecular weight for this constituent
-    real(r8), intent(in)    :: mw_facm(pcols,pver+1) ! composition dependent sqrt(1/M_q + 1/M_d) for this constituent
-    real(r8), intent(in)    :: dpidz_sq(ncol,pver+1) ! (g*rho)**2 (square of vertical derivative of pint)
+    real(r8), intent(in)    :: mw_facm(:,:) ! composition dependent sqrt(1/M_q + 1/M_d) for this constituent, interfaces
+    real(r8), intent(in)    :: dpidz_sq(:,:) ! (g*rho)**2 (square of vertical derivative of pint), interfaces
     type(Coords1D), intent(in) :: p                  ! Pressure coordinates
     type(BoundaryType), intent(in) :: interface_boundary ! Boundary on grid edge.
     type(BoundaryType), intent(in) :: molec_boundary ! Boundary at edge of molec_diff region.
-    real(r8), intent(in)    :: tint(pcols,pver+1)    ! Interface temperature [ K ]
+    real(r8), intent(in)    :: tint(:,:)    ! Interface temperature [ K ], interfaces
```
nusbaume: Probably don't need to say "interface" twice in the comments here:

```suggestion
    real(r8), intent(in)    :: tint(:,:)    ! Air temperature [ K ], interfaces
```

src/physics/cam/molec_diff.F90
```diff
@@ -234,27 +234,26 @@ function vd_lu_qdecomp( &
...
     real(r8), intent(in)    :: ztodt                 ! 2 delta-t [ s ]
 
-    integer,  intent(in)    :: lchnk		    ! Chunk number
-    real(r8), intent(in)    :: t(pcols,pver)	    ! temperature
+    real(r8), intent(in)    :: t(:,:)	    ! temperature
     integer,  intent(in)    :: m 		    ! cnst index
+    real(r8), intent(in)    :: mbarv(:,:)  ! composition dependent atmosphere mean mass
```
nusbaume: I might update these comments slightly:

```suggestion
    real(r8), intent(in)    :: t(:,:)	    ! Air temperature [ K ]
    integer,  intent(in)    :: m 		    ! cnst index
    real(r8), intent(in)    :: mbarv(:,:)  ! composition dependent molar mass of dry air [ kg kmol-1]
```

src/physics/cam/vertical_diffusion.F90
```diff
@@ -94,22 +90,32 @@ module vertical_diffusion
...
 ! ( when 'diag_TKE' scheme is selected )
 logical              :: do_pseudocon_diff = .false.  ! If .true., do pseudo-conservative variables diffusion
 
 character(len=16)    :: shallow_scheme               ! Shallow convection scheme
 
-type(vdiff_selector) :: fieldlist_wet                ! Logical switches for moist mixing ratio diffusion
-type(vdiff_selector) :: fieldlist_dry                ! Logical switches for dry mixing ratio diffusion
-type(vdiff_selector) :: fieldlist_molec              ! Logical switches for molecular diffusion
+logical              :: do_diffusion_const_dry(pcnst)! Do vertical diffusion (as dry) for this constituent?
+logical              :: do_diffusion_const_wet(pcnst)! Do vertical diffusion (as wet) for this constituent?
```
nusbaume: I might specify what "dry" and "wet" specifically mean in this context:

```suggestion
logical              :: do_diffusion_const_dry(pcnst)! Do vertical diffusion (as a dry mixing ratio) for this constituent?
logical              :: do_diffusion_const_wet(pcnst)! Do vertical diffusion (as a wet mixing ratio) for this constituent?
```

src/physics/cam/vertical_diffusion.F90
```diff
@@ -870,105 +906,180 @@ subroutine vertical_diffusion_tend( &
...
-
   call pbuf_get_field(pbuf, ksrftms_idx, ksrftms)
   call pbuf_get_field(pbuf, tautmsx_idx, tautmsx)
   call pbuf_get_field(pbuf, tautmsy_idx, tautmsy)
 
-  tautotx(:ncol) = cam_in%wsx(:ncol) + tautmsx(:ncol)
-  tautoty(:ncol) = cam_in%wsy(:ncol) + tautmsy(:ncol)
+  ! Add turbulent mountain stress to total surface stress for PBL scheme.
+  ! tautotx(:ncol) = cam_in%wsx(:ncol) + tautmsx(:ncol)
+  ! tautoty(:ncol) = cam_in%wsy(:ncol) + tautmsy(:ncol)
```
nusbaume: Remove commented-out code?

src/physics/cam/vertical_diffusion.F90
```diff
@@ -870,105 +906,180 @@ subroutine vertical_diffusion_tend( &
...
+
+  ! Use CCPPized interstitial schemes for setting
+  ! the necessary inputs (taux,tauy,shflux,cflux) for vertical_diffusion_compute_run from the coupler
+  ! as well as Coords1D (p) pressure coordinates used by the solver.
+
+  !REMOVECAM: no longer needed when pcols no longer exists
+  q_wv_cflx(:) = 0._r8
+  !END REMOVECAM
+
+  if(eddy_scheme .eq. 'CLUBB_SGS') then
```
nusbaume: Nit-pick, but I might change the syntax here to better match the CAM standards:

```suggestion
  if(eddy_scheme == 'CLUBB_SGS') then
```

src/physics/cam/vertical_diffusion.F90
```diff
@@ -1122,164 +1363,433 @@ subroutine vertical_diffusion_tend( &
...
+             gravit = gravit, &
+             cpairv = cpairv(:ncol,:,state%lchnk), &
+             zi = state%zi(:ncol,:), &
+             tint = tint(:ncol,:), &
+             dse_top = dse_top(:ncol), &
+             errmsg = errmsg, &
+             errflg = errflg)
      else
-       cflux(:,2:) = cam_in%cflx(:,2:)
+        dse_top(:ncol) = 0._r8
```
nusbaume, line 1599: I might add a comment here stating that setting `dse_top` to zero is just an artifact of how the routines behave, and is not supposed to be physically realistic.

src/physics/cam/vertical_diffusion.F90
```diff
@@ -281,59 +290,26 @@ subroutine vertical_diffusion_init(pbuf2d)
...
-     write(iulog,*)'Initializing vertical diffusion (vertical_diffusion_init)'
-     if (allocated(kvm_sponge)) then
-        write(iulog,*)'Artificial sponge layer vertical diffusion added:'
-        do k=1,size(kvm_sponge(:),1)
-           write(iulog,'(a44,i2,a17,e7.2,a8)') 'vertical diffusion coefficient at interface',k,' is increased by ', &
-                                                kvm_sponge(k),' m2 s-2'
-        end do
-     end if !allocated
-  end if
+  call vertical_diffusion_sponge_layer_init( &
```
nusbaume, line 305: I think we should check the value of `errflg` at the end of each CCPP-ized scheme call in this module, and if non-zero call `endrun` with the `errmsg` string included in the final end run message.

We could also call `handle_errmsg` instead of `endrun` as well, as long as we make sure that `handle_errmsg` properly ends the model simulation instead of just printing the error to the log.

nusbaume (changes requested):
Thanks for getting the CCPP-ized version of HB working in CAM @jimmielin!  My only major request is to make sure that all of the error codes/flags/messages from the CCPP-ized subroutines are being checked after they are called, in order to make sure we are catching fatal errors at the right location.

Otherwise I just had some (hopefully) minor requests, and possibly some comments that are already out-of-date (and so could just be resolved immediately).  Thanks again for seeing this through!

src/physics/cam/vertical_diffusion.F90
jimmielin: You're right, it seems like it was fixed in a later push https://github.com/jimmielin/CAM/commit/8b43f1d7b96a9c6181eea14c058f1e91f740d769

src/physics/cam/diffusion_solver_cam.F90
jimmielin: Seems to be `kg mol-1` as it is initialized to `mwdry`; backing out where it is calculated in `air_composition::get_mbarv_1hd`

```fortran
             mm = tracer(idx, kdx, m_cnst) * factor(idx, kdx)
             mbarv_in(idx, kdx) = mbarv_in(idx, kdx) + thermodynamic_active_species_mwi(qdx) * mm
! ...
       mbarv_in(:,:) = 1.0_r8 / mbarv_in(:,:)
```

if `[tracer] = [kg kg(air)-1]` and `[mwi] = [1 / mw] = [mol kg-1]` it also works out that `[mbarv] = [kg mol-1]` so it is molar mass.

`air_composition.F90` says
```fortran
   ! get_mbarv: molecular weight of dry air
   public :: get_mbarv
...

   ! mbarv: composition dependent atmosphere mean mass
   real(r8), public, protected, allocatable :: mbarv(:,:,:)
```

so I'll also change the comment of `mbarv` here to "mean molar mass" to be consistent

src/physics/cam/vertical_diffusion.F90
jimmielin: Thanks, added! I ended up just using `endrun`

src/physics/cam/vertical_diffusion.F90
jimmielin: Thanks, removed!

src/physics/cam/vertical_diffusion.F90
jimmielin: Thanks, added the same comment as in the atmos_phys code
```fortran
  ! Note: this is an artifact to the way the vertical diffusion calculations are done;
  ! it does not represent physical reality.
```

src/control/cam_snapshot_common.F90
jimmielin: Thanks for the suggestion! I _think_ the `real(tmpptr2d_int, r8)` will also make a temporary and compile to the same thing. Do you mind if I just kept it as-is? I don't have a strong argument for it other than I would have to go back and remake the snapshot to retest it to be sure it works as intended, but if you have a strong opinion on cleaning this up I'm happy to make the change!

nusbaume: If re-creating the snapshots and testing everything over again is a pain then I am totally fine to leave this as-is, as I do agree that the compiler will be doing the same thing under-the-hood.  Just note that for the (hopefully) longer-lived CAM-SIMA or atmospheric_physics repos I'll probably push harder against this.

Anyways, will go ahead and mark this comment as resolved now.  Thanks!

nusbaume (approved):
Everything looks good to me now.  Thanks again @jimmielin!

src/control/cam_snapshot_common.F90
jimmielin: Thanks @nusbaume! Yes I definitely agree with your change - my assumption is that this will be relatively short-lived. If I end up having to add more integer snapshot variables into this in the future I'll try to remember to make the suggested change.

Issue #1360, opened 2025-08-15
Excessive data movement in extend_panel_interpolate


johnmauff:
### What happened?

The subroutine extend_panel_interpolate is written such that the compiler will generate more data movement than is necessary.  This excessive data movement intensifies a computational load imbalance in the CSLAM advection.  While it is impossible to eliminate the load imbalance that is caused by the special treatment of panels at the corners of the cubed sphere, it is possible to reduce the cost of this subroutine by changing the way that the subroutine is written.  

### What are the steps to reproduce the bug?

Any CAM configuration that uses the spectral element dynamical core with CSLAM advection will encounter this issue.  

### What CAM tag were you using?

cam6_4_100

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Yes, but I will need some help

### Extra info

I have placed a modified version of fvm_reconstruction_mod.F90 in the directory ~dennis/CAM_mods/DataMovement/cam/src/dynamics/se/dycore

Note that this change is B4B, and it reduces the execution time of CAM in a likely CMIP configuration by ~3.5%

PeterHjortLauritzen:
PR opened: https://github.com/ESCOMP/CAM/pull/1365

cacraigucar:
Different approach adopted - closing

Issue #941, opened 2023-12-13
Runtime error when turning on threading in a CAM simulation


sjsprecious:
### What happened?

I tried to turn on the threading option in a CAM simulation (`F2000climo` compset, `ne30pg3` resolution). I used one compute node on Derecho with 64 MPI tasks and 2 threads per MPI task. It built successfully but I encountered lots of runtime errors (partials of them are listed below):
```
dec2481.hsn.de.hpc.ucar.edu 16: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 43: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 51: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 44: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 62: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 39: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 40: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 45: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 15: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 24: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 17: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 59: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 42: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 57: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 18: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 27: munmap_chunk(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 32: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 33: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 35: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 38: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 55: free(): invalid pointer
dec2481.hsn.de.hpc.ucar.edu 50: free(): invalid pointer
[...]

PR #1365, opened 2025-08-18
cam6_4_131: Performance improvements for CSLAM (from John Dennis, CISL, and Lauritzen)

PeterHjortLauritzen:
Closes https://github.com/ESCOMP/CAM/issues/1360


PeterHjortLauritzen:
I am not getting B4B in this test:

ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s

```
 ./create_test --output-root /glade/derecho/scratch/pel/ --project P93300042 ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s
```

(low top FHISTC with CSLAM)

@johnmauff: Could these changes be round-off (order of operation changes?)

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -960,19 +981,19 @@ subroutine extend_panel_interpolate(nc,nhc,nhr,nht,ns,nh,fcube,cubeboundary,halo
...
         do i=halo-nh,nc+nh-(halo-1)
           ibaseref=ibase(i,halo,1)!ibase(i,halo,2)
-          fpanel      (i,1-halo ) = matmul_w(w(:,i,halo),fcube(ibaseref:ibaseref+ns-1,1-halo),ns)  !south
+          fpanel      (i,1-halo ) = dotproduct(hWeight(:,i,halo,1),fcube(ibaseref:ibaseref+ns-1,1-halo),ns)  !south
         end do
       end do
       if (present(fotherpanel)) then
         fotherpanel (1-nht:nc+nht,1-nht:0 ,1)=fcube(1-nht:nc+nht,1-nht:0 )
         do halo=1,nhr
+          ftmp(:) = fcube(:,halo)
```
johnmauff: I just realized that the copy to ftmp is not needed here.

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -1299,20 +1317,19 @@ subroutine extend_panel_interpolate(nc,nhc,nhr,nht,ns,nh,fcube,cubeboundary,halo
       if (present(fotherpanel)) then
         fotherpanel(1-nht:nc,1-nht:0,1)  = fcube(1-nht:nc,1-nht:0)
         !
-        w = halo_interp_weight(:,:,:,2)
         !
         ! fill in "n" on Figure above
         !
         do halo=1,nhr
+          ftmp(:) = fcube(:,halo)
```
johnmauff: The ftmp(:) is not used.

johnmauff:
Peter,

  I would guess that they are round-off changes.  I have not check with a
low top model.  I only tested this with a high-top model, looking at the
nstep output.

John

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -640,43 +643,62 @@ subroutine slopelimiter_val(value, cell_value, local_min, local_max, min_phi)
...
+    real (kind=r8)                          :: dotproduct
+    real (kind=r8),dimension(:), intent(in) :: w,f      !dimension(ns)
+    integer,                     intent(in) :: ns
+    integer                                 :: k
+
+    if(ns==3) then
+      dotproduct = DotProduct_3(w,f)
+    else
+      dotproduct = DotProduct_gen(w,f,ns)
+    endif
```
sjsprecious, line 644: @johnmauff I guess the NBFB answer comes from this function? When `ns = 3`, the dot product is hard coded for performance optimization, but the truncation error will be different from the general version since they are now computed in a single line.

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
PeterHjortLauritzen: fixed

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
PeterHjortLauritzen: fixed

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
PeterHjortLauritzen: I tried the general algorithm but that did not change BFB differences with the baseline.

PeterHjortLauritzen:
> I am not getting B4B in this test:
> 
> ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s
> 
> ```
>  ./create_test --output-root /glade/derecho/scratch/pel/ --project P93300042 ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s
> ```
> 
> (low top FHISTC with CSLAM)
> 
> @johnmauff: Could these changes be round-off (order of operation changes?)

I ran the baroclinic wave test case (FKESSLER) and compared the optimized version against the baseline/trunk. Below is PS at day 10:

Screenshot 2025-08-20 at 11 56 55 AM

For comparison, here is a pertlim test (in this case perturbing PS by 1E-14):

Screenshot 2025-08-20 at 2 35 56 PM

The pertlim test produces errors about 100× smaller. It’s unclear whether it matters that the optimized code introduces round-off errors at every time step, whereas the pertlim test only introduces them at initialization. I'll keep looking/thinking ...

UPDATE: all tests were due to code bug ... all tests (I am running) are now BFB

sjsprecious:
@PeterHjortLauritzen I am curious: I thought threading was not supported for the SE dycore (https://github.com/ESCOMP/CAM/issues/941). Thus why the ERP test still works here?

sjsprecious:
@PeterHjortLauritzen I ran the `ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s` test on Derecho and found that the second run cut the node number by half but did not change the thread number (still 1). Thus this is actually an ERS test?

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -1774,25 +1783,25 @@ subroutine extend_panel_interpolate(nc,nhc,nhr,nht,ns,nh,fcube,cubeboundary,halo
...
-             fotherpanel(nc+1-halo,i,2) = matmul_w(w(:,i,halo),fcube(nc+1-halo,ibaseref:ibaseref+ns-1),ns)
+             fotherpanel(nc+1-halo,i,2) = dotproduct(hWeight(:,i,halo,1),ftmp(ibaseref:ibaseref+ns-1),ns)
            end do
          end do
          !
          ! fill in "n" on Figure above
          !
-         w = halo_interp_weight(:,:,:,2)
          do halo=1,nhr
+           ! ftmp(:) = fcube(nc+1-halo,:)   ! copy to a temporary
```
sjsprecious: @PeterHjortLauritzen @johnmauff I think this line should not be commented out?

johnmauff: Good catch.  I missed that.

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
sjsprecious: @PeterHjortLauritzen I was able to produce BFB result with the baseline after doing the following things:

1. Fixed the typo (confirm by @johnmauff ) at line 1797 (https://github.com/ESCOMP/CAM/pull/1365/files#diff-8beef36006cafdecbf26406cf4357fc0797eb6c7316d9c6aa0a2486fade88f25R1797).
2. Merged the latest `cam_development` branch into the `dennis_perf_cslam1` branch (currently the `dennis_perf_cslam1` branch is based on the tag cam6_4_089).

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
PeterHjortLauritzen: Thanks @sjsprecious ! Now the results make sense ... see plots above (in a couple of minutes)

PeterHjortLauritzen:
> @PeterHjortLauritzen I ran the `ERP_D_Ln9.ne30pg3_ne30pg3_mg17.FHISTC_LTso.derecho_intel.cam-outfrq9s` test on Derecho and found that the second run cut the node number by half but did not change the thread number (still 1). Thus this is actually an ERS test?

Don't know; I just took the test from the CAM test list without thinking much about the actual test (correct! Threading is currently broken in the SE dycore). @nusbaume Do you know the answer to @sjsprecious 's question?

PeterHjortLauritzen:
All the tests I am running are BFB now! Thanks @johnmauff  and @sjsprecious  ... @cacraigucar: This PR is ready to go ...

nusbaume:
Hi @PeterHjortLauritzen @sjsprecious, an ERP test takes the default thread and task count from the first case and divides it by 2 for the second case if the number is greater than one.  You can see that logic in the CIME code here:

https://github.com/ESMCI/cime/blob/master/CIME/SystemTests/erp.py#L32  

Given that the default configuration for an SE dycore run is one thread but multiple MPI tasks, it ends up adjusting the tasks but not the threads (which is what allows the test to run with this CAM configuration in the first place).

Also, my understanding is that the difference between ERS and ERP is that an ERS test won't change the task layout at all and simply checks that the restart run is bit-for-bit, while the ERP test will halve the processor count for the restarted run before checking if the results are bit-for-bit.

Anyways, I hope that helps, and thanks again for getting these improvements into CAM!

sjsprecious:
> Hi @PeterHjortLauritzen @sjsprecious, an ERP test takes the default thread and task count from the first case and divides it by 2 for the second case if the number is greater than one. You can see that logic in the CIME code here:
> 
> https://github.com/ESMCI/cime/blob/master/CIME/SystemTests/erp.py#L32
> 
> Given that the default configuration for an SE dycore run is one thread but multiple MPI tasks, it ends up adjusting the tasks but not the threads (which is what allows the test to run with this CAM configuration in the first place).
> 
> Also, my understanding is that the difference between ERS and ERP is that an ERS test won't change the task layout at all and simply checks that the restart run is bit-for-bit, while the ERP test will halve the processor count for the restarted run before checking if the results are bit-for-bit.
> 
> Anyways, I hope that helps, and thanks again for getting these improvements into CAM!

Thanks @nusbaume for your clarification. Clearly I misunderstood the ERS and ERP tests before, but now it is clear to me.

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
johnmauff, line 1: @PeterHjortLauritzen could you look at line 487 of this file?  The denominator is in a form that is inconsistent with other polynomial evaluations.  Is this missing a parens?  I.e. should it be  invtmp = 1.0_r8 / (recons(6,i,j) + spherecentroid(1,i,j))

johnmauff:
@pel I just noticed that commit [6067381](https://github.com/ESCOMP/CAM/commit/60673810aede57b3fc1139bcff3f4d9553982d17) appears to deleate all of my optimizations.  Is there a reason for this?

PeterHjortLauritzen:
> @pel I just noticed that commit [6067381](https://github.com/ESCOMP/CAM/commit/60673810aede57b3fc1139bcff3f4d9553982d17) appears to deleate all of my optimizations. Is there a reason for this?

apologies ... mistake ... reverted!

johnmauff:
Peter,

  Thanks for reverting the commit.  It looks like Rory's changes are only a
couple of lines.

John

cacraigucar:
I heard at a meeting the optimization work may be completed, with no deliverable for CAM (other than something which Peter will include in a future PR).  Is this correct, and should this PR be closed?

cacraigucar:
never mind - @PeterHjortLauritzen  is working on this

PeterHjortLauritzen:
FYI: performance results on this Wiki

https://github.com/PeterHjortLauritzen/CAM/wiki/Performance-notes-(2025)

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -418,127 +421,142 @@ subroutine slope_limiter(nhe,nc,nhc,fcube,jx,jy,irecons,recons,spherecentroid,re
...
-                recons(2:6,i,j)=min_phi*recons(2:6,i,j)
+                   r2 = recons(2,i,j,itr);  r3 = recons(3,i,j,itr)
+                   r4 = recons(4,i,j,itr);  r5 = recons(5,i,j,itr);  r6 = recons(6,i,j,itr)
+                   ! Check if the quadratic is minimized within the element
+                   ! Extrema in the interior of the element (there might be just one candidate)
+                   ! DO NOT NEED ABS here, if disc<0 we have a saddle point (no maximum or minimum)
+                   disc = 4.0_r8 * r4 *r5 - r6*r6
+                   if (abs(disc) > threshold) then
+!not b4b                      ex1 = (r6*r3 - 2.0_r8*r5*r2) / disc + scx
+!not b4b                      ex2 = (r6*r2 - 2.0_r8*r4*r3) / disc + scy
```
nusbaume: Remove commented-out code?

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -418,127 +421,142 @@ subroutine slope_limiter(nhe,nc,nhc,fcube,jx,jy,irecons,recons,spherecentroid,re
...
+                            max_val = max(max_val, v1)
+                            min_val = min(min_val, v1)
+                         endif
+                      enddo
+                   endif
+                   !
+                   ! Top/bottom edge, y=const., du/dx=0
+                   !
+                   if (abs(r4) > threshold) then
+                      invtmp = 1.0_r8 / (2.0_r8 * r4)! + spherecentroid(1,i,j)
```
nusbaume: Remove commented-out code here?


```suggestion
                      invtmp = 1.0_r8 / (2.0_r8 * r4)
```

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
```diff
@@ -418,127 +421,142 @@ subroutine slope_limiter(nhe,nc,nhc,fcube,jx,jy,irecons,recons,spherecentroid,re
...
+                            max_val = max(max_val, v1)
+                            min_val = min(min_val, v1)
+                         endif
+                      enddo
+                   endif
+                   !
+                   ! Top/bottom edge, y=const., du/dx=0
+                   !
+                   if (abs(r5) > threshold) then
+                      invtmp = 1.0_r8 / (2.0_r8 * r5)! + spherecentroid(1,i,j)
```
nusbaume: Remove commented-out code here?


```suggestion
                      invtmp = 1.0_r8 / (2.0_r8 * r5)
```

nusbaume (approved):
Thanks all!  I found a few lines of commented-out code that could potentially be removed, but otherwise everything looks good to me.

src/dynamics/se/dycore/fvm_reconstruction_mod.F90
PeterHjortLauritzen: @johnmauff thanks for noting this!

I re-did the math for computing the min/max values and I think you are right. The if (abs(recons(6,i,j)) > threshold) code block is buggy and actually redundant.

Details, if interested, below:

We wish to find the min/max values of a polynomial on the form

Screenshot 2025-11-24 at 12 48 15 PM


## First we find the interior critical points:

Compute the gradient and set it to zero

Screenshot 2025-11-24 at 12 51 04 PM

If

Screenshot 2025-11-24 at 12 52 03 PM

is non-zero

then the extrema are

Screenshot 2025-11-24 at 12 51 44 PM

These formulas are correct in the code.

## Evaluate on the Boundaries

The boundary consists of four line segments. On each, $f$ restricts to a 1D quadratic function, whose max/min occur at its critical point (if inside the segment) or at endpoints (corners).

Screenshot 2025-11-24 at 1 00 23 PM
Screenshot 2025-11-24 at 1 00 41 PM

All these formulas are correct in the code.

However, the code block with if (abs(recons(6,i,j)) > threshold) is buggy (as you notice) and I think redundant as the cases are already covered with the search above.

PeterHjortLauritzen: I have run some FKESSLER cases to verify what happens without the buggy if (abs(recons(6,i,j)) > threshold) section:

1. the  non-linear correlation between cl1 and cl2 is still maintained: cly=constant!
2. the min(upper plot)/max(lower plot) values for the slotted cylinders are pretty much the same.

Screenshot 2025-11-24 at 2 38 19 PM

(initial value: 0.100000000000000E+00; after 10 days:  0.999999999993388E-01 )

Screenshot 2025-11-24 at 2 38 44 PM

(initial value: 0.333333333333333E+00 0.333333333210990E+00 )

So no new over or undershoots introduced hence the " if (abs(recons(6,i,j)) > threshold)" code block can be removed!

The solutions are bit4bit until time-step 114 ()

<  nstep, te      114   0.26393936945915651E+10   0.26393936951915169E+10   0.32751870252974244E-07   0.99999998855001060E+05   0.20548077300190920E+03
>  nstep, te      114   0.26393936945915661E+10   0.26393936951915169E+10   0.32751818191091291E-07   0.99999998855001075E+05   0.20548077300190920E+03

i.e. bit4bit until day 2.375

PeterHjortLauritzen:
Thanks @nusbaume for the review. I have redone the math for finding extrema in CSLAM and found buggy code. I fixed it. Changes are not B4B but it takes over 2 days for differences to show up in FKESSLER. I did a science validation with FKESSLER (looking at traer properties) and things look good!

PR #1379, opened 2025-08-27
cam6_4_111: Making blending region consistent with p_top for radiation schemes

cacraigucar:
Contributions from @JulioTBacmeister


peverwhee (approved):

src/physics/cam/radheat.F90
```diff
@@ -123,7 +124,7 @@ subroutine radheat_init(pref_mid)
     else
        min_pressure_sw =  5._r8
        max_pressure_sw = 50._r8
```
adamrher: why even set min/max_pressure_sw variables in this version of radheat if they're not used?

adamrher: I think we should rip out all the nlte-related sw vairables in radheat.F90 since we are only blending tendencies for the lw terms This includes the above max/min vars, as well as ```qrs_wt```, ```midpoint_sw```, ```delta_merge_sw```.

adamrher (changes requested):

src/physics/cam/radheat.F90
peverwhee: removed!

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -575,9 +575,10 @@
 <rrtmgp_coefs_lw_file>src/physics/rrtmgp/data/rrtmgp-gas-lw-g128.nc</rrtmgp_coefs_lw_file>
 <rrtmgp_coefs_sw_file>src/physics/rrtmgp/data/rrtmgp-gas-sw-g112.nc</rrtmgp_coefs_sw_file>
 
-<p_top_for_rrtmgp>                 10.D0 </p_top_for_rrtmgp>
-<p_top_for_rrtmgp  waccmx="1" >     1.D0 </p_top_for_rrtmgp>
-<p_top_for_rrtmgp  waccm_phys="1" > 1.D0 </p_top_for_rrtmgp>
+<!-- Pressure top for blending region for radiation heating -->
+<p_top_for_equil_rad  rad="rrtmgp">    2.D0 </p_top_for_equil_rad>
```
cacraigucar: For consistency, I would put the "0" default in here as well

peverwhee: added!

Issue #1393, opened 2025-09-15
Add Spectral Nudging


patcal:
### What is the feature/what would you like to discuss?

This feature will add the ability to apply scale-selective filtering to applied nudging tendencies. This provides a methodology to constrain large-scale motions which evolve over longer time scales without corrupting the small scale convective parameterizations/processes of the model. Utilizing spherical harmonic amplitudes computed from unstructured grid point values, the applied nudging is independent of grid resolution. Thus, for example, equivalent large-scale nudging can be applied to grids with different refined resolution domains.  

### Will this change (regression test) answers?

Yes

### Will you be implementing this enhancement yourself?

Yes

cacraigucar:
@jtruesdal - should this have been closed with the recent PR?

jtruesdal:
@cacraigucar Yes.  I must not have included the correct syntax in the PR to automatically close.  I'll close it now.

Issue #1273, opened 2025-03-13
Nudging code doesn't error out when there are no nudging files


adamrher:
### What happened?

@mattldawson ran a nudging case but the model couldn't access the streams set by the ```nudge_path``` directory. The model did not error out and rather printed this message for each nudging time-step:

```
 NUDGING: WARNING - analyses file NOT FOUND. Switching
 NUDGING:           nudging OFF to coast thru the gap.
```

Coasting through the gaps can be useful, but it doesn't distinguish whether there are literally no nudging files for the entirety of a simulation. Ideally the model would error out if there are no nudging files.

### What are the steps to reproduce the bug?

```
./create_newcase --case $CASE_DIR --compset FMTHIST --res ne30pg3_ne30pg3_mg17 --walltime 00:29:00 --mach derecho --pecount 768 --project P19010000 --compiler gnu --queue main --run-unsupported

./xmlchange RUN_STARTDATE=2010-09-01

open up user_nl_cam w/ an editor and add these entries:
model_times_per_day            = 24
nudge_beg_day          = 1
nudge_beg_month                = 9
nudge_beg_year         = 2010
nudge_end_day          = 7
nudge_end_month                = 9
nudge_end_year         = 2010
nudge_path                     ='/glade/derecho/scratch/aherring/restart/nudging/ne30pg3/L93/2010/'
nudge_file_template            ='ERA5_x_ne30pg3_L93_rgC1_WO.%y-%m-%d-%s.nc'
nudge_force_opt                = 0
nudge_hwin_invert              = .false.
nudge_hwin_lat0                = 0.0
nudge_hwin_latdelta            = 1.0
nudge_hwin_latwidth            = 9999.0
nudge_hwin_lon0                = 180.
nudge_hwin_londelta            = 1.0
nudge_hwin_lonwidth            = 9999.0
nudge_model            = .true.
nudge_pscoef           = 0
nudge_psprof           = 0
nudge_qcoef            = 0.0
nudge_qprof            = 2
nudge_tcoef            = 0.16667
nudge_times_per_day            = 24
nudge_timescale_opt            = 0
nudge_tprof            = 2
nudge_ucoef            = 0.16667
nudge_uprof            = 2
nudge_vcoef            = 0.16667
nudge_vprof            = 2
nudge_vwin_hdelta              = 0.001
nudge_vwin_hindex              = 0.
nudge_vwin_invert              = .true.
nudge_vwin_ldelta              = 0.001
[...]

PR #1402, opened 2025-09-30
cam6_4_191: Add Spectral Nudging Option and Fix Nudging Behavior When Files Are Missing

patcal:
This pull request addresses issues #1393 and #1273

For Issue #1273 an error exit with a message is invoked if a missing nudging file is encountered.

For Issue #1393 a module for computing spherical harmonics on the physics grid is implemented. The nudging module has additional functionality that uses this module to optionally filter nudging tendencies to scales larger then a specified truncation scale. 

This functionality is controlled by 2 additional variables in the &nudging_nl namelist. 
**Nudge_SpectralFilter**- LOGICAL Option to apply spherical harminic filtering to 
                                                          the model state and target data so that nudging 
                                                          tendencies are only applied to scales larger than
                                                           the specified truncation.

**Nudge_SpectralNtrunc** - INT The number of meridional spherical harmonic modes used 
                                                     for spectral filtering. The nominal horizontal scale of 
                                                     the filtering can be estimated as:

                                                              Hscale = PI*6350/Nudge_SpectralNtrunc

                                              i.e. Nudge_SpectralNtrunc=40 corresponds to a horizontal 
                                                     nudging scale  Hscale~500km.


src/physics/cam/nudging.F90
```diff
@@ -273,14 +290,21 @@ module nudging
...
+  !-----------------------------------------
   logical             :: Nudge_ZonalFilter =.false.
   integer             :: Nudge_ZonalNbasis = -1
   type(ZonalMean_t)   :: ZM
   real(r8),allocatable:: Zonal_Bamp2d(:)
   real(r8),allocatable:: Zonal_Bamp3d(:,:)
 
+  logical             :: Nudge_SpectralFilter =.false.
+  integer             :: Nudge_SpectralNtrunc = -1
+  integer             :: Nudge_SpectralNbasis = -1
```
fvitt, line 382: I am not seeing where `Nudge_SpectralNbasis` gets sets.  Is it always `-1`?

src/physics/cam/nudging.F90
```diff
@@ -1286,9 +1348,11 @@ subroutine nudging_timestep_init(phys_state)
      endif
      if(.not.Nudge_ON) then
        if(masterproc) then
-         write(iulog,*) 'NUDGING: WARNING - analyses file NOT FOUND. Switching '
-         write(iulog,*) 'NUDGING:           nudging OFF to coast thru the gap. '
+         write(iulog,*) 'NUDGING: WARNING - analyses file NOT FOUND. You can switch nudging '
+         write(iulog,*) 'NUDGING:           OFF to coast thru a known gap in your files '
+         write(iulog,*) 'NUDGING:           by commenting out the following endrun command.'
        endif
+       call endrun('nudging_timestep_init:: ERROR Missing Nudging File')
```
fvitt, line 1770: Should a namelist switch to turn on / off the abort behavior?  I agree that aborting the run should be the default behavior.

src/utils/spherical_harmonic_mod.F90
```diff
@@ -0,0 +1,1235 @@
...
+         call endrun(subname//': DGESV error return: matrix is singular')
+      end if
+
+      deallocate(Indx)
+
+    end subroutine Invert_Matrix
+    !=======================================================================
+
+    !=======================================================================
+    ! legacy spherepack routines
```
fvitt: As far as I can tell, these spherepack routines are duplicates of the spherepack routines in `zonal_mean_mod.F90`.  If so, the spherepack routines in `zonal_mean_mod` should be removed and  `zonal_mean_mod` should use the routines here.

src/utils/spherical_harmonic_mod.F90
```diff
@@ -0,0 +1,1235 @@
...
+
+      if(allocated(this%area )) deallocate(this%area)
+      if(allocated(this%basis)) deallocate(this%basis)
+
+    end subroutine final_SphericalHarmonic
+    !=======================================================================
+
+
+    !=======================================================================
+    subroutine Invert_Matrix(I_Mat,Nbas,O_InvMat)
```
fvitt: This looks like a duplicate of `Invert_Matrix` in `zonal_mean_mod`

src/utils/spherical_harmonic_mod.F90
```diff
@@ -0,0 +1,1235 @@
...
+  use cam_logfile,     only: iulog
+
+  implicit none
+  private
+
+  public :: SphericalHarmonic_t
+
+  ! Type definitions
+  !-------------------
+  type SphericalHarmonic_t
```
fvitt: Could the existing `ZonalMean` class be replaced by "special" use of this `SphericalHarmonic` class (m=0)?

fvitt (commented):
Some initial questions and comments.  Do you have a test case for spectral nudging that you can point me to?  I can implement a regression test case.

jtruesdal:
Thanks @patcal I'll see if I can update this to the head of the trunk.

src/physics/cam/nudging.F90
```diff
@@ -149,6 +151,35 @@ module nudging
 !                              0 -->  TimeScale = 1/Tdlt_Anal                      [DEFAULT]
 !                              1 -->  TimeScale = 1/(t'_next - t_curr )
 !
+!      Nudge_SpectralFilter - LOGICAL Option to apply spherical harminic filtering to 
+!                                     the model state and target data so that nudging 
+!                                     tendencies are only applied to scales larger than
+!                                     the specified truncation.
+!
+!      Nudge_SpectralNtrunc - INT The number of meridional spherical harmonic modes used 
+!                                 for spectral filtering. The nominal horizontal scale of
```
jtruesdal:
```suggestion
!                                 for spectral filtering. The nominal horizontal scale (km) of 
```

src/physics/cam/nudging.F90
```diff
@@ -273,14 +304,22 @@ module nudging
...
+  ! Nudging Zonal/Spectral Filter variables
+  !-----------------------------------------
   logical             :: Nudge_ZonalFilter =.false.
   integer             :: Nudge_ZonalNbasis = -1
   type(ZonalMean_t)   :: ZM
   real(r8),allocatable:: Zonal_Bamp2d(:)
   real(r8),allocatable:: Zonal_Bamp3d(:,:)
 
+  logical             :: Nudge_SpectralFilter =.false.
+  integer             :: Nudge_SpectralNtrunc = -1
```
jtruesdal, line 381: Nudge_SpectralNtrunc should have a check in readnl to make sure its set to something other than the default of -1 along with a message letting the user know  about the missing truncation value before attempting the initialization

bld/namelist_files/namelist_definition.xml
```diff
@@ -220,6 +220,53 @@
...
+                                     Hscale = PI*6350/Nudge_SpectralNtrunc
+
+                             i.e. Nudge_SpectralNtrunc=40 corresponds to a horizontal
+                                  nudging scale  Hscale~500km.
+       Default: -1
+       </entry>
+
+<entry id="Nudge_SpectralNring" type="integer" category="nudging"
+       group="nudging_nl" valid_values="" >
+       Set size of local-area grid used to suppress Samping Errors.
```
jtruesdal: Samping -> Sampling here and another 3 times below.

src/utils/ug_spectralmethods_mod.F90
```diff
@@ -0,0 +1,3755 @@
...
+            do lchnk=begchunk,endchunk
+               ncols = get_ncols_p(lchnk)
+               do cc = 1,ncols
+                  count=count+1
+                  Csum(count,nn) = I_Gdata(cc,ll,lchnk)*this%basis(cc,lchnk,nn)*this%area(cc,lchnk)
+               end do
+            end do
+         end do
+
+         call shr_reprosum_calc(Csum, Bamp, count, nlcols, this%nbas, gbl_count=ngcols_p, commid=mpicom)
```
jtruesdal: This shr_reprosum_calc call within the level loop will be a bottleneck.  Can you compute the local covariance with input data and basis functions for all levels first and then make a single reprosum call.  Something like this. Not tested.

! Flatten the vertical and basis dimensions to allow a single MPI reduction
nbas_x_nlev = this%nbas * nlev
allocate(Bamp_flat(nbas_x_nlev), stat=astat)
allocate(Csum(nlcols, nbas_x_nlev), stat=astat)

do ll= 1,nlev 
  do nn= 1,this%nbas 
    idx = (ll - 1) * this%nbas + nn 
    count = 0 
    do lchnk=begchunk,endchunk 
      ncols = get_ncols_p(lchnk) 
      do cc = 1,ncols 
        count=count+1 
        Csum(count,idx) = I_Gdata(cc,ll,lchnk)*this%basis(cc,lchnk,nn)*this%area(cc,lchnk) 
       end do 
     end do 
   end do 
 end do

call shr_reprosum_calc(Csum, Bamp_flat, count, nlcols, nbas_x_nlev, gbl_count=ngcols_p, commid=mpicom)

! Unpack the flattened amplitudes into the 2D output array
do ll= 1,nlev
   do nn=1,this%nbas
      idx = (ll - 1) * this%nbas + nn
      O_Bamp(nn,ll) = Bamp_flat(idx)
    end do
  end do

src/utils/ug_spectralmethods_mod.F90
```diff
@@ -0,0 +1,3755 @@
...
+      allocate(Csum(nlcols, this%nbas), stat=astat)
+      call handle_allocate_error(astat, subname, 'Csum')
+
+      Csum(:,:) = 0._r8
+      O_Bamp(:,:) = 0._r8
+
+      ! Compute Covariance with input data and basis functions
+      !--------------------------------------------------------
+      do ll= 1,nlev
+
```
jtruesdal: Pull reprosum out of level loop like in calc_SphericalHarmonic_3Damps_GS. Since  this%map(n2, nn) is applied after the reduction, you can unpack Gcov into a temporary Gcov_by_level(nbas) array for each level, and then perform the map multiplication against Gcov_by_level.

src/utils/ug_spectralmethods_mod.F90
```diff
@@ -0,0 +1,3755 @@
...
+            do lchnk=begchunk,endchunk
+               ncols = get_ncols_p(lchnk)
+               do cc = 1,ncols
+                  count=count+1
+                  Csum(count,nn) = I_Gdata(cc,ll,lchnk)*this%basis(cc,lchnk,nn)*this%area(cc,lchnk)
+               end do
+            end do
+         end do
+
+         call shr_reprosum_calc(Csum, Gcov, count, nlcols, this%nbas, gbl_count=ngcols_p, commid=mpicom)
```
jtruesdal: Same as other 2 3d routines, pull shr_reprosum_calc out of level loop.

jtruesdal (changes requested):
This is a nice refactor.  I didn't test the code to pull shr_reprosum out of the level loop it was just an example but something like that should work.

bld/namelist_files/namelist_definition.xml
patcal: Corrected

src/physics/cam/nudging.F90
patcal: Corrected

src/physics/cam/nudging.F90
patcal: Hi @fvitt , for the spherical harmonics, the user specifies the meridional truncation(L). For each index (l < L+1)  there are (2l+1) basis functions, so the init() routine sums up the total number of these basis functions and returns the value into Nudge_SpectralNbasis.

src/physics/cam/nudging.F90
patcal: We could add that, but I don't know if it would be worth the effort. The only case I know of where it was needed was for YOTC data which had processing errors that led to some randomly dispersed gaps in the available data. So it is not a common problem, and most users who have a gap in there data will figure out a way to fill in their missing data. Also, allowing for this option would make the pending implementation of higher order time interpolation methods a bit more complex as they would have to account for the gaps.

src/utils/ug_spectralmethods_mod.F90
patcal: Implemented

src/utils/ug_spectralmethods_mod.F90
patcal: Implemented this here and in other similar use cases. At L58, the performance improvement was not substantial, but it is a good fix and should be of more benefit for higher vertical resolution cases.

src/utils/ug_spectralmethods_mod.F90
patcal: Implemented

src/utils/ug_spectralmethods_mod.F90
jtruesdal: @patcal I don't see the changes to calc_SphericalHarmonic_3Damps_GS (and the _COV equivalent)  routines.  Maybe you didn't push the changes?

src/utils/ug_spectralmethods_mod.F90
jtruesdal: Same as above, I don't see the mods.

src/utils/ug_spectralmethods_mod.F90
jtruesdal: Same as above.  I don't see the mods.

patcal:
Hi John,

   Sorry for the confusion. The last test I ran last night before pushing
the mods had a 50% increase in runtime, so something I did is causing
problems that I am trying to undo. I thought I might be able to get it
sorted out
before this morning. I will go ahead and update with the current files
shortly
so you and Francis can go through the revisions. The problem seems to be
isolated to one use case, so I will need to add a patch for nudging.F90 to
fix
cases with user-defined nudging windows (Nudge_Xprof=3 cases) as soon
as I get it worked out.

--> Patrick

jtruesdal:
Thanks Pat.

jt

src/utils/ug_spectralmethods_mod.F90
jtruesdal: I see it now.  Looks good.

src/utils/ug_spectralmethods_mod.F90
jtruesdal: I see it now.  Looks good.

src/utils/ug_spectralmethods_mod.F90
jtruesdal: I see it now.  Looks good.

jtruesdal (approved):
All looks good now for me.

patcal:
Thanks John,

  I will try and get the Nudge_Xprof=3 problem resolved today.

--> Patrick

src/physics/cam/nudging.F90
jtruesdal: @patcal @fvitt was there ever a regression test added so we can verify if something gets broken in a future commit.

fvitt: We have these aux_cam tests:
```
aux_cam           : ERP_Ln9.ne30pg3_ne30pg3_mg17.FCnudged.derecho_intel.cam-outfrq9s
aux_cam           : SMS_D_Ln9.f09_f09_mg17.FCts2nudged.derecho_intel.cam-outfrq9s_leapday
```
and these prealpha and prebeta tests:
```
prealpha          : ERP_Ln9.f09_f09_mg17.FCnudged_GC.derecho_intel.cam-outfrq9s
prealpha          : SMS_D_Ln9_P1536x1.ne0CONUSne30x8_ne0CONUSne30x8_mt12.FCnudged.derecho_intel.cam-outfrq9s                                                   # "Nudged CAM-Chem on a regionally refined grid"
prebeta           : SMS_D_Ln9.f09_f09_mg17.FCnudged.derecho_intel.cam-outfrq3s_zmean_nudging                                                                   # Zonal Mean Nudged
```

src/physics/cam/nudging.F90
fvitt: OK. Thanks for explaining.

fvitt (approved):
Looks good to me, assuming the nudged regression tests pass if they are expected to.

PR #1404, opened 2025-09-30
cam6_4_135: Abstract aerosol interfaces extended for bulk representation

### LINKED ISSUE #1400: Extend abstract interfaces for bulk aerosol optics
### Issue Type

Infrastructure Update

### Issue Description

Extend the abstract aerosol interfaces for bulk aerosol representations for optical properties of aerosols, including volcanic / stratospheric aerosols.

Expect round-off level differences in optical properties for configurations that BAM and volcanic aerosols.

### Will this change answers?

Yes

Yes

fvitt:
For the purpose of aerosol optical properties input into radiative transfer, aerosol interfaces have been extended for bulk aerosols.

closes #1400 

Roundoff level answer changing.


src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+       nullify(newobj)
+       return
+    end if
+    allocate( f1(naero),stat=ierr )
+    if( ierr /= 0 ) then
+       nullify(newobj)
+       return
+    end if
+
+    nspecies(:) = 1
```
jimmielin, line 100: Does `nspecies(:naero) = 1` here mean that each bulk aerosol specified in the namelist has 1 tracer corresponding to it?

The definition in the abstract `aerosol_properties_mod.F90` just says "number of species" and perhaps it could be clarified for folks like me who are new to this object. Thanks!

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+
+    alogsig(:) = log(2._r8)
```
jimmielin, line 103: Should this be made as a `parameter` and then assigned to `alogsig`?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  end subroutine destructor
+
+
+  !------------------------------------------------------------------------------
+  ! returns number of transported aerosol constituents
+  !------------------------------------------------------------------------------
+  integer function number_transported(self)
+    class(bulk_aerosol_properties), intent(in) :: self
+    ! to be implemented later
+    number_transported = -1
```
jimmielin, line 135: It seems like this is to be implemented in both the modal and CARMA objects as well but it seems trivial to implement for bulk aerosols, maybe it could be assigned as an example for future use?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  !------------------------------------------------------------------------------
+  pure elemental real(r8) function amcube(self, bin_ndx, volconc, numconc)
+
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx  ! bin number
+    real(r8), intent(in) :: volconc ! volume conc (m3/m3)
+    real(r8), intent(in) :: numconc ! number conc (1/m3)
+
+    !    call endrun('ERROR: bulk_aerosol_properties_mod%amcube not yet implemented')
+    amcube = -huge(1._r8)
```
jimmielin: Maybe add a comment here that `amcube` cannot be defined for bulk aerosols since there are no number concentrations in the bulk representation?

I understand `endrun` cannot be called here because this implements a `pure` function so maybe a comment of why the intentionally invalid value is returned here will be useful, to distinguish between 'does not exist for bulk representation' and 'not yet implemented'

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  subroutine actfracs(self, bin_ndx, smc, smax, fn, fm )
+
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx   ! bin index
+    real(r8),intent(in) :: smc       ! critical supersaturation for particles of bin radius
+    real(r8),intent(in) :: smax      ! maximum supersaturation for multiple competing aerosols
+    real(r8),intent(out) :: fn       ! activation fraction for aerosol number
+    real(r8),intent(out) :: fm       ! activation fraction for aerosol mass
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%actfracs not yet implemented')
```
jimmielin, line 423: Is there a plan to implement this in the future to possibly use it in `ndrop_bam_run` like how it is used in `ndrop::dropmixnuc`? Might be too much work

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  !------------------------------------------------------------------------
+  ! returns constituents names of aerosol number mixing ratios
+  !------------------------------------------------------------------------
+  subroutine num_names(self, bin_ndx, name_a, name_c)
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+    character(len=*), intent(out) :: name_a ! constituent name of ambient aerosol number dens
+    character(len=*), intent(out) :: name_c ! constituent name of cloud-borne aerosol number dens
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%num_names not yet implemented')
```
jimmielin, line 437: I suppose since `number mixing ratios` do not exist in the bulk representation could write it as

```suggestion
    call endrun('ERROR: bulk_aerosol_properties_mod%num_names does not exist in bulk representation')
```

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  ! returns constituents names of aerosol mass mixing ratios
+  !------------------------------------------------------------------------
+  subroutine mmr_names(self, bin_ndx, species_ndx, name_a, name_c)
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+    integer, intent(in) :: species_ndx       ! species number
+    character(len=*), intent(out) :: name_a ! constituent name of ambient aerosol MMR
+    character(len=*), intent(out) :: name_c ! constituent name of cloud-borne aerosol MMR
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%mmr_names not yet implemented')
```
jimmielin, line 452: Could this be implemented with something like
```fortran
    call rad_cnst_get_info(list_idx, aernames=tmp_names)
    name_a = tmp_names(species_ndx)
```
or the information isn't available in `rad_constituents` anyway for bulk masses?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+
+  !------------------------------------------------------------------------
+  ! returns constituent name of ambient aerosol number mixing ratios
+  !------------------------------------------------------------------------
+  subroutine amb_num_name(self, bin_ndx, name)
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+    character(len=*), intent(out) :: name   ! constituent name of ambient aerosol number dens
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%amb_num_name not yet implemented')
```
jimmielin, line 465: Similarly here
```suggestion
    call endrun('ERROR: bulk_aerosol_properties_mod%amb_num_name does not exist in bulk representation')
```

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  !------------------------------------------------------------------------
+  ! returns species type
+  !------------------------------------------------------------------------
+  subroutine species_type(self, bin_ndx, species_ndx, spectype)
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+    integer, intent(in) :: species_ndx       ! species number
+    character(len=*), intent(out) :: spectype ! species type
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%species_type not yet implemented')
```
jimmielin: Maybe the code in lines 170-186 to return `spectype` could be put here so this is implemented? Not sure if it is the same `spectype` since it's not from rad_cnst directly.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  !------------------------------------------------------------------------------
+  subroutine apply_number_limits( self, naerosol, vaerosol, istart, istop, m )
+    class(bulk_aerosol_properties), intent(in) :: self
+    real(r8), intent(inout) :: naerosol(:)  ! number conc (1/m3)
+    real(r8), intent(in)    :: vaerosol(:)  ! volume conc (m3/m3)
+    integer,  intent(in) :: istart          ! start column index (1 <= istart <= istop <= pcols)
+    integer,  intent(in) :: istop           ! stop column index
+    integer,  intent(in) :: m               ! mode or bin index
+
+    call endrun('ERROR: bulk_aerosol_properties_mod%apply_number_limits not yet implemented')
```
jimmielin, line 533: Maybe could just be set to `.false.` since number does not exist for bulk?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+       nullify(newobj)
+       return
+    end if
+
+    nspecies(:) = 1
+
+    alogsig(:) = log(2._r8)
+    f1(:) = 1._r8
+
+    call newobj%initialize(naero, naero, nspecies, nspecies, alogsig, f1, f1, ierr)
```
jimmielin: Should the first argument to %initialize here (corresponding to `nbin`) be `1` or `naero`? From how it is called in `modal_aerosol_properties_mod` as `newobj%initialize(nmodes,ncnst_tot,...` I am not sure what is the actual size of the arrays that should be allocated inside the aerosol properties object in the case of bulk aerosols.

This and the question on `nspecies(:naero) = 1` I think pertain to my lack of understanding of how `nbin, ncnst, nspec, nmasses` are defined for different cases of bulk, modal, and sectional aerosol representations in the abstract interface (I think I am just confused by the use of `nbin` for non-binned representations) so I would appreciate a few comments here to clarify the intent of the design. Thanks!!

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+    if( ierr /= 0 ) then
+       nullify(newobj)
+       return
+    end if
+    deallocate(nspecies)
+    deallocate(alogsig)
+    deallocate(f1)
```
jimmielin: Could the `deallocate` calls be moved above the `ierr` checking to avoid memory leaks for `nspecies, alogsig, f1` if `ierr /= 0` and the constructor returns early?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,695 @@
...
+  end function hetfrz_species
+
+  !------------------------------------------------------------------------------
+  ! returns TRUE if soluble
+  !------------------------------------------------------------------------------
+  logical function soluble(self,bin_ndx)
+    class(bulk_aerosol_properties), intent(in) :: self
+    integer, intent(in) :: bin_ndx           ! bin number
+
+    soluble=.false.
```
jimmielin: I think some bulk aerosols are soluble (e.g., sulfate)? Or is this set to `.false.` to intentionally make incompatible with `watact_mfactor`?

src/chemistry/aerosol/bulk_aerosol_state_mod.F90
```diff
@@ -0,0 +1,426 @@
...
+  ! returns ambient aerosol number mixing ratio for a given species index and bin index
+  !------------------------------------------------------------------------------
+  subroutine get_ambient_num(self, bin_ndx, num)
+    class(bulk_aerosol_state), intent(in) :: self
+    integer, intent(in) :: bin_ndx     ! bin index
+    real(r8), pointer   :: num(:,:)    ! number densities
+
+    nullify(num)
+
+    call endrun('ERROR: bulk_aerosol_state_mod%get_ambient_num not yet implemented')
```
jimmielin, line 177: I would suggest returning the error as this cannot be defined for bulk
```suggestion
    call endrun('ERROR: bulk_aerosol_state_mod%get_ambient_num cannot be defined for bulk representation')
```

src/physics/cam/aerosol_optics_cam.F90
```diff
@@ -763,17 +820,35 @@ subroutine aerosol_optics_cam_sw(list_idx, state, pbuf, nnite, idxnite, tauxar,
...
              aero_optics=>hygrocoreshell_aerosol_optics(aeroprops, aerostate, list_idx, &
                                                         ibin, ncol, pver, relh(:ncol,:))
           case('hygroscopic_wtp')
              aero_optics=>hygrowghtpct_aerosol_optics(aeroprops, aerostate, list_idx, &
                                                       ibin, ncol, pver, sulfwtpct(:ncol,:))
+          case('hygro')
+             aero_optics=>hygro_aerosol_optics(aeroprops, aerostate, list_idx, &
+                                                     ibin, ncol, pver, numrh, relh(:ncol,:))
+          case('hygroscopic')
+             aero_optics=>hygroscopic_aerosol_optics(aeroprops, aerostate, list_idx, &
```
jimmielin, line 835: The difference between `hygro` and `hygroscopic` is quite confusing. Maybe copy the comment from the top of `hygro_aerosol_optics_mod` and `hygroscopic_aerosol_optics_mod` to here
```suggestion
          case('hygro')
          ! Short-wave hygroscopic aerosol, Long-wave non-hygroscopic
             aero_optics=>hygro_aerosol_optics(aeroprops, aerostate, list_idx, &
                                                     ibin, ncol, pver, numrh, relh(:ncol,:))
          case('hygroscopic') 
          ! Short-wave and long-wave hygroscopic
             aero_optics=>hygroscopic_aerosol_optics(aeroprops, aerostate, list_idx, &
```

src/chemistry/aerosol/bulk_aerosol_state_mod.F90
```diff
@@ -0,0 +1,426 @@
...
+  ! returns cloud-borne aerosol number mixing ratio for a given species index and bin index
+  !------------------------------------------------------------------------------
+  subroutine get_cldbrne_num(self, bin_ndx, num)
+    class(bulk_aerosol_state), intent(in) :: self
+    integer, intent(in) :: bin_ndx             ! bin index
+    real(r8), pointer :: num(:,:)
+
+    nullify(num)
+
+    call endrun('ERROR: bulk_aerosol_state_mod%get_cldbrne_num not yet implemented')
```
jimmielin, line 191: Similarly I would suggest returning the error as this cannot be defined for bulk instead of not yet implemented (as it never will)
```suggestion
   call endrun('ERROR: bulk_aerosol_state_mod%get_cldbrne_num cannot be defined for bulk representation')
```

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -169,6 +173,25 @@ subroutine aero_optics_params(self, list_ndx, bin_ndx, opticstype, extpsw, absps
...
+       real(r8),  optional, pointer :: sw_hygroscopic_ext(:,:)
+       real(r8),  optional, pointer :: sw_hygroscopic_ssa(:,:)
+       real(r8),  optional, pointer :: sw_hygroscopic_asm(:,:)
+       real(r8),  optional, pointer :: lw_hygroscopic_ext(:,:)
+
+       ! non-hygroscopic
+       real(r8),  optional, pointer :: sw_nonhygro_ext(:)
+       real(r8),  optional, pointer :: sw_nonhygro_ssa(:)
+       real(r8),  optional, pointer :: sw_nonhygro_asm(:)
+       real(r8),  optional, pointer :: lw_nonhygro_ext(:)
```
jimmielin: Some suggested comment additions
```suggestion
       ! hygroscopic - Short-wave and long-wave hygroscopic
       real(r8),  optional, pointer :: sw_hygroscopic_ext(:,:) ! short wave extinction table
       real(r8),  optional, pointer :: sw_hygroscopic_ssa(:,:) ! short wave single-scatter albedo table
       real(r8),  optional, pointer :: sw_hygroscopic_asm(:,:) ! short wave asymmetry table
       real(r8),  optional, pointer :: lw_hygroscopic_ext(:,:) ! long wave absorption table

       ! non-hygroscopic (insoluble)
       real(r8),  optional, pointer :: sw_nonhygro_ext(:) ! short wave extinction table
       real(r8),  optional, pointer :: sw_nonhygro_ssa(:) ! short wave single-scatter albedo table
       real(r8),  optional, pointer :: sw_nonhygro_asm(:) ! short wave asymmetry table
       real(r8),  optional, pointer :: lw_nonhygro_ext(:) ! long wave absorption table
```


Also, the nonhygro aerosol optics mod is named `insoluble_aerosol_optics_mod` but the arrays here are named `nonhygro`. Maybe they could be made consistent.

src/chemistry/aerosol/insoluble_aerosol_optics_mod.F90
```diff
@@ -0,0 +1,118 @@
+!-------------------------------------------------------------------------------
+! Insoluble (non-hygrostatic) aerosol optical properties
```
jimmielin: Nitpick for typo
```suggestion
! Insoluble (non-hygroscopic) aerosol optical properties
```

jimmielin (changes requested):
Thanks @fvitt for implementing the bulk aerosol properties and state module and other abstractions! I have mostly clarification questions and some minor suggestions.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
gold2718: It seems to me that the need to look for workarounds like this is a sign that something is wrong with the abstract aerosol class structure. In this case, the abstract properties class is requiring information that is not present in all aerosol models. 
In a perfect world, all code that used `%amcube` (and the other problematic methods below) would reside in the appropriate subclass and all uses would be restricted to that subclass (which could still be a superclass of final models if a set of aerosol models shared a property such as `amcube`).
I don't see CAM (or CAM-SIMA) getting ready for such a clean class hierarchy soon so in the interim, I think it would be best if the abstract base class (`aerosol_properties`) provided a default, overridable method to produce an invalid value. I am happy to present this change as a PR to this PR branch so you can evaluate it.

Speaking of an invalid value, what happens if `<bulk_aero_prop_obj>%amcube` gets called? The code where it is called has no value checks.

gold2718 (changes requested):
I would like to see a bit of a shift in how the non-bulk methods are implemented (see below).

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: @jimmielin and @gold2718 
Many of these undefined methods (such as `amcube`) could probably be implemented for bulk aerosols.  They are not needed for optical properties of prescribed bulk aerosols, which is the purpose of this PR.  In the interest of time, I did not make the effort.   They can be implemented later if needed.

gold2718: @fvitt, thanks for the explanation. Would it be possible to set the value to `NaN` instead of `-huge`? That way we would know if any not-implemented method was used in active code.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: Bulk aerosol have one species per bin.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: What will we gain be inserting a parameter here?

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: This is outside the scope of this PR

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: For this PR it is not needed.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
jimmielin: One aspect I struggled when reviewing the PR is that perhaps there could be better distinction of what is not implemented _yet_ (as @fvitt notes as the scope of this PR) versus what cannot be physically defined for bulk aerosols (as @gold2718 notes as "information that is not present in all aerosol models"). Either `-huge` or `NaN` serve as good magic numbers for me but I believe the comments should make the intent of the code clear.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
jimmielin: > What will we gain be inserting a parameter here?

I'm just not sure where this magic number is coming from, so a comment for at least the value would help. It seems to be the same value as CARMA, and MAM has species-specific deviations.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: Added some comments in the code here

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: This may never be used for bulk aerosols.  The settings are adopted from CARMA properties. Comments are added.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: Now using NaN, as requested by @gold2718 .

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: I just use the `%get` method here.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: Added some comments in the code which I hope are helpful.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: Done

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: I tried to implement the `soluble` and `hydrophilic` methods. However I don't have test cases for these.

src/physics/cam/aerosol_optics_cam.F90
fvitt: I agree the naming is confusing here. These are the optics method names in the phys props files. I am not up changing the phys props files at this point.

I added comments, as you suggested.  Hopefully this helps.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Renamed the arguments and added comments.

src/chemistry/aerosol/insoluble_aerosol_optics_mod.F90
fvitt: fixed

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
jimmielin: Thanks @fvitt!

jimmielin (approved):

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -74,6 +74,7 @@ module aerosol_properties_mod
      procedure(aero_resuspension_resize), deferred :: resuspension_resize
      procedure(aero_rebin_bulk_fluxes), deferred :: rebin_bulk_fluxes
      procedure(aero_hydrophilic), deferred :: hydrophilic
+     procedure :: is_bulk
```
gold2718: It would be really nice if this could instead be a general term or if terms could be added for other types. One suggestion is:
```suggestion
     procedure ::aerosol_is
```
`aerosol_is` could function similarly to how `dycore_is` is used today. This could then be used to identify the bulk model but also could replace current model tests such as `(nmodes>0)` (currently used to identify MAM but collides with our modal and sectional models) or `(nbins>0)` (currently used to identify CARMA but also collides with our sectional model).

I'm happy to submit this as an independent PR or as a PR to this PR branch.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -101,7 +102,7 @@ end function aero_number_transported
      !  species morphology
      !------------------------------------------------------------------------
      subroutine aero_props_get(self, bin_ndx, species_ndx, list_ndx, density, hygro, &
-                               spectype, specname, specmorph, refindex_sw, refindex_lw)
+          spectype, specname, specmorph, refindex_sw, refindex_lw)
```
gold2718, line 105: I still think each of these properties should be a separate method which would simplify use and maintenance but I can pitch this as an independent PR later if that sounds plausible.

cacraigucar:
@gold2718 - Do you have any further feedback, or are you okay with us proceeding with tagging this?

gold2718:
> @gold2718 - Do you have any further feedback, or are you okay with us proceeding with tagging this?

@cacraigucar, I do have some comments. I will turn in what I have by the end of the day even if it is partial.

src/physics/cam/aerosol_optics_cam.F90
```diff
@@ -783,24 +861,31 @@ subroutine aerosol_optics_cam_sw(list_idx, state, pbuf, nnite, idxnite, tauxar,
...
                    column: do icol = 1,ncol
-                      dopaer(icol) = pext(icol)*mass(icol,ilev)
+                      dopaer(icol) = pext(icol) * mass(icol,ilev)
                       tauxar(icol,ilev,iwav) = tauxar(icol,ilev,iwav) + dopaer(icol)
                       wa(icol,ilev,iwav) = wa(icol,ilev,iwav) + dopaer(icol)*palb(icol)
                       ga(icol,ilev,iwav) = ga(icol,ilev,iwav) + dopaer(icol)*palb(icol)*pasm(icol)
                       fa(icol,ilev,iwav) = fa(icol,ilev,iwav) + dopaer(icol)*palb(icol)*pasm(icol)*pasm(icol)
 
-                      call update_diags
+                      if (.not.aeroprops%is_bulk()) then
```
gold2718: Is this somehow generally true for bulk aerosol models or is it specific to the bulk model currently in CAM? This is another place (and below) where an `aerosol_is` interface could be both more specific and more informative.

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
```diff
@@ -0,0 +1,721 @@
...
+    integer, intent(in) :: bin_ndx  ! bin number
+
+    character(len=aero_name_len) :: name
+    character(len=64), allocatable :: names(:)
+    integer :: naer
+
+
+    call rad_cnst_get_info(list_ndx, naero=naer)
+
+    allocate( names(naer))
```
gold2718: Should this `allocate` statement be checked for failure?

src/physics/cam/aer_vis_diag_mod.F90
```diff
@@ -0,0 +1,97 @@
...
+    character(len=64), allocatable :: aernames(:)
+    logical :: history_aero_optics  ! Output aerosol optics diagnostics
+
+    ! number of bulk aerosols in climate list
+    call rad_cnst_get_info(0, naero=numaerosols)
+
+    if (numaerosols<1) return
+
+    ! get names of bulk aerosols
+    allocate(aernames(numaerosols))
```
gold2718: Should `allocate` statements be checked for failure? This question applies to this module, I think most other `allocate` statements are already checked.

src/physics/cam/aer_vis_diag_mod.F90
```diff
@@ -0,0 +1,97 @@
...
+
+    ! currently only implemented for climate calc
+    if (diag_idx > 0) return
+
+    ! compute total column aerosol optical depth
+    tmp(1:ncol) = sum(tau(1:ncol,:), 2)
+    ! use fillvalue to indicate night columns
+    do i = 1, nnite
+       tmp(idxnite(i)) = fillvalue
+    end do
```
gold2718: Should this work only be done if the history field is active?
```suggestion
    if ((iaer > 0) .and. hist_fld_active(odv_names(iaer)) .or. hist_fld_active('AEROD_v')) then
       tmp(1:ncol) = sum(tau(1:ncol,:), 2)
       ! use fillvalue to indicate night columns
       do i = 1, nnite
          tmp(idxnite(i)) = fillvalue
       end do
    end if
```

src/physics/cam/aer_rad_props.F90
```diff
@@ -19,25 +19,27 @@ module aer_rad_props
...
-! Private data
-character(len=fieldname_len), pointer :: odv_names(:)  ! outfld names for visible OD
+! debug diags
+character(len=32) :: sw_ta_name(nswbands)
+character(len=32) :: sw_wa_name(nswbands)
+character(len=32) :: sw_ga_name(nswbands)
+character(len=32) :: sw_fa_name(nswbands)
+character(len=32) :: lw_ta_name(nlwbands)
 
+logical :: bam_debug = .false.
```
gold2718, line 42: Is this logical and code that uses it below still needed?

gold2718 (changes requested):
First, it is nice to see more of the essential aerosol interface needs being met (e.g., with more of the optics defined in the objects). Thanks for all the work on this.
However, I feel this PR takes the aerosol objects in a direction that makes them more CAM specific and which shuts out other model types.
I think that for now, a lot of the problems can be addressed by having the models identify themselves instead of trying to guess by looking at particular properties. See the suggestion in aerosol_properties_mod.F90.

cacraigucar:
@gold2718:  FYI - Since you are actively contributing change requests to this PR (thanks!), I will depend on your review approval before I advance this into the tagging phase.  Once @fvitt makes the requested changes (or says why he does not want to do so), I will suggest he ask for re-review, so you are aware that he thinks he's done.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: @gold2718 I like this suggestion.  Since you offered, please submit a PR to this branch with the changes you have in mind.  Thanks.

src/chemistry/aerosol/aerosol_properties_mod.F90
fvitt: Implementing this in a separate PR sounds reasonable to me.

src/physics/cam/aerosol_optics_cam.F90
fvitt: This is specific to how BAM is currently implemented in CAM.  With some work we might be able to remove the conditional.

cacraigucar:
At a planning meeting today, this PR was listed as a roadblock for some high priority development.  We would like to bring this into CAM later this week, early next week.  I would suggest we open issue(s) for significant changes and try to move this PR forward.

src/physics/cam/aer_vis_diag_mod.F90
fvitt: Your suggested change gave me an index out of range error whe iaer is zero.  So my implementation is a little different.

src/physics/cam/aer_vis_diag_mod.F90
fvitt: done

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
fvitt: done

src/chemistry/aerosol/aerosol_properties_mod.F90
gold2718: I agree. It will be nice to see it as a clean change to see if it really helps. Shall I open an issue for this (and maybe for `optics_params`?

gold2718:
> At a planning meeting today, this PR was listed as a roadblock for some high priority development. We would like to bring this into CAM later this week, early next week. I would suggest we open issue(s) for significant changes and try to move this PR forward.

Is it okay for me to open issues for these items? Assuming my `is_bulk` to `aerosol_is` suggestion goes through, I have:
- refactor `aero_props_get` into separate methods. 
- refactor `optics_params` into separate methods.
- It appears that aerosol properties objects and especially aerosol state objects are created in multiple places and possibly more than once during a run. For some aerosol models, the state is not the same as fields in the CAM state and evolve over time. Thus, I think we need a factory to provide a single aerosol state object per aerosol type, per run so that all routines can use it.
- Some aerosol models have internal states that are different from the mixing ratios used in chemistry. Perhaps we need `set_chemistry` and `get_chemistry` methods to make this transition? Do we also need methods for radiation?

@cacraigucar, are any or all of these okay as new CAM issues?

jimmielin:
Hi @gold2718,

> * It appears that aerosol properties objects and especially aerosol state objects are created in multiple places and possibly more than once during a run. For some aerosol models, the state is not the same as fields in the CAM state and evolve over time. Thus, I think we need a factory to provide a single aerosol state object per aerosol type, per run so that all routines can use it.
>
> * Some aerosol models have internal states that are different from the mixing ratios used in chemistry. Perhaps we need `set_chemistry` and `get_chemistry` methods to make this transition? Do we also need methods for radiation?

I like this idea, but I am concerned with potentially unnecessary copying if the aerosol model does not need this distinction.

In general I think a discussion of who holds the 'state' copy of mixing ratios for chemistry and aerosol species will be helpful. If no different internal state is needed then other modules would just point to that copy, otherwise create/manage their own internal state. It would be helpful to have a small proof of concept for discussion.

cacraigucar:
@gold2718 - yes please open separate issues on all of your points.  It would be great if you can get them opened in the next day or so, so that @jimmielin has time to give some initial thoughts before he goes on PTO for a couple of weeks.

Also, I should point out that there are funding grants which require that BAM and MAM be in CAM-SIMA in the "near" future, so we locally may take some shortcuts for the short-term and retrofit when things are fully developed.

With this in mind, we would like to finalize "soon" the interfaces between chemistry and  physics.  The interfaces between aerosols and chemistry can be finalized at a later date as they are not required for our immediate development purposes (and those aerosol interfaces will be developed by the ACOM group)

@jimmielin  @fvitt @nusbaume  - Please chime in if I misspoke, as you folks are down in the trenches with this

src/chemistry/aerosol/aerosol_properties_mod.F90
gold2718: @fvitt, please let me know when you have had a time to look at [my PR](https://github.com/fvitt/CAM/pull/7), especially if it does not look okay. I'm having a bit of trouble getting running on Derecho.

gold2718 (approved):
This looks good now.
I really like how you fleshed out the `model_is` method.
I will open issues on the other changes I think would be beneficial.
Thanks for all the work on this!

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
gold2718: I am okay with these changes but leaving it for @jimmielin to resolve.

src/chemistry/aerosol/aerosol_properties_mod.F90
gold2718: This looks great now, thanks!

src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
jimmielin: Thanks @fvitt and @gold2718, resolving this now!

PR #1406, opened 2025-10-01
cam6_4_125: remove unicon; remove -ocn option; misc

brian-eaton:
* Remove the UNICON code package

* Remove `-ocn` option from configure.  It was only used by `build-namelist` to
  allow different tunings for active and non-active ocean.  It was only
  being used to set 1 cam4 tuning variable (which is no longer relevent).

* Fix longname for `FSCAMARM95`


jimmielin (approved):
Thank you @brian-eaton!

Issue #1421, opened 2025-10-22
Incorrect subtree ./lib/git-fleximod/git-fleximod


jedwards4b:
### What happened?

There are two copies of git-fleximod currently in cam.
The older version is installed correctly, the update was done incorrectly.  
I think that may be why cam is still having problems with git-fleximod.
When installed correctly you should have a directory .lib with only a git-fleximod subdirectory and
no other files.   

### What are the steps to reproduce the bug?

Look here: https://github.com/ESCOMP/CAM/tree/cam_development/.lib

### What CAM tag were you using?

cam_development cam6_4_126

### What machine were you running CAM on?

CGD machine (e.g. izumi), CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Yes

PR #1422, opened 2025-10-22
Fix git fleximod subtree

jedwards4b:
There were two copies of git-fleximod in cam_development.  The updated copy was imported incorrectly and so the older version was still in use.   This explains why cam users continue to have problems with git-fleximod.

Fixes issue #1421


nusbaume (approved):
Thanks @jedwards4b!

PR #1424, opened 2025-10-23
cam6_4_137: Incorporate final RRTMGP mods to complete CPU-ized ccppization

peverwhee:
1. Bring in latest atmospheric physics tag (To be made) that includes interface updates.
2. Add missing snapshot field
3. Modify & add new constituent stubs to allow (unused) rrtmgp schemes to build


nusbaume (commented):
This PR looks good to me @peverwhee!  

However, if we can get rid of the references to `machine` in RRTMGP in your current RRMTGP atmospheric_physics PR, then can we also get rid of the `machine.F90` file in `src/utils/cam_ccpp` as well in this PR?  Of course if not then that's fine too, I Just figured it couldn't hurt to check.  Thanks!

peverwhee:
@nusbaume file removed :santa:

nusbaume (approved):
Thanks for getting rid of the file!  Everything looks great to me now.

PR #1437, opened 2025-11-18
Made ERI F1850C test work with gregorian calendar.

kdraeder:
Motivation 
-----------
Enable testing of the gregorian calendar for all of the start modes of a multi-instance case,
which will help ensure that DART can be used with every release (maybe tag) of CESM.

Description
-------------
The ERI test cannot use both the _CG (gregorian calendar) _C# (multi-instance) modifiers
because some conditionals in cime refer only to _C and unsuitable years are requested 
from some CAM external forcing files. 
This PR makes the test work for leap days.
It should require no changes to other uses of the ERI test, 
but I have not run other tests because I don't know which tests to run.

Fixes
-----
Several more changes are needed in cime, which are being handled in CIME PR [#4899](https://github.com/ESMCI/cime/pull/4899).

This was tested using
ERI_D_C3_CG_Ld8.ne30pg3_g17.F1850C_LTso.derecho_intel.cam-outfrq9s_leapday

bld/namelist_files/use_cases/1850_cam_lt.xml

- Change year of some forcing files to 1852 so that leap day is handled.

cime_config/testdefs/testmods_dirs/cam/outfrq9s_leapday/shell_commands

- Change start year to 2008, so that adding 4 (cime) yields a date that's within the time span of all of the forcing files.


kdraeder:
I can't request reviewers, assign the issue, or select labels, so please add those as appropriate.
@billsacks and @jedwards4b are requested reviewers in the CIME issue.

bld/namelist_files/use_cases/1850_cam_lt.xml
```diff
@@ -8,15 +8,15 @@
 
 <!-- LBC Files : -->
 <flbc_file>atm/waccm/lb/LBC_17500116-20150116_CMIP6_0p5degLat_c180905.nc</flbc_file>
-<flbc_cycle_yr>1850</flbc_cycle_yr>
+<flbc_cycle_yr>1852</flbc_cycle_yr>
```
jedwards4b: I'm not sure that it will be acceptable to change the start year for this usecase, could you create a new usecase instead? 1852_cam_lt.xml perhaps

cime_config/testdefs/testmods_dirs/cam/outfrq9s_leapday/shell_commands
```diff
@@ -3,4 +3,4 @@ if [ "$driver" = "nuopc" ]; then
   ./xmlchange ROF_NCPL=\$ATM_NCPL
   ./xmlchange GLC_NCPL=\$ATM_NCPL
 fi
-./xmlchange RUN_STARTDATE=2012-02-29
+./xmlchange RUN_STARTDATE=2008-02-29
```
jedwards4b, line 6: Why this change?  Aren't these both leapyears?

kdraeder: Some of the external forcing files end in 2014.  Since I'm changing the increment between the forecasts from 2 to 4 years I wanted to avoid problems from requesting a year that's beyond
the time span of those files.  If that was too pre-emptive, I'll try a test with 2012.

bld/namelist_files/use_cases/1850_cam_lt.xml
kdraeder: I've moved the new 1850_cam_lt.xml to leapyears_cam_lt.xml and reverted 1850_cam_lt.xml.
My push of these changes to my fork is failing with "The requested URL returned error: 500".
Should I push it somewhere else instead?

jedwards4b (approved):

kdraeder:
@jedwards4b 
I was pushing your requested changes to the use_case files as you were approving the pull request,
so I'm not sure that you saw them before approval.

jedwards4b:
You can see on the conversation tab of the pull request that your commit was made before my review was made.

bld/namelist_files/use_cases/leapyears_cam_lt.xml
```diff
@@ -0,0 +1,101 @@
+<?xml version="1.0"?>
+<namelist_defaults>
+
```
cacraigucar: I don't see any hooks to this use_case.  That probably should be added with this PR or the accompanying cime PR https://github.com/ESMCI/cime/pull/4899

cacraigucar (changes requested):

bld/namelist_files/use_cases/leapyears_cam_lt.xml
kdraeder: > I don't see any hooks to this use_case. That probably should be added with this PR or the accompanying cime PR [ESMCI/cime#4899](https://github.com/ESMCI/cime/pull/4899)

Yes, I ended up going in a different direction, which doesn't require this use_case.
I've removed it from my branch, but not committed that yet.

kdraeder:
A git(hub) question; I'm trying to update my cam branch (gregorian_ERI) before committing and pushing changes.
My fork in github.com/kdraeder has my branch, but I can't see how sync it with cam_development.
Only main seems to be available for syncing, and that doesn't seem right.
I tried pulling cam_development to my local clone, but that failed due to diverging branches.
Doing a merge --no-ff also failed.  The last alternative seems to be to rebase, but I have the impression
that that's a somewhat drastic thing to do.  Any recommendations?

cacraigucar:
In your local clone, I would do the following:
git remote add upstream https://github.com/ESCOMP/CAM
git fetch upstream
git merge upstream/cam_development

You might need to do a fresh clone of your branch to make sure that remnants from your previous attempts are no longer there.

Do these commands work for you, or is this what you already did?

jedwards4b (approved):

ekluzek:
Just noting that the cime tag for this has come in as cime6.1.146. So an update to that cime tag should come in as part of this. And the two tags should be coordinated in CESM and CAM.

cacraigucar:
I assume we should add a regression test for something like: ERI_D_C3_CG_Ld8.ne30pg3_g17.F1850C_LTso.derecho_intel.cam-outfrq9s_leapday?  I don't see that being added in this CAM tag yet.  We can add that when we bring in this tag.  Aside from that, is this PR ready to go?

kdraeder:
I believe that a test like that needs to wait for the resolution of CAM issue #1447,
which raises the question of whether spinup compsets (1850) _should be made to_ work with Gregorian.
When that test is implemented it will probably use the new gregorian modifier _cG instead of _CG (https://github.com/ESMCI/cime/issues/4811).

The context is that the final changes in this PR have been tested with FHIST, but not F1850,
and re not expected to work with 1850.  So the title ended up being misleading.  Can it be changed without adding to the confusion?

kdraeder:
@cacraigucar Thanks for shepherding this through the current cam-cime inconsistency!

There are 5 commits, but some of the earlier commits cancel others, because we had not settled on the preferred strategy to handle this situation.  If there's a cleaner way to merge just the last commit ([1f63e57](https://github.com/ESCOMP/CAM/pull/1437/commits/1f63e574fc01ceb040e493e73d53961a64c72f40)),
then that seems like a better choice.

cacraigucar:
Replaced by #1473

PR #1457, opened 2025-12-15
cam6_4_143: Extended non-LTE CO2 cooling parameterization

### LINKED ISSUE #1358: Extended CO2 non-LTE cooling

Implement the option to use an extended parameterization of CO2 cooling described here:
https://doi.org/10.5194/egusphere-2023-2424, by Manuel Lopez-Puertas, et al.

Some test results are posted here:
https://github.com/NCAR/wawg_dev/issues/95

_No response_

Yes

Yes

fvitt:
Closes #1358


src/physics/waccm/nlte_extco2.F90
```diff
@@ -0,0 +1,141 @@
...
+  subroutine nlte_extco2_init(co2_mw,n2_mw,o1_mw,o2_mw)
+    use ref_pres, only: pref_mid
+
+    ! Input variables
+    real(r8), intent(in) :: o1_mw     ! O molecular weight
+    real(r8), intent(in) :: o2_mw     ! O2 molecular weight
+    real(r8), intent(in) :: co2_mw    ! CO2 molecular weight
+    real(r8), intent(in) :: n2_mw     ! N2 molecular weight
+
+#ifdef EXT_CO2_COOL
```
cacraigucar: I'm wondering about why you felt the need to encapsulate this code within a #ifdef.  We try to avoid using these directives and backtracing this particular routine, it is only called when `nlte_use_extco2` is true.  I believe this #ifdef is not needed.   I've not analyzed the other #ifdef directives, but I would suggest there might also might be others that are not needed.  If all can be removed, please remove the definition from configure as well.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -182,7 +182,7 @@
 <ncdata hgrid="ne16np4"  nlev="126" waccmx="1"                                 >atm/waccm/ic/waccmx_ne16np4_126L_c200108.nc</ncdata>
 <ncdata hgrid="ne16np4"  nlev="126" waccmx="1" aquaplanet="1"                  >atm/waccm/ic/waccmx_aqua_ne16np4_126L_c191108.nc</ncdata>
 <ncdata hgrid="ne16np4"  nlev="126" waccmx="1" aquaplanet="1" ionosphere="none">atm/waccm/ic/waccmx4_neutral_aquap_ne16np4_126lev_c200827.nc</ncdata>
-<ncdata hgrid="ne16np4"  nlev="130" waccmx="1"     >atm/waccm/ic/waccmx_ne16pg3_L130_c250318.nc</ncdata>
+<ncdata hgrid="ne16np4"  nlev="130" waccmx="1"     >atm/waccm/ic/FX2000_ne16pg3_spinup01.cam.i.0002-01-01_c251215.nc</ncdata>
```
cacraigucar, line 185: Has this file been "rimport"ed?

bld/namelist_files/namelist_definition.xml
```diff
@@ -7994,6 +7994,13 @@ Call ALI-ARMS every X timesteps
 Default: 1
 </entry>
 
+<entry id="nlte_use_extco2" type="logical" category="waccm_phys"
+       group="radheat_nl" valid_values="" >
+If TRUE, then use the extended CO2 scheme of Lopez-Puertas et al. 2024 as an alernative method of computing non-LTE
```
cacraigucar, line 7999: Since this is a comment which will show up on a web page, would it be appropriate to include the DOI here also?

cacraigucar (changes requested):

src/physics/waccm/nlte_extco2.F90
fvitt: The external code ext_co2_cooling code does not compile with NAG compiler.

cacraigucar: With that in mind, does it make sense to only #ifdef the calls and the use statements (keep it contained around the code which only needed to not compiled)?

fvitt: I would prefer not to add the history fields when the model is not configured to use the extended cooling method.

bld/namelist_files/namelist_defaults_cam.xml
fvitt: done

bld/namelist_files/namelist_definition.xml
fvitt: DOI has been added here

src/physics/waccm/nlte_extco2.F90
cacraigucar: Is there something preventing you from passing in the variable `nlte_use_extco2` and using it to determine if the history fields are being set or not?

fvitt: I have concluded with the checks in higher level modules, the cpp ifdefs in init routine was not necessary.

cacraigucar (approved):

Issue #1414, opened 2025-10-13
Frontogenesis function should be computed on pressure levels


PeterHjortLauritzen:
### What happened?

Walter Hannah (DOE) reported a bug in the computation of the frontogenesis function (computed in dp_coupling; dycore dependent) that was brough to our attention by @chihchen24. Code modification for cam_development provided by Walter Hannah and Jack Chen.

This needs to be fixed for all dynamical cores computing the frontogensis function. @kuanchihwang  @JulioTBacmeister @mgduda @mbramberger @mijeong135

@PeterHjortLauritzen will do PR for SE (since we need it ASAP for CESM3)

Also: we need to make same changes in CAM-SIMA for SE and MPAS.

### What are the steps to reproduce the bug?

Run any version of CAM7 or WACCM!

### What CAM tag were you using?

any tag

### What machine were you running CAM on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Yes, but I will need some help

cacraigucar:
@kuanchihwang - has this been fixed for MPAS?  If so, then this issue can be closed

PR #1467, opened 2026-01-05
cam6_4_150: fix FV frontogenesis bug and diagnostic PS/PSDRY bug in SE log file

PeterHjortLauritzen:
Fix bug in frontogenesis function for FV dycore: partially closes https://github.com/ESCOMP/CAM/issues/1414
(science tested by ACOM; there will be namelist changes to re-tune QBO with the fixed frontogenesis function)
@fvitt @mbramberger @tilmes @chihchen24

Minor bug fix for diagnoatics written to log file with spectral-element dynamical core (PS and PSDRY are missing the pressure at model top).


src/dynamics/fv/gravity_waves_sources.F90
```diff
@@ -108,15 +119,30 @@ subroutine gws_src_fnct(grid, u3,v3,pt, q3, pe, frontgf, frontga)
                ! Pot. Temperature
                pty(i,k,j) = ( ptg(i,k,j+1) - ptg(i,k,j-1) ) / (2._r8 * grid%dp)
                pty(i,k,j) = pty(i,k,j) / aearth
+               !
+               ! Topography correction term:
+               ! Horizontal temperature gradient for fronotogenesis function was originally computed over
```
cacraigucar: Is it currently being computed over terrain-following coordinates and then correction terms applied?  If so, I would change this from `was` to `is`.  I believe this is the case.  If on the other hand, this is talking about how things were computed prior to this PR, we typically do not have comments like that in the code.

cacraigucar (changes requested):

src/dynamics/fv/gravity_waves_sources.F90
chihchen24: 120-121 is terrain following.
127-128 is the correction terms.

cacraigucar: With that info, my request is to change `was` to `is` in this sentence

cacraigucar: @PeterHjortLauritzen - if you make this minor comment update, I believe this PR is ready to go, correct?

PeterHjortLauritzen: Comment beautified!

cacraigucar (approved):

PR #1470, opened 2026-01-07
cam6_4_153: Short-lived halogen chemistry

### LINKED ISSUE #1469: Short-lived halogen chemistry

Bring in short-lived halogen chemistry developed by Rafael Pedro Fernandez et al.

@RafaPedroFernandez
@CAlbertoCuevas

Yes

Yes

fvitt:
Closes #1469


fvitt:
From @RafaPedroFernandez  (CC'd @CAlbertoCuevas):

Please note that I have just made a push for 2 F90 routines in my following GIT:
https://github.com/RafaPedroFernandez/CAM/tree/slh_chemistry_n04_cam6_4_097_rpf

Which is also updated local in derecho:
/glade/u/home/rpfernan/cesm_versions_rpf/derecho_cesm/git_fetch/cesm3_cam6dev_francis/slh_chemistry_n04_cam6_4_097_rpf

Routines are
chemistry.F90: 
For SLH, we must call iodine_emissions_srf before the loop to outfld( sflxnam(m), cam_in%cflx(:ncol,m), ncol,lchnk ) in order for SFHOI and SFI2 to be properly written in the output file. If we do not so, the SFHOI and SFI2 are zero in the output file but are somehow included in cam_in%cflx array that is passed to other routines

mo_gas_phase_chmdr.F90
For Iodine washout ... we have always used SAD_ICETROP, which is computed based on CLDICE. This CLDICE was before computed as only ice
             cldice(:ncol,k)  = q(:ncol,k,cldice_ndx)
and now considers also SNOW
             cldice(:ncol,k)  = q(:ncol,k,cldice_ndx) + q(:ncol,k,snow_ndx)

Including snow in the variable makes the iodine deposition too efficient, and then we created a new array (ONLYICE) just to compute iodine wet-depositions without affecting other model developments.

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -2037,8 +2061,8 @@
 
 <!-- photolysis inputs -->
 <xs_coef_file>atm/waccm/phot/effxstex.txt</xs_coef_file>
-<xs_short_file>atm/waccm/phot/xs_short_c221005mm.nc</xs_short_file>
-<xs_long_file >atm/waccm/phot/temp_prs_GT200nm_c221005mm.nc</xs_long_file>
+<xs_short_file>atm/waccm/phot/xs_short_c221005mm_slh_c250930.nc</xs_short_file>
+<xs_long_file >atm/waccm/phot/temp_prs_GT200nm_c221005mm_slh_c250930.nc</xs_long_file>
```
cacraigucar, line 2068: These two file still need to be "rimport"ed

bld/namelist_files/namelist_definition.xml
```diff
@@ -6294,6 +6294,53 @@ List of gas-phase species that undergo wet deposition via the wet deposition sch
...
+       group="slh_nl" valid_values="" >
+Scaling-factor for SLH Halocarbon surface emissions
+Default: set by build-namelist
+</entry>
+
+<entry id="ICEfraprx_ScalingFactor_Br" type="real" category="cam_chem"
+       group="slh_nl" valid_values="" >
+Scaling-factor for SLH Halocarbon surface emissions
+Default: set by build-namelist
+</entry>
```
cacraigucar, line 6373: Comments probably need updating (since they are all the same)

cime_config/config_compsets.xml
```diff
@@ -685,7 +710,8 @@
 	<value  compset="HIST_CAM60%CVBSX">1995-01-01</value>
 	<value  compset="HIST_CAM60%CFIRE">1995-01-01</value>
 	<value  compset="RCP[2468]_CAM\d+">2005-01-01</value>
- 	<value  compset="_CAM.*%NUDG"                   >2010-01-01</value>
+ 	<value  compset="_CAM.*%NUDG"                   >2000-01-01</value>
```
cacraigucar: Why is this PR changing the date for nudging runs?

cacraigucar (changes requested):
Partial review - @jimmielin will be reviewing the chemistry section of this PR.  My review was on the setup side.  

I noticed that there are testing use_cases setup, but I do not see any regression tests being added.  Is this an oversight?

src/chemistry/mozart/clybryiy_fam.F90
```diff
@@ -0,0 +1,461 @@
...
+    enddo
+    call get_short_lived_species( mmr, lchnk, ncol, pbuf )
+
+    !--------------------------------------------------------------
+    !       ... form updated chlorine atom mass mixing ratio
+    !--------------------------------------------------------------
+    if ( id_cly>0 ) then
+       wrk(:,:) = cloy( mmr, pcols, ncol )
+
+       factor(:ncol,:) = mmr(:ncol,:,id_cly) / wrk(:ncol,:)
```
jimmielin: Maybe a safeguard against divide by zero here if `wrk(:ncol,:)` are close to zero (when `cloy` initial conditions are very low?)

src/chemistry/mozart/mo_slh_routines.F90
```diff
@@ -0,0 +1,300 @@
...
+       end if
+       close(unitn)
+       call freeunit(unitn)
+    end if
+
+!rpf_CESM2_SLH
+    call mpibcast (SSAdehal_ScalingFactor,      1, mpir8,    0, mpicom)
+!     call mpibcast (SLHemiss_ScalingFactor,      1, mpir8,    0, mpicom)
+!     call mpibcast (ICEfraprx_ScalingFactor,     1, mpir8,    0, mpicom)
+!     call mpibcast (LIQfraprx_ScalingFactor,     1, mpir8,    0, mpicom)
```
jimmielin: Do these also need to be broadcast?

src/chemistry/mozart/mo_slh_routines.F90
```diff
@@ -0,0 +1,300 @@
...
+    o3_surf (:ncol) = state%q(:ncol,pver,o3_cnst) * conv_fact_o3
+    ws_mask(:ncol) = sqrt( state%u(:ncol,pver)*state%u(:ncol,pver) &
+                         + state%v(:ncol,pver)*state%v(:ncol,pver) )
+
+!rpf_CESM2_SLH
+!rpf and cac: This threshold is required to aviod huge iodine emissions for low wind peed
+    where ( ws_mask(:ncol) < 3._r8 )
+      ws_mask(:ncol) = 3._r8
+    endwhere
+!rpf_CESM2_SLH
```
jimmielin: Maybe clean up this comment a bit and fix typos:
```suggestion
!rpf and cac: This threshold is required to avoid huge iodine emissions for low wind speeds
    where ( ws_mask(:ncol) < 3._r8 )
      ws_mask(:ncol) = 3._r8
    endwhere
```

src/chemistry/mozart/mo_slh_routines.F90
```diff
@@ -0,0 +1,300 @@
...
+
+!rpf_CESM2_SLH
+    do i = 1,ncol
+       !!! Be CAREFULL that if SST = 0 then the division is not possible. Should it be forced to zero?
+       if (ocnfrac(i) == 0) then ! Actually it should be landfrac = 1 ... but we do not want to bring landfrac as an additional agrument
+          iodide_aq(i) = 0 
+       else
+!         iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) ) * ocnfrac(i)
+          iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) )
+       end if
```
jimmielin: Since SST is in Kelvin it shouldn't get close to zero but I would check if it is intended that `iodine_aq` is computed when `sst` is below freezing? This may be a science question

src/chemistry/mozart/mo_slh_routines.F90
```diff
@@ -0,0 +1,300 @@
...
+    cam_in%cflx(:ncol,hoi_cnst) = cam_in%cflx(:ncol,hoi_cnst) + flx_hoi(:ncol) * conv_fact_hoi
+
+  end subroutine iodine_emissions_srf
+  
+
+  subroutine iodine_emissions_set( ncol, ocnfrac, o3_surf, sst, ws_mask,  &
+                                  iodide_aq, flx_i2, flx_hoi )
+!rpf_CESM2_SLH
+    implicit none
+!rpf_CESM2_SLH
```
jimmielin: Can be removed since module is declared `implicit none`

src/chemistry/mozart/mo_srf_emissions.F90
```diff
@@ -325,6 +361,11 @@ subroutine set_srf_emissions( lchnk, ncol, sflx )
     use string_utils, only : to_lower, GLC
     use phys_grid,    only : get_rlat_all_p, get_rlon_all_p
 
+!rpf_CESM2_SLH
+!     SLHemiss_ScalingFactor = 1.0 always. If needed emissions should be modified directly in srf_emis_specifier file (rpf)
+!     use mo_slh_routines,  only : SLHemiss_ScalingFactor
+!rpf_CESM2_SLH
+
```
jimmielin: Is this still needed? It seems like `SLHemiss_ScalingFactor` is overall unused so maybe all references of it can be removed throughout (here and in `mo_slh_routines.F90`)

src/chemistry/mozart/mo_strato_rates.F90
```diff
@@ -117,13 +128,49 @@ subroutine init_strato_rates
...
           id_clono2 = get_spc_ndx( 'CLONO2' )
           id_hcl    = get_spc_ndx( 'HCL' )
           id_hocl   = get_spc_ndx( 'HOCL' )
           id_hobr   = get_spc_ndx( 'HOBR' )
           id_n2o5   = get_spc_ndx( 'N2O5' )
+!rpf_CESM2_SLH
 
+!rpf_CESM2_SLH
+! I do not include the new rid_# into ids(:) so has_strato_chem condition is not afected
+!rpf_CESM2_SLH
```
jimmielin: The meaning of "new rid_#" will become unclear after the code is merged in, I would suggest something like

```suggestion
! Short-lived halogen chemistry rid_# (rid_het5_hbr to rid_nat_str_i_4) are not included into ids(:)
! so has_strato_chem condition is not affected.
```

src/chemistry/mozart/mo_strato_rates.F90
```diff
@@ -714,15 +876,90 @@ subroutine ratecon_sfstrat( ncol, ad, pmid, temp, rad_sulfate, sad_sulfate, &
...
+                        end if
+                     end if
+!rpf_CESM2_SLH                  
+
+!rpf_CESM2_SLH
+                     if ( rid_slf_str_i_2 > 0 ) then
+                        if( hclvmr > small_div .and. hoivmr > small_div ) then
+! CAC 2020 RPF START OF implmentation of the HOI + HCl(liq) =  ICl + H2O  Sulfate Aerosol Reaction
+! NOTE: gprob for HOI species is mapped to gprob_hobr_hcl 
+                           if ( rid_slf_str_i_2 > 0 ) then
```
jimmielin: The check in line 918 is a duplicate of the check in line 914 for `rid_slf_str_i_2 > 0`.

src/chemistry/mozart/mo_strato_rates.F90
```diff
@@ -714,15 +876,90 @@ subroutine ratecon_sfstrat( ncol, ad, pmid, temp, rad_sulfate, sad_sulfate, &
...
+                              end if           
+                           end if
+                        end if
+                     end if
+
+                     if ( rid_slf_str_i_3 > 0 ) then
+                        if( hbrvmr > small_div .and. hoivmr > small_div ) then
+! CAC 2020 RPF START OF implmentation of the HOI + HBr(liq) =  IBr + H2O  Sulfate Aerosol Reaction
+! NOTE: gprob for HOI species is mapped to gprob_hobr_hcl 
+                           if ( rid_slf_str_i_3 > 0 ) then
```
jimmielin: Same duplicate checks here

src/chemistry/mozart/mo_strato_rates.F90
```diff
@@ -714,15 +876,90 @@ subroutine ratecon_sfstrat( ncol, ad, pmid, temp, rad_sulfate, sad_sulfate, &
...
+                              end if
+                           end if
+                        end if
+                     end if
+
+                     if ( rid_slf_str_i_4 > 0 ) then
+                        if( hivmr > small_div .and. hoivmr > small_div ) then
+! CAC 2020 RPF START OF implmentation of the HOI + HI (liq) =  I2  + H2O  Sulfate Aerosol Reaction
+! NOTE: gprob for HOI species is mapped to gprob_hobr_hcl 
+                           if ( rid_slf_str_i_4 > 0 ) then
```
jimmielin: Same duplicate checks here

src/chemistry/mozart/mo_usrrxt.F90
```diff
@@ -891,7 +1277,11 @@ subroutine usrrxt( state, rxt, temp, tempi, tempe, invariants, h2ovmr,  &
        if( usr_NO2XNO3_M_ndx > 0 ) then
           if( tag_NO2_NO3_ndx > 0 ) then
              call comp_exp( exp_fac, -10840.0_r8*tinv, ncol )
-             rxt(:,k,usr_NO2XNO3_M_ndx) = rxt(:,k,tag_NO2_NO3_ndx) * 1.734138e26_r8 * exp_fac(:)
+!rpf_CESM2_SLH
+! Is there a typo in the pre-exponential factor? (1.72 instead of 1.73?) Which one is correct?
+!            rxt(:,k,usr_NO2XNO3_M_ndx) = rxt(:,k,tag_NO2_NO3_ndx) * 1.734138e26_r8 * exp_fac(:)
+             rxt(:,k,usr_NO2XNO3_M_ndx) = rxt(:,k,tag_NO2_NO3_ndx) * 1.724138e26_r8 * exp_fac(:)
+!rpf_CESM2_SLH
```
jimmielin: This comment is leftover and should be checked for correctness

jimmielin (changes requested):
Thanks @fvitt, I have a few suggested changes but nothing major.

src/chemistry/mozart/mo_usrrxt.F90
fvitt: @RafaPedroFernandez Can you comment on this?  Do you know what the correct value is?

bld/namelist_files/namelist_defaults_cam.xml
RafaPedroFernandez: @fvitt 
The files are located at:
/glade/campaign/cesm/cesmdata/cseg/inputdata/atm/waccm/phot
I understand that both files should be moved to
/glade/campaign/collections/gdex/data/d651077/cesmdata/inputdata/atm/waccm/phot
and then include a symbolic link (ln -s) in the original path.

Francis, I assume you are the responsible for doing so

bld/namelist_files/namelist_definition.xml
RafaPedroFernandez: @fvitt 
Please update the "long-name" description as follows (highlighted in bold):
(Note that there are 6 Scaling Factors ... I added a complete new block for SSAdehal_ScalingFactor that should be just a few lines above in the same file)

<entry id="**SSAdehal_ScalingFactor**" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**SSAdehal_ScalingFactor**
**Scaling-factor for SLH dehalogenation reaction (XONO2,XNO2,HOX->X2,XY ... X,Y=Br,Cl,I) on Sea-Salt Aerosols**
Default: set by build-namelist
</entry>

<entry id="SSAhno3_ScalingFactor" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**Scaling-factor for SLH acid displacement reaction (HNO3->HCl) on Sea-Salt Aerosols**
Default: set by build-namelist
</entry>

<entry id="SSAn2o5_ScalingFactor" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**Scaling-factor for SLH dinitrogen pentoxide reaction (N2O5->HNO3+CLNO2) on Sea-Salt Aerosols**
Default: set by build-namelist
</entry>

<entry id="LIQfraprx_ScalingFactor_I" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**Scaling-factor for SLH washout (Iodine) on liquid droplets based on Free-Regime Approx.**
Default: set by build-namelist
</entry>

<entry id="ICEfraprx_ScalingFactor_I" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**Scaling-factor for SLH washout (Iodine) on ice crystals based on Free-Regime Approx.**
Default: set by build-namelist
</entry>

<entry id="ICEfraprx_ScalingFactor_Br" type="real" category="cam_chem"
       group="slh_nl" valid_values="" >
**Scaling-factor for SLH washout (Bromine) on ice crystals based on Free-Regime Approx.**
Default: set by build-namelist
</entry>

cime_config/config_compsets.xml
RafaPedroFernandez: When I initially ported the code I just copy/paste the most similar PR ... and I shifted the date to 2000 because in the description paper we only validated SLH simulations in the nudgged setup for the period 2000-2005.

I suggest keeping the default value use for year 2010 and in any case the user can modify this using user_nl_cam.

src/chemistry/mozart/clybryiy_fam.F90
RafaPedroFernandez: I do not know what to say here about dividing by zero. Francis, do you have any suggestion? Note I just copy/paste and updated those lines as they were initially implemented in the old clybry.F90 file when Iodine was not considered. And we never found an issue with this.

In my oppinion, the most important thing to control here are lines 101-116: In case the chem_mech.in DOES NOT include IODINE, then the "original" ids_clybry species should be considered ... so any user working with a mechanism that does not include SLH can compute the total Cl and Br abundance properly.

!rpf_CESM2_SLH
    ids_clybry   = (/ id_cly,id_bry, &
             id_cl,id_clo,id_hocl,id_cl2,id_cl2o2,id_oclo,id_hcl,id_clono2, &
             id_br,id_bro,id_hbr,id_brono2,id_brcl,id_hobr /)

    ids_clybryiy = (/ id_cly,id_bry,id_iy, &
             id_cl,id_clo,id_hocl,id_cl2,id_cl2o2,id_oclo,id_hcl,id_clono2,id_clno2, &
             id_br,id_bro,id_hbr,id_brono2,id_brcl,id_hobr,id_br2,id_brno2, &
             id_i,id_i2,id_io,id_oio,id_hi,id_hoi,id_ino,id_ino2,id_iono2,id_ibr,id_icl,id_i2o2,id_i2o3,id_i2o4 /)

    if ( id_iy>0 ) then
      has_clybryiy = all( ids_clybryiy(:) > 0 )
    else
      has_clybryiy = all( ids_clybry(:) > 0 )
    endif
!rpf_CESM2_SLH

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: Please remove those commented lines. That was the original name of the ScalingFactors and we updated them. There should only be 6 scaling factors for SLH.

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: Thanks for finding the typo. I agree with the change. Please replace "huge" by "large"

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: You are right ... SST is in Kelvin and therefore no division by zero will occur here. 
The "scientific" condigion about considereing or not iodide_aq fields for iodine emissions is considered above in the logical condition "if (ocnfrac(i) == 0) then"

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: You are right. I found "implicit none" declarations in lines 28, 73, 194 and 247. I understand only the first one is needed.

src/chemistry/mozart/mo_srf_emissions.F90
RafaPedroFernandez: Yes, please remove. That was applied for some initial developments and later discarded.

src/chemistry/mozart/mo_strato_rates.F90
RafaPedroFernandez: Perfect. Thanks !!!

src/chemistry/mozart/mo_strato_rates.F90
RafaPedroFernandez: You are correct. This is a duplicate condition and the second one could be removed.
Note the same happens for rid_slf_str_i_3 in line 932 and also for rid_slf_str_i_4 in line 948.

Idem for rid_nat_str_i_2, rid_nat_str_i_3, rid_nat_str_i_4 in lines 1131, 1147, 1163.

Idem for rid_ice_str_i_2, rid_ice_str_i_3, rid_ice_str_i_4 in lines 1337, 1353, 1369.

src/chemistry/mozart/mo_strato_rates.F90
RafaPedroFernandez: correct

src/chemistry/mozart/mo_strato_rates.F90
RafaPedroFernandez: correct

src/chemistry/mozart/mo_usrrxt.F90
RafaPedroFernandez: @fvitt @lkemmons @tilmes 
Francis. I have not developed not used those reactions for our SLH scheme. I just noticed that the 1.734138e26_r8 value used in original line 1282 was different to the equivalent expresions used in lines 1264 ad 1272 for equivalent reactions (that used the value 1.724138e26_r8). As numbers are very similar and expresions are equivalent I thought this could be an unnoticed typo. I believe Louisa or Simone can tell you which is the correct value (or who should we ask for).

src/chemistry/mozart/clybryiy_fam.F90
jimmielin: Thanks @RafaPedroFernandez, looks like @fvitt added a check
```fortran
    if ( id_cly>0 ) then
       wrk(:,:) = cloy( mmr, pcols, ncol )

       where( wrk(:ncol,:)>0._r8 )
          factor(:ncol,:) = mmr(:ncol,:,id_cly) / wrk(:ncol,:)
       elsewhere
          factor(:ncol,:) = 0._r8
       end where
```
I am happy with this after the added checks

bld/namelist_files/namelist_definition.xml
fvitt: @RafaPedroFernandez 
See my updates to these descriptions here: https://github.com/fvitt/CAM/commit/b060e327686f20a42dc37839632b60337a227008

src/chemistry/mozart/mo_usrrxt.F90
lkemmons: @fvitt Please correct this to 1.724.
Thank you, Rafa, for finding this.

bld/namelist_files/namelist_defaults_cam.xml
fvitt: @cacraigucar We have now other new inputs.  Is using the `rimport` script sufficient for adding new inputs to the input data repo?

cime_config/config_compsets.xml
fvitt: Reverted this back

src/chemistry/mozart/mo_slh_routines.F90
fvitt: @RafaPedroFernandez  Is this block of code okay as it is?  Should the `"Be CAREFULL ..."` be removed?

src/chemistry/mozart/mo_strato_rates.F90
fvitt: Fixed

src/chemistry/mozart/mo_strato_rates.F90
fvitt: Fixed

src/chemistry/mozart/mo_strato_rates.F90
fvitt: Fixed

src/chemistry/mozart/mo_usrrxt.F90
fvitt: Thanks for this. Cleaned up the comments here and accepted the correction.

bld/namelist_files/namelist_defaults_cam.xml
cacraigucar: Yes, the script performs the copy and link steps.  As it has always been, once a file is moved there, it exists "forever", so files should only be `rimport`ed once they are "cast-in-stone".

src/chemistry/mozart/clybryiy_fam.F90
RafaPedroFernandez: @fvitt 
Francis ... I believe that in the extrange case thet wrk = 0.0 (we never found that issue) ... then factor should be equal to 1, and not to zero. Don't you think?
       elsewhere
          factor(:ncol,:) = 1._r8
       end where
I always understood this portion of the code as a "double checking" that the sum of the advected individual Cly species was equivalent to the CLOY variable.

In either case, please note that in the same routine, there are also equivalent "factor" computations for id_bry (line 257 of the original file) and id_iy (line 278). All three elsewhere conditions should be implemented for Chlorine, Bromine and Iodine.

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: @fvitt 
Yes francis. You can remove the "Be carefull ..." line completely.

In addition, please also notice that we keep the following line 272 commented as this was the original computation in CESM1.
!         iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) ) * ocnfrac(i)

But while porting the code to CESM2, we recognized that the multiplication by ocnfrac(i) was incorrect and replace it by line 273
          iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) )

You know better than me if both (commented and uncommented) lines should remain in the trunk for "trazability" of the implementation. In case it is better to keep only one, please remove line 272 (commented) and keep only line 273

bld/namelist_files/namelist_defaults_cam.xml
fvitt: done

src/chemistry/mozart/clybryiy_fam.F90
fvitt: Changed to elsewhere factor to 1.

src/chemistry/mozart/mo_slh_routines.F90
fvitt: The `mo_slh_routines` module has been cleaned up.

src/chemistry/mozart/mo_slh_routines.F90
jimmielin: Thanks @RafaPedroFernandez and @fvitt - based on Rafa's comment on the line with `ocnfrac` being the original CESM1 computation, I would suggest adding a comment to complete the traceability purpose, otherwise it is not clear why that line was commented out years later after we look at it again. Something like

```fortran
! this was the original computation in CESM1 but the multiplication
! by ocnfrac(i) is incorrect:
!         iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) ) * ocnfrac(i)
          iodide_aq(i) = 1.46e6_r8 * exp( -9134._r8 / sst(i) )
```

or if more detail could be added. Otherwise I am also happy if the commented out line is removed. Either way is fine. Thank you!

src/chemistry/mozart/mo_usrrxt.F90
jimmielin: Thanks all!

jimmielin (approved):
Thanks @fvitt and @RafaPedroFernandez for resolving my code review comments. I am happy to approve this now.

I have one remaining suggestion, in regards to the commented-out line of code, that perhaps could be documented better with an extra comment, if you would like to keep that line. But I do not need to review the PR again as it will just be a comment update.

Congratulations and thank you for bringing in SLH chemistry for CAM-chem, this is a great accomplishment.

cacraigucar (approved):

src/chemistry/mozart/mo_slh_routines.F90
RafaPedroFernandez: @jimmielin @fvitt 
Thanks Lin for including that comment for trazability. It is perfect !!!

fvitt: updated as suggested

jimmielin: Thanks all! Resolving this conversation

Issue #986, opened 2024-03-01
Update some older math routines


cacraigucar:
### Issue Type

Code Clean-up

### Issue Description

A number of older math routines need updated code

A spreadsheet with the targeted codes will be linked in here when it is ready

@do

### Will this change answers?

I Don't Know

### Will you be implementing this yourself?

Any CAM SE can do this

cacraigucar:
@briandobbins - This may introduce roundoff level changes.  Should it be done soon?

briandobbins:
Yes, but it's still secondary to some of the science changes & tunings -- I gave Dave a heads-up on this a while back (eg, that we'd need time after the tuning is set), and will remind him again.  Roundoff-level changes _shouldn't_ impact the tuning.

PR #1485, opened 2026-02-19
part of cam6_4_154: remove unused routines

PeterHjortLauritzen:
partial implementation of #986


cacraigucar (approved):
I have confirmed that these three routines appear to not be used in CAM

PR #1492, opened 2026-03-03
part of cam6_4_164: Point to correct locations for datm, satm and xatm

billsacks:
Without this change, you couldn't run compsets with some non-CAM atm from a standalone CAM checkout.

Note that the location for xatm is tied to
https://github.com/ESCOMP/CMEPS/pull/637 and
https://github.com/ESMCI/cime/pull/4946 (but this PR can be merged before or after those because it's unusual to run X compset tests from a standalone CAM checkout - and this is something that I think was broken up until now anyway).

### DIFF
--- a/.config_files.xml
+++ b/.config_files.xml
@@ -5,7 +5,7 @@
 <entry_id>
 
   <!-- This is the same as the default entry in
-       cime/config/cesm/config_files.xml except for the value for clm:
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Thanks @billsacks!

cacraigucar (approved):

PR #1506, opened 2026-03-13
cam6_4_163: Refactor rad_constituents, aerosol_optics_cam for portability; extend abstract aerosol interface

### LINKED ISSUE #1504: Refactor rad_constituents and abstract aerosol interface to share with CAM-SIMA
### Issue Type

Infrastructure Update

### Issue Description

Clean up / split rad_constituents into gases-aerosols and make portable to the same modules can be used in CAM-SIMA with reasonable changes (ideally most modules as-is)

Will be a bit-for-bit update

### Will this change answers?

No

Yes

### LINKED ISSUE #1505: Refactor aerosol_optics_cam to extract core computations for sharing with CAM-SIMA
### Issue Type

Infrastructure Update

### Issue Description

`aerosol_optics_cam.F90` contains monolithic SW/LW aerosol optics routines that mix portable computation with CAM-specific diagnostics, aerosol abstract interface calls, and state access. 

The goal is to extract the portable optics core into `aerosol_optics_core.F90` so it can be shared with CAM-SIMA, leaving the CAM-specific diags code etc. in place.

This will be a bit-for-bit change.

### Will this change answers?

No

Yes

jimmielin:
Motivation for changes:
- Make portable the rad_constituents infrastructure for porting to CAM-SIMA. Previously, the monolithic module combined radiatively active gases and aerosols (which are split in SIMA) and handled namelist read, pbuf and constituent indexing and access (which will be different in SIMA).
- Expand use of aerosol abstract interface to allow for physics code to swap underlying aerosol representations without changes to how they interface with aerosol; most modules now use the abstract types only.
- Make portable the aerosol_optics, aerosol_optics_cam infrastructure for porting to CAM-SIMA.

Purpose of changes (include the issue number and title text for each relevant GitHub issue):

- Closes #1504
- Closes #1505

This pull request is an extensive refactor of `rad_constituents.F90` to facilitate bringing into CAM-SIMA.
- It splits off the previous `rad_constituents.F90` into aerosol-related modules; the original file will only handle gases. This is because RRTMGP port to CAM-SIMA split off radiatively active aerosols from gases+aerosols.
- Isolate CAM-specific data structure retrieval (from pbuf and constituents) to `aerosol_mmr_cam.F90`, which can be modified to use whichever data structures are in other host models (e.g., the CCPP constituents object in CAM-SIMA will replace both pbuf and state for the purpose of radiatively active aerosol.)
- Aerosol-related calls are now prefixed with `rad_aer_*` for clarity.

This pull request also extends the use of the abstract aerosol interface (`aerosol_properties` and `aerosol_state`) into other physics and chemistry code, instead of the previous calls to `rad_cnst_*`.
- `aerosol_properties` and `aerosol_state` are now per-`list_idx` (climate or diagnostic list) instead of having separate `_0list` or `_rlist` functions; separate instantiations of the aerosol properties and state objects per `(aero_model, list_idx)` pair is available. This allows us to keep track of the objects themselves instead of `list_idx`. Since most calls are on the climate list (`list_idx=0`) it will simplify calls and reduce repeating the "0" argument into `rad_cnst_*`.
- `aerosol_instances_mod` owns both `aerosol_properties` (per model/list) and `aerosol_state` (per model/list/chunk) as persistent module data. State objects store only pointers to the persistent `phys_state(lchnk)` and `pbuf` arrays, and is thread-safe. A per-call `create_states`/`destroy_states` factory remains for transient states bound to local copies of `physics_state` (i.e., `microp_aero_run`'s `state1`).
- `aerosol_instances_mod` and its use in `physpkg.F90` most closely mimics the envisioned implementation under CAM-SIMA where `aerosol_properties` and `aerosol_state` become inputs to CCPP physics schemes. Under CAM-SIMA the chunk dimension is removed and objects are passed directly — the factory pattern will be retired.

This pull request also extracts the creation of optics objects and per-bin sw/lw property calculations from `aerosol_optics_cam` to `aerosol_optics_core.F90`; the latter is fully-portable.

After this refactoring, the following modules will be
**fully-portable**:
* `aerosol_properties_mod.F90` - abstract base class.
* `aerosol_state_mod.F90` - abstract base class.
* `radiative_aerosol_definitions.F90` - base types for radiatively active aerosol.
* `bulk_aerosol_properties_mod.F90`
* `modal_aerosol_properties_mod.F90`
* `aerosol_optics_core.F90`

**almost-portable** (need to remove `state` and `pbuf` or minor CAM structures):
* `bulk_aerosol_state_mod.F90` - remove state and pbuf, use host-model MMR module (see below)
* `radiative_aerosol.F90` - use host-model MMR module to resolve indices; diags
* `aerosol_instances_mod.F90` - dechunkize, change init_states to remove pbuf, remove create_states factory once unused

**host-model specific**:
* `aerosol_mmr_cam.F90` - pbuf and state%q indexing to replace with host-model specific code (i.e., CAM-SIMA CCPP constituents)

**needing some future work** to separate out dependencies - MAM/CARMA changes were made at a best-effort here and need to be for another day:
* `modal_aerosol_state_mod.F90`: need to CCPPize `modal_aero_wateruptake_dr` and `modal_aero_calcsize_diag`; role of `modal_aero_data` is to be scoped out (out of scope for this work)
* `carma_aerosol_properties_mod.F90`: depends on `carma_intr` (out of scope for this work)
* `carma_aerosol_state_mod.F90`: depends on `carma_intr` (out of scope for this work)

All changes are **bit-for-bit unchanged** with CAM (based off `cam6_4_156`)

List all files eliminated: N/A

List all files added and what they do:
```
A       src/chemistry/aerosol/aerosol_instances_mod.F90
  - owns persistent aerosol_properties (per model/list) and aerosol_state
    (per model/list/chunk) objects. Provides getters and a per-call factory
    for transient states.

A       src/chemistry/aerosol/radiative_aerosol_definitions.F90
  - split-off structural data and key definitions from rad_constituents.
    purpose of this file is to be the common dependency and not have the gas rad_cnst
    and aerosol rad_aer depend on each other for types and vars like N_DIAG

A       src/chemistry/aerosol/radiative_aerosol.F90
  - split-off aerosol subroutines from rad_constituents.
    this is the "replacement" for rad_constituents in CAM-SIMA

A       src/physics/cam/aerosol_mmr_cam.F90
  - split-off host-model-specific indexing (pbuf, constituents) for radiatively active
    aerosol.

A       src/chemistry/aerosol/aerosol_optics_core.F90
  - split-off logic to create the aerosol_optics_object and per-bin sw/lw optics calculations
    from aerosol_optics_cam.F90 using the abstract aerosol interface.
```

List all existing files that have been modified, and describe the changes:
```
M       src/chemistry/aerosol/aerosol_properties_mod.F90
  - add list_idx_ to store which radiation list this properties instance belongs to.
  - add nbins, nspecies, alogsig, dgnum, dgnumhi, dgnumlo, rhcrystal, rhdeliques
  - add num_to_mass_aer, dryrad properties (for BAM)
  - concrete optics_params implementation for all aerosol models

M       src/chemistry/aerosol/aerosol_state_mod.F90
  - add list_idx_ to store which radiation list this properties instance belongs to.

M       src/chemistry/aerosol/bulk_aerosol_properties_mod.F90
  - implement alogsig (previously a stub)

M       src/physics/cam/rad_constituents.F90
  - remove all code related to aerosol; this module is now gas-only.

M       src/physics/cam/aerosol_optics_cam.F90
  - now uses aerosol_instances_mod to query which aerosol models are active,
    and how many aerosol models are active.
  - now uses aerosol properties and persistent aerosol_state from
    aerosol_instances_mod (replaces per-call create/destroy).
  - moves core calculations to portable aerosol_optics_core.F90.

M       src/physics/cam7/physpkg.F90
M       src/physics/cam/physpkg.F90
  - initialization for radiative_aerosol
  - initialization for aerosol_instances_mod (properties + persistent states)
  - use aerosol_instances_get_state for modal_aer_calcsize, wateruptake
    (replaces per-call create/destroy).

M       src/chemistry/aerosol/bulk_aerosol_state_mod.F90
M       src/chemistry/aerosol/carma_aerosol_properties_mod.F90
M       src/chemistry/aerosol/carma_aerosol_state_mod.F90
M       src/chemistry/aerosol/modal_aerosol_state_mod.F90
M       src/chemistry/aerosol/hygro_aerosol_optics_mod.F90
M       src/chemistry/aerosol/hygrocoreshell_aerosol_optics_mod.F90
M       src/chemistry/aerosol/hygroscopic_aerosol_optics_mod.F90
M       src/chemistry/aerosol/hygrowghtpct_aerosol_optics_mod.F90
M       src/chemistry/aerosol/insoluble_aerosol_optics_mod.F90
M       src/chemistry/aerosol/refractive_aerosol_optics_mod.F90
M       src/chemistry/aerosol/volcrad_aerosol_optics_mod.F90
  - changes to support list_idx change

M       src/chemistry/aerosol/modal_aerosol_properties_mod.F90
  - changes to support list_idx change
  - port mode_size_order from modal_aero_data.
  - add dgnumlo, dgnumhi, dgnum, rhcrystal, rhdeliques
  - remove aero_optics_params in favor of unified implementation

M       src/chemistry/aerosol/aero_wetdep_cam.F90
  - use persistent aerosol_state from aerosol_instances_mod
    (replaces direct modal/carma constructor calls and per-call deallocation).

M       src/chemistry/aerosol/modal_aero_data.F90
M       src/chemistry/carma_aero/carma_aero_gasaerexch.F90
M       src/chemistry/carma_aero/sox_cldaero_mod.F90
M       src/chemistry/carma_aero/aero_model.F90
M       src/chemistry/geoschem/chemistry.F90
M       src/chemistry/modal_aero/aero_model.F90
M       src/chemistry/modal_aero/modal_aero_rename.F90
  - changes to rename rad_cnst_ to rad_aer_ where needed
    (cannot use abstract interface during initialization/register phases.)

M       src/physics/cam/phys_debug.F90
M       src/physics/camrt/radiation.F90
M       src/physics/rrtmg/radiation.F90
M       src/physics/rrtmgp/radiation.F90
M       src/physics/cam/radiation_data.F90
  - changes to rename rad_cnst_ to rad_aer_ where needed
    (these modules will not be ported to CAM-SIMA so were less modified.)

M       src/chemistry/modal_aero/modal_aero_gasaerexch.F90
  - changes to rename rad_cnst to rad_aer_ where needed
    (part of modal_aero model itself so no need for abstract interface.)

M       src/chemistry/mozart/mo_usrrxt.F90
  - changes to rename rad_cnst to rad_aer_ where needed
    (will be superceded by MICM and just dimension setup anyway.)

M       src/physics/cam/microp_aero.F90
  - now use aero_props and aero_state abstract interface.
  - persistent per-chunk aerosol_state moved to aerosol_instances_mod; the
    aerosol_state_object() shim now delegates to aerosol_instances_get_state()
    for backward compatibility with the dycore stepon modules
    (so we do not have to edit the fv3 external)
  - microp_aero_run still uses the per-call factory for state1 (local copy).

M       src/chemistry/modal_aero/dust_model.F90
M       src/chemistry/modal_aero/seasalt_model.F90
M       src/chemistry/mozart/fire_emissions.F90
M       src/chemistry/utils/modal_aero_deposition.F90
  - now use aero_props abstract interface (modal aerosols).

M       src/chemistry/utils/modal_aero_calcsize.F90
M       src/chemistry/utils/modal_aero_wateruptake.F90
  - now use aero_props and aero_state abstract interfaces.

M       src/chemistry/utils/prescribed_aero.F90
  - now uses aerosol_instances_mod::aerosol_instances_is_active to query
    whether the modal aerosol representation is active.

M       src/physics/cam/aer_rad_props.F90
  - now uses aerosol_instances_mod to query which aerosol models are active,
    and how many aerosol models are active.

M       src/physics/cam/aer_vis_diag_mod.F90
  - now use aero_props abstract interface (bulk aerosols).

M       src/physics/cam/ndrop_bam.F90
  - use aero_props and aero_state abstract interfaces (bulk aerosols).

M       src/physics/cam/nucleate_ice_cam.F90
  - use aero_props and aero_state abstract interfaces.

M       src/physics/cam/clubb_intr.F90
M       src/physics/cam/zm_conv_intr.F90
  - remove unused use statements
```


src/chemistry/aerosol/aerosol_instances_mod.F90
```diff
@@ -0,0 +1,344 @@
...
+  type :: aero_state_entry_t
+     class(aerosol_state),      pointer :: obj => null()
+  end type aero_state_entry_t
+
+  ! Module holds aerosol properties objects, dimensioned (iaermod, 0:N_DIAG).
+  type(aero_props_entry_t), allocatable, target :: aero_props_all(:,:)
+
+  ! Persistent per-chunk aerosol state objects, dimensioned (iaermod, 0:N_DIAG, begchunk:endchunk).
+  ! States store pointers to phys_state(c) and pbuf which persist for the run.
+  type(aero_state_entry_t), allocatable, target :: aero_states_all(:,:,:)
```
fvitt: Since this has a chunk dimension, is this a cam specific module, i.e., not portable to CAM-SIMA?

src/chemistry/aerosol/aerosol_instances_mod.F90
```diff
@@ -0,0 +1,344 @@
...
+  ! When retrieving properties via aerosol_instances_get_props, or creating states from
+  ! aerosol_instances_create_states, ensure that the aerosol model matches what is needed (e.g., aero_props%model_is('MAM') == .true.)
+  integer :: num_aero_models_ = 0
+
+  logical :: modal_active_ = .false.
+  logical :: carma_active_ = .false.
+  logical :: bulk_active_  = .false.
+
+contains
+  subroutine aerosol_instances_init()
```
fvitt, line 96: How about a description of what the `_init` subroutine actually does?

src/chemistry/aerosol/aerosol_instances_mod.F90
```diff
@@ -0,0 +1,344 @@
...
+          iaermod = iaermod + 1
+          if (nbulk_aerosols > 0) then
+             aero_props_all(iaermod, ilist)%obj => bulk_aerosol_properties(ilist)
+          end if
+       end if
+    end do
+
+  end subroutine aerosol_instances_init
+
+  function aerosol_instances_get_props(iaermod, list_idx) result(props)
```
fvitt, line 166: Should we document these public interfaces a little better?  Some of these are pretty easy to figure out, but a short description of the inputs and what is returned might be helpful to some users / developers.

src/chemistry/aerosol/aerosol_properties_mod.F90
```diff
@@ -513,6 +488,26 @@ subroutine aero_props_init(self, nbin, ncnst, nspec, nmasses, alogsig, f1,f2, ie
...
+       return
+    end if
+    allocate(self%rhcrystal_(nbin),stat=ierr)
+    if( ierr /= 0 ) then
+       return
+    end if
+    allocate(self%rhdeliques_(nbin),stat=ierr)
+    if( ierr /= 0 ) then
+       return
+    end if
```
fvitt: Should these allocates be contingent on the present of corresponding inputs?

src/physics/cam/aerosol_optics_cam.F90
```diff
@@ -818,19 +700,20 @@ subroutine aerosol_optics_cam_sw(list_idx, state, pbuf, nnite, idxnite, tauxar,
 
     aeromodel: do iaermod = 1,num_aero_models
 
-       aeroprops => aero_props(iaermod)%obj
-       aerostate => aero_state(iaermod)%obj
+       aeroprops => aerosol_instances_get_props(iaermod, list_idx)
+       if (.not. associated(aeroprops)) cycle aeromodel
```
fvitt, line 705: It is good to have this check, but wouldn't the `aeroprops` pointer always be associated?

fvitt (commented):
I started a review.  You have done a lot of nice work on this.  I have a couple of questions on some of this.

src/chemistry/aerosol/aerosol_instances_mod.F90
jimmielin: This file will have a lightly-edited equivalent in CAM-SIMA (https://github.com/jimmielin/CAM-SIMA/blob/hplin/bulk_aero_2/src/aerosol/aerosol_instances_mod.F90) that removes the `aerosol_instances_create_states()` (since there will be no intermediate copy of state `state1` - such operations will be broken up into separate CCPP physics schemes and update tendency in-between) and the chunk dimension. I tried to keep everything else portable.

src/chemistry/aerosol/aerosol_instances_mod.F90
jimmielin: Thanks for the suggestion, I added (this should also answer your question about checking `associated(aeroprops)` below - some diagnostic lists may not have all the aerosol models enabled on the climate list)
```
  ! Determine which aerosol models are active (modal, CARMA, bulk) and
  ! create persistent aerosol_properties objects for each (model, list) pair.
  ! Must be called after the radiative aerosol module has parsed aerosol definitions.
  !
  ! NOTE: A model is "globally active" if the climate list (list 0) has > 0 entries
  ! for aerosol with that representation, but individual diagnostic lists
  ! may have zero entries — in that case the corresponding properties slot is left null.
  !
  ! Callers that use lists other than the climate list thus need to check if the aerosol
  ! model for that combination of (aero_model, list) is associated or not.
```

src/chemistry/aerosol/aerosol_instances_mod.F90
jimmielin: Thanks, added throughout. The top of the module also shows the lifecycle of the module variables and how the subroutines are called in order.

src/chemistry/aerosol/aerosol_properties_mod.F90
jimmielin: Thanks, now contingent on the `present` of optional inputs; if they are not allocated, then the corresponding getters (that now check for associated) will return `-huge(1.0_r8)`

src/physics/cam/aerosol_optics_cam.F90
jimmielin: Thanks for the question, since props are based on the pair of `(aerosol model, list index)` some diagnostic lists might only have modal, or only bulk, etc. and the other props corresponding to models that are not present would be disassociated

e.g.
climate list: (bulk, modal)
diag_list_1: (bulk) only bulk aerosols in this list
diag_list_2: (modal) only modal aerosols in this list
diag_list_3: (bulk, modal)

src/physics/cam/aerosol_optics_cam.F90
```diff
@@ -841,108 +725,272 @@ subroutine aerosol_optics_cam_sw(list_idx, state, pbuf, nnite, idxnite, tauxar,
...
-                      ! ref: Fig. 1d of Jasper F. Kok et al. (2017),
-                      ! Smaller desert dust cooling effect estimated from analysis of dust size and abundance
+          ! Call portable aerosol optics driver:
+          if (associated(geometric_radius)) then
+             ! volcanic_radius:
+             call aerosol_optics_sw_bin(aeroprops, aerostate, ibin, &
+                  ncol, pver, top_lev, nswbands, nlwbands, numrh, &
+                  idx_sw_diag, &
+                  relh(:ncol,:), sulfwtpct(:ncol,:), mass(:ncol,:), crefwsw, crefwlw, &
+                  geometric_radius=geometric_radius(:ncol,:), &
```
fvitt: I believe Fortran allows you to pass a null pointer as an actual argument corresponding to an optional pointer dummy argument.   I you take advantage of this, the `if (associated(geometric_radius))` check could be removed and these 2 calls to `aerosol_optics_sw_bin` can be combined.  This may require some changes in `aerosol_optics_core` to declare `geometric_radius` the dummy argument with both `pointer` and `optional` attributes.

src/physics/cam/aerosol_mmr_cam.F90
```diff
@@ -0,0 +1,977 @@
...
+   module procedure rad_cnst_get_aer_mmr_by_idx
+   module procedure rad_cnst_get_mam_mmr_by_idx
+end interface
+
+! values for constituents with requested value of zero
+real(r8), allocatable, target :: zero_cols(:,:)
+
+public :: aerosol_mmr_init    ! allocate zero_cols
+public :: get_cam_idx
+public :: resolve_mode_idx, resolve_bin_idx
```
fvitt: To be consistent, can `resolve_bin_idx` be declared `public` in a separate line?
```suggestion
public :: resolve_mode_idx
public :: resolve_bin_idx
```

src/physics/cam/aerosol_mmr_cam.F90
```diff
@@ -0,0 +1,977 @@
...
+! CAM-specific aerosol MMR retrieval routines.  These routines access
+! state%q and physics buffer (pbuf) to return mixing ratio pointers.
+!
+!------------------------------------------------------------------------------------------------
+
+use shr_kind_mod,   only: r8 => shr_kind_r8
+
+implicit none
+private
+save
```
fvitt: The `save` statement at the module level is redundant.  This line can be removed.

src/chemistry/aerosol/radiative_aerosol.F90
```diff
@@ -0,0 +1,1326 @@
...
+   module procedure rad_aer_get_info_by_mode_spec
+   module procedure rad_aer_get_info_by_spectype
+end interface
+
+interface rad_aer_get_props
+   module procedure rad_aer_get_props_by_idx
+   module procedure rad_aer_get_mam_props_by_idx
+end interface
+
+! Public routines — aerosol queries (rad_aer_* naming)
```
fvitt: The dash in line 29 is showing up as some strange characters in xxdiff. 
```suggestion
! Public routines -- aerosol queries (rad_aer_* naming)
```

src/chemistry/aerosol/aerosol_instances_mod.F90
```diff
@@ -0,0 +1,362 @@
...
+  logical :: bulk_active_  = .false.
+
+contains
+  ! Determine which aerosol models are active (modal, CARMA, bulk) and
+  ! create persistent aerosol_properties objects for each (model, list) pair.
+  ! Must be called after the radiative aerosol module has parsed aerosol definitions.
+  !
+  ! NOTE: A model is "globally active" if the climate list (list 0) has > 0 entries
+  ! for aerosol with that representation, but individual diagnostic lists
+  ! may have zero entries — in that case the corresponding properties slot is left null.
```
fvitt: Line 78 contains strange characters (the "dash").

src/chemistry/aerosol/aerosol_instances_mod.F90
```diff
@@ -0,0 +1,362 @@
...
+  type(aero_props_entry_t), allocatable, target :: aero_props_all(:,:)
+
+  ! Persistent per-chunk aerosol state objects, dimensioned (iaermod, 0:N_DIAG, begchunk:endchunk).
+  ! States store pointers to phys_state(c) and pbuf which persist for the run.
+  type(aero_state_entry_t), allocatable, target :: aero_states_all(:,:,:)
+
+  ! Number of aerosol models active at runtime.
+  ! Note: Multiple aerosol models can be active at once, e.g., using bulk for volcanic aerosol and modal for others.
+  ! When retrieving properties via aerosol_instances_get_props, or creating states from
+  ! aerosol_instances_create_states, ensure that the aerosol model matches what is needed (e.g., aero_props%model_is('MAM') == .true.)
```
fvitt: Some of these lines in the comments seem extra long.  Maybe limit the line lengths to 80, or so, characters.

fvitt (commented):
A few more random things.

src/physics/cam/aerosol_mmr_cam.F90
jimmielin: Thanks, moved to separate line. Originally as suggested by Simone I tried to merge them together but opted not to within this PR.

src/physics/cam/aerosol_mmr_cam.F90
jimmielin: Thanks, removed.

src/chemistry/aerosol/radiative_aerosol.F90
jimmielin: Thanks, removed throughout.

src/chemistry/aerosol/aerosol_instances_mod.F90
jimmielin: Thanks, removed throughout.

src/chemistry/aerosol/aerosol_instances_mod.F90
jimmielin: Thanks, tried to limit to 80 by rearranging the text a bit.

src/physics/cam/aerosol_optics_cam.F90
jimmielin: Thanks Francis, this is a great idea. I changed to `optional, pointer` and check for both `present` and `associated` in `aerosol_optics_core`.

I wasn't sure of compiler behavior, so I did some digging. You probably know this so I'm just writing it down for future reference.

The F2008 standard is quite clear on the behavior of passing null() to optional dummy arguments (emphasis mine)
> 12.5.2.12 Argument presence and restrictions on arguments not present
A dummy argument or an entity that is host associated with a dummy argument **is not present** if the dummy argument
• does not correspond to an actual argument,
• corresponds to an actual argument that is not present, or
• does not have the ALLOCATABLE or POINTER attribute, and corresponds to an actual argument that
– has the ALLOCATABLE attribute and is not allocated, or
– **has the POINTER attribute and is disassociated.**

The F2003 standard was not very clear on this so I suspect I need a `associated()` check to go together with the `present()` check.

I found an example of this that's currently in CAM

https://github.com/ESCOMP/CAM/blob/140295a2270d4b008b1c75680dacb2e3fce4e736/src/chemistry/utils/modal_aero_calcsize.F90#L1264-L1271

so I think the compiler support should be solid if both checks are used, so that's what I adopted.

src/chemistry/aerosol/aerosol_optics_mod.F90
```diff
@@ -34,8 +34,8 @@ subroutine aeropts_sw_props(self, ncol, ilev, iwav, pext, pabs, palb, pasm)
        integer, intent(in) :: iwav        ! wave length index
        real(r8),intent(out) :: pext(ncol) ! parameterized specific extinction (m2/kg)
        real(r8),intent(out) :: pabs(ncol) ! parameterized specific absorption (m2/kg)
-       real(r8),intent(out) :: palb(ncol) ! parameterized asymmetry factor
-       real(r8),intent(out) :: pasm(ncol) ! parameterized single scattering albedo
+       real(r8),intent(out) :: palb(ncol) ! parameterized single scattering albedo
+       real(r8),intent(out) :: pasm(ncol) ! parameterized asymmetry factor
```
jimmielin, line 38: I noticed these comments were possibly swapped so I just updated them in this PR

src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -1197,23 +1199,24 @@ end subroutine modal_aero_calcsize_sub
 !----------------------------------------------------------------------
 
 
-subroutine modal_aero_calcsize_diag(state, pbuf, list_idx_in, dgnum_m, &
```
fvitt, line 1200: I was initially puzzled by the removal of the `list_idx` argument here.  I believe you are instantiating a separate `aerosol_state` object for each `rad_climate` and `rad_diag` namelist specifier, so the `list_idx` arg is replaced by `aero_state`.   Is my understanding correct here?

src/physics/cam/nucleate_ice_cam.F90
```diff
@@ -610,10 +610,6 @@ subroutine nucleate_ice_cam_calc( &
 
    if (clim_modal_carma) then
 
-      if (.not.(present(aero_props).and.present(aero_state))) then
-         call endrun('nucleate_ice_cam_calc: aero_props and aero_state must be present')
-      end if
```
fvitt, line 615: Is there a need to remove this check?

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -345,7 +347,7 @@ subroutine loadaer( self, aero_props, istart, istop, k,  m, cs, phase, &
 
     do l = 1, aero_props%nspecies(m)
 
-       call self%get_ambient_mmr(l,m, raer)
+       call self%get_ambient_mmr(species_ndx=l, bin_ndx=m, mmr=raer)
        call self%get_cldbrne_mmr(l,m, qqcw)
```
fvitt: Consistent with the line above:
```suggestion
       call self%get_cldbrne_mmr(species_ndx=l, bin_ndx=m, mmr=qqcw)
```

src/physics/cam/nucleate_ice_cam.F90
```diff
@@ -717,7 +713,7 @@ subroutine nucleate_ice_cam_calc( &
 
                                  idxtmp = aer_cnst_idx(m,l)
 
-                                 call aero_state%get_ambient_mmr(l,m,amb_mmr)
+                                 call aero_state%get_ambient_mmr(species_ndx=l, bin_ndx=m, mmr=amb_mmr)
                                  call aero_state%get_cldbrne_mmr(l,m,cld_mmr)
```
fvitt: Can you make the arguments in this line consistent with the line above?

```suggestion
                                 call aero_state%get_cldbrne_mmr(species_ndx=l, bin_ndx=m, mmr=cld_mmr)
```

src/chemistry/aerosol/aerosol_state_mod.F90
```diff
@@ -514,7 +516,7 @@ subroutine icenuc_type_wght_base(self, bin_ndx, ncol, nlev, species_type, aero_p
        if (cldbrne) then
           call self%get_cldbrne_mmr(ispc, bin_ndx, aer_bin)
```
fvitt:
```suggestion
          call self%get_cldbrne_mmr(species_ndx=ispc, bin_ndx=bin_ndx, mmr=aer_bin)
```

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
```diff
@@ -355,47 +407,52 @@ end function number_transported
...
+       call endrun('modal_aerosol_properties_mod%get: dryrad should not be read from file for modal aerosols')
+    end if
+
   end subroutine get
 
   !------------------------------------------------------------------------
   ! returns optics type and table parameters
   !------------------------------------------------------------------------
-  subroutine optics_params(self, list_ndx, bin_ndx, opticstype, extpsw, abspsw, asmpsw, absplw, &
+  subroutine optics_params(self, bin_ndx, opticstype, extpsw, abspsw, asmpsw, absplw, &
```
fvitt: It seems to me that we could have a generalized optics_params method defined in the `aersosol_properties` abstract class module, rather than have each concrete class define it's own method.

src/chemistry/aerosol/radiative_aerosol_definitions.F90
```diff
@@ -0,0 +1,1199 @@
...
+     's-organic', 'black-c  ', 'seasalt  ', 'dust     '/)
+
+  integer, public, parameter :: num_bin_morphs  = 2
+  character(len=8), public, parameter :: bin_morph_names(num_bin_morphs) = &
+       (/ 'shell   ', 'core    ' /)
+
+  !===========================
+  ! Shared constants (shared with rad_constituents for gases) part 1.
+  !===========================
+  integer,          public, parameter :: cs1 = 256
```
fvitt: Could we just use `SHR_KIND_CL`?

src/chemistry/aerosol/radiative_aerosol_definitions.F90
```diff
@@ -0,0 +1,1199 @@
...
+  logical,          public            :: verbose = .true.
```
fvitt: Should this be a parameter?
```suggestion
  logical,          public, parameter :: verbose = .true.
```

src/chemistry/aerosol/radiative_aerosol_definitions.F90
```diff
@@ -0,0 +1,1199 @@
...
+  character(len=1), public, parameter :: nl = achar(10)
```
fvitt: Can `nl` be a longer, more descriptive name?

src/chemistry/utils/modal_aero_calcsize.F90
jimmielin: Yes, that is exactly my intent. Thank you!

src/physics/cam/nucleate_ice_cam.F90
jimmielin: Good question, I was confused here. I put it back but moved it up to where `aero_props` is first used and merged the two `if (clim_modal_carma) ... else ...` clauses

src/chemistry/aerosol/radiative_aerosol_definitions.F90
jimmielin: Thanks, fixed!

src/chemistry/aerosol/radiative_aerosol_definitions.F90
jimmielin: Thanks, updated to `shr_kind_cl`

src/chemistry/aerosol/radiative_aerosol_definitions.F90
jimmielin: Agreed, it's horrible for `grep`ing for it within code too. Changed to `newline`

src/chemistry/aerosol/modal_aerosol_properties_mod.F90
jimmielin: Thanks, this is a great idea. I implemented it in https://github.com/ESCOMP/CAM/pull/1506/commits/cf78638fa59922f935eb02452626a241c7878717 and now can also remove `radiative_aerosol::rad_aer_get_bin_props` since `optics_params` was the only place using it. 

update: verified to be b4b with izumi/nag and izumi/gnu

src/chemistry/aerosol/aerosol_optics_mod.F90
fvitt: Makes sense. Thanks.

src/chemistry/aerosol/aero_wetdep_cam.F90
```diff
@@ -796,10 +801,7 @@ subroutine aero_wetdep_tend( state, dt, dlf, cam_out, ptend, pbuf)
 
     end do bins_loop
 
-    if (associated(aero_state)) then
-       deallocate(aero_state)
-       nullify(aero_state)
-    end if
+    nullify(aero_state)
```
fvitt, line 804: Shouldn't we `deallocate` before `nullify` to prevent memory leaks?

jimmielin: Thanks for the question. For most modules excluding `microp_aero.F90` (which has intermediate `state1` updates) the aerosol properties and aerosol state modules are persistently allocated by `aerosol_instances_mod.F90`:

```fortran
  ! Persistent aerosol properties objects
  ! dimensioned (iaermod, 0:N_DIAG).
  type(aero_props_entry_t), allocatable, target :: aero_props_all(:,:)

  ! Persistent per-chunk aerosol state objects
  ! dimensioned (iaermod, 0:N_DIAG, begchunk:endchunk).
  ! States store pointers to phys_state(c) and pbuf which persist for the run.
  type(aero_state_entry_t), allocatable, target :: aero_states_all(:,:,:)
```
https://github.com/jimmielin/CAM/blob/hplin/rad_cnst_refactor_5_optics/src/chemistry/aerosol/aerosol_instances_mod.F90

so the `aero_wetdep_cam.F90`'s `aero_state` is just pointing to that persistent state:
```fortran
    !REMOVECAM - get persistent state from factory; under CAM-SIMA states will be passed as scheme inputs
    nullify(aero_state)
    do iaermod = 1, aerosol_instances_get_num_models()
       props_tmp => aerosol_instances_get_props(iaermod, 0)
       if (associated(props_tmp)) then
          if (.not. props_tmp%model_is('BAM')) then
             aero_state => aerosol_instances_get_state(iaermod, 0, state%lchnk)
```
https://github.com/jimmielin/CAM/blob/hplin/rad_cnst_refactor_5_optics/src/chemistry/aerosol/aero_wetdep_cam.F90

at the end of the subroutine it only has to be `nullify()`'d but the state should not be deallocated as it is managed by `aerosol_instances_mod.F90`. 

In the future in CAM-SIMA `aero_state`, `aero_props` will be provided as subroutine input pointers so the nullify will also not be necessary. (and the chunk dimension will also be gone.)

Differently, in `microp_aero.F90`, because of `aero_state` requiring access to `state1` and not `state` (since there are aerosol tendencies in ice nucleation) it is created within `microp_aero_run`:
```fortran
   !REMOVECAM: when microp_aero is brought into SIMA intermediate state1 updates should split into separate
   ! physics schemes, run tendency updaters, then the aerosol state is updated, so no need for factory pattern.
   ! create aerosol state objects via factory
   call aerosol_instances_create_states(list_idx=0, state=state1, pbuf=pbuf, aero_states=aero_states1, nstates=nstates1)
   !REMOVECAM_END
```

thus there is a destructor call at the end that deallocates the state object:
```fortran
   ! destroy all aerosol state objects created for this chunk
   nullify(aero_state1_obj)
   call aerosol_instances_destroy_states(aero_states1)
```
https://github.com/jimmielin/CAM/blob/hplin/rad_cnst_refactor_5_optics/src/physics/cam/microp_aero.F90

In CAM-SIMA there won't be "intermediate states" like this (instead they're split into separate schemes running sequentially and the tendencies are updated in-between) so in the CAM-SIMA implementation there will be no `aerosol_instances_{create,destroy}_states`: subroutines do not need to manage aero_props/aero_state anymore and they are all just passed as input pointers

fvitt (approved):
Thank you for the good work on this.

jimmielin:
Thank you so much for your feedback and review, @fvitt!

Hi @cacraigucar, this is merged up to head and I ran the regression tests and they're still bit-for-bit (subject to expected failures). It is ready for tagging from my end. Thank you!

PR #1538, opened 2026-04-15
part of cam6_4_182: Force CAM to build yaml-cpp (vs searching for it first when building TUV-x)

### LINKED ISSUE #1537: TUV-x build issue when users have Conda environments active
### What happened?

Several users have reported build failures relating to TUV-x -- it seems this is tied to the FindPackage call for yaml-cpp, so this introduces a one-line change to force CAM to build this itself, avoiding incompatible versions/environments.

On Derecho, 'module load conda; conda activate; ./case.build' after configuring a case.

(It's more problematic for people who have conda activated by default, since it propagates to qcmd-based builds.)

cam_development

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

_No response_

briandobbins:
There's an issue with CAM / CESM cases and TUV-x when users have Conda environments active, since it finds a yaml-cpp library there, but it often isn't compatible.  This PR introduces a one-line change to force CAM to use its own copy so we have consistent behavior regardless of someone's conda environment.

Fixes #1537

### DIFF
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -242,9 +242,8 @@ def _build_tuvx(caseroot, libroot, bldroot):
         incldir = os.environ.get('USER_INCLDIR')
         if incldir is None:
             incldir = ''
-        #os.environ['USER_INCLDIR'] = incldir + \
-        #    f" -I{_tuvx_include_dir(libroot)} "
-        os.environ['USER_INCLDIR'] = f" -I{_tuvx_include_dir(libroot)} " + incldir
+        os.environ['USER_INCLDIR'] = incldir + \
+            f" -I{_tuvx_include_dir(libroot)} "
 
         # create symlink to library in folder CIME expects libraries to be in
         dst = os.path.join(libroot, "libtuvx.a")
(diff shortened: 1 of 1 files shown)


cime_config/buildlib
```diff
@@ -218,6 +218,7 @@ def _build_tuvx(caseroot, libroot, bldroot):
         cmake_args += "-DCMAKE_Fortran_COMPILER_WORKS=1 "
         cmake_args += "-DCMAKE_C_COMPILER_WORKS=1 "
         cmake_args += "-DCMAKE_CXX_COMPILER_WORKS=1 "
+        cmake_args += "-DCMAKE_DISABLE_FIND_PACKAGE_yaml-cpp=1 "
         if (case.get_value("MACH") == "izumi") :
             cmake_args += f"-DCMAKE_PREFIX_PATH={arg_dict['NETCDF_PATH']} "
             cmake_args += f"-DCMAKE_IGNORE_PATH={os.environ.get('PYTHONHOME')} "
```
fvitt, line 224: With the CMAKE_DISABLE arg at line 221, can the CMAKE_IGNORE_PATH for izumi builds at line 224 be removed?  The conda environment issue sounds like an issue I had on izumi, so the IGNORE_PATH arg was added.

briandobbins: I'm an idiot - I didn't even think of Izumi, and that's a good question.  Let me try things out and report back.  Got a few other things I need to focus on tonight, so it may be tomorrow.

(I'm embarrassed to say I didn't even check out the other -DCMAKE args!)

cacraigucar: any updates on this?

fvitt:
@briandobbins 
If you don't mind I would like to close this PR.  I would like to address this issue an other PR I am preparing to address other TUV-X related issues.  Thanks.

fvitt:
Fixed in tag cam6_4_182

PR #1543, opened 2026-04-23
cam6_4_170: Updates to TEM diags

### LINKED ISSUE #1542: TEM diagnostics updates
The following updates are requested for the TEM diagnostics:

 - Set up default namelist settings so that ESMF regrid method is used by default in cam7 configurations
 - ESMF bilinear interpolation for the zonal mean calculations
 - Use constant scale height of 7 km
 - Correct calculation of potential temperature (Exner bug fix)
 - Vertically interpolate to constant pressures before TEM terms are calcuated

fvitt:
Closes #1542

### DIFF
--- a/bld/namelist_files/namelist_defaults_cam.xml
+++ b/bld/namelist_files/namelist_defaults_cam.xml
@@ -951,6 +951,10 @@
 <cam_physics_mesh hgrid="ne30np4" npg="3">atm/cam/coords/ne30pg3_esmf_20200428.nc</cam_physics_mesh>
 <cam_physics_mesh hgrid="ne120np4" npg="3">share/meshes/ne120pg3_ESMFmesh_cdf5_c20211018.nc</cam_physics_mesh>
 
+<!-- For ESMF regrid zonal-mean TEM diagnostics -->
+<ctem_diags_numlats hgrid="ne16np4">90</ctem_diags_numlats>
+<ctem_diags_numlats hgrid="ne30np4">180</ctem_diags_numlats>
+
 <!-- For scaling lightning sources of NOx -->
 <lght_no_prd_factor                                              >1.00D0</lght_no_prd_factor>
 <lght_no_prd_factor                hgrid="ne30np4"                  >1.80D0</lght_no_prd_factor>
--- a/bld/namelist_files/namelist_defaults_cam.xml
+++ b/bld/namelist_files/namelist_defaults_cam.xml
@@ -938,20 +938,8 @@
 <edyn_grid hgrid="0.47x0.63">320x385</edyn_grid>
 <edyn_grid hgrid="ne120np4" >320x385</edyn_grid>
 
-<!-- ESMF mesh files for physics decomposition -->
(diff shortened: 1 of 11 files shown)


bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -959,18 +959,10 @@
...
-<cam_physics_mesh hgrid="ne5np4" npg="3">atm/cam/coords/ne5np4.pg3_esmf_mesh_c210121.nc</cam_physics_mesh>
-<cam_physics_mesh hgrid="ne16np4">atm/cam/coords/ne16np4_esmf_c210305.nc</cam_physics_mesh>
-<cam_physics_mesh hgrid="ne16np4" npg="3">share/meshes/ne16pg3_ESMFmesh_cdf5_c20211018.nc</cam_physics_mesh>
-<cam_physics_mesh hgrid="ne30np4">atm/cam/coords/ne30np4_esmf_c210305.nc</cam_physics_mesh>
-<cam_physics_mesh hgrid="ne30np4" npg="3">atm/cam/coords/ne30pg3_esmf_20200428.nc</cam_physics_mesh>
-<cam_physics_mesh hgrid="ne120np4" npg="3">share/meshes/ne120pg3_ESMFmesh_cdf5_c20211018.nc</cam_physics_mesh>
+<!-- For ESMF regrid zonal-mean TEM diagnostics -->
+<ctem_diags_numlats>0</ctem_diags_numlats>
+<ctem_diags_numlats hgrid="ne16np4">90</ctem_diags_numlats>
+<ctem_diags_numlats hgrid="ne30np4">180</ctem_diags_numlats>
```
cacraigucar, line 965: The issue said turning on ESMF regrid for all cam7.  It is just setting them for these grids - is that sufficient?

cacraigucar (changes requested):

bld/namelist_files/namelist_defaults_cam.xml
fvitt: @cacraigucar 
Do you think attribute `phys="cam7"` should be included?

cacraigucar: This is a science question.  If it should be set for all resolutions of cam7, then it should probably JUST have `phys="cam7"`.  This would allow it to work for variable resolution grids as well.  If on the other hand, it is to just be used with these two grids and would apply for cam6 runs, then leaving it as-is is fine.

fvitt: In `build-namelist` the default `ctem_diags_numlats` namelist setting is set only for cam7 physics.   We will set the defaults for only these 2 cam7 grids for now. So I am inclined to leave this as it is.

cacraigucar: I confirmed with @PeterHjortLauritzen that this is fine with its limited scope for just these two grids.

cacraigucar (approved):

PR #1554, opened 2026-05-12
cam6_4_173: fix energy bug in physics-dynamics coupling

### LINKED ISSUE #1550: ZM scheme no longer conserving energy
### What happened?

Running F2000climo on the head of the trunk with ```print_energy_errors = .true. ``` reveals lots of energy conservation leaks in the ZM scheme. @PeterHjortLauritzen estimated it to be 6E-2 W/m2. I traced this energy leak back to the cam6_4_022 PR, and in particular the atmospheric_physics update to .gitmodules. The bug is in the atmospheric_physics module ```zm_conv_momtran.F90```. Specifically, this change in the ketend calculation is the bug:

Image

Reverting to the old ketend calculation results in no energy errors.

run F2000climo with ```print_energy_errors = .true. ``` in user_nl_cam.

cam6_4_022

CISL machine (e.g. cheyenne)

Intel

/glade/derecho/scratch/aherring/cam6_4_022_F2000climo_ne30pg3_ne30pg3_mg17_etest.001/

Any CAM SE can do this

### Extra info

_No response_

### LINKED ISSUE #1551: Bug in physics tendency scaling
### What happened?

To support new dynamical cores that use discrete total energy definitions different from CAM physics, we configured the total energy fixer to enforce energy consistent with the dynamical core. This required scaling the physics tendencies in the physics-dynamics coupling layer — and there was a convoluted bug in that scaling!

Run CAM with moist physics

.

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

_No response_

### LINKED ISSUE #1553: RRTMGP diagnostic net_flx bug
### What happened?

@PeterHjortLauritzen and @brianpm identified a bug introduced with the CCPPization of RRTMGP. These two pbuf fields are zeroed every timestep, but should only be zeroed when we're performing the longwave calculation.

```
flns(:) = 0._r8
flnt(:) = 0._r8
```

Fix is to move these lines to within the `if (dolw)` block

Run CAM with print_energy_errors=.true.

cam6_4_171

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

_No response_

PeterHjortLauritzen:
Closes #1551 - Bug in physics tendency scaling
Closes #1553 - RRTMGP diagnostic net_flx bug
Closes #1550 - ZM scheme no longer conserving energy

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,7 +29,7 @@
 [submodule "atmos_phys"]
 	path = src/atmos_phys
 	url = https://github.com/ESCOMP/atmospheric_physics
-	fxtag = atmos_phys0_21_002
+	fxtag = atmos_phys0_21_003
         fxrequired = AlwaysRequired
         fxDONOTUSEurl = https://github.com/ESCOMP/atmospheric_physics
 
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -29,7 +29,7 @@ List all existing files that have been modified, and describe the changes:
 M       .gitmodules
  - Update atmos_phys external to atmos_phys0_21_003
 
-M       src/atmos_phys (external)
+M       src/atmos_phys
  - Update to atmos_phys0_21_003: fix ZM scheme factor for energy conservation
 
 M       src/physics/cam/physpkg.F90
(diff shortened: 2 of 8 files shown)


PeterHjortLauritzen:
Please note that we need to make these changes in CAM-SIMA as well!

jimmielin:
Thanks @PeterHjortLauritzen! I checked and the same issue is present in CAM-SIMA.

As I understand from reading your fix, my proposed fix for CAM-SIMA is to create a registry variable saving `cp_or_cv_dycore_at_start_of_physics_timestep` (to avoid saving it into a temp variable) with the value computed in `d_p_coupling`, where it calls `cam_thermo_water_update` to update `cp_or_cv_dycore`. This value computed at the beginning of the physics timestep is the one that should always be provided to the energy checker/fixer.

When `thermo_water_uptake` is called within physics, updating `cp_or_cv_dycore`, its only intended use is for `tot_energy_phys` diagnostics.

Is that understanding correct? If yes then I can start drafting the equivalent fix in CAM-SIMA which will look different since we don't have `physpkg`.

PeterHjortLauritzen:
Yes in current CAM! Once we move towards a true variable latent heat formulation in physics we would update cp_pr_cv after each parameterization ...

nusbaume (approved):
Thanks @PeterHjortLauritzen!

PR #1557, opened 2026-05-13
Fix for CARMA radiation diagnostics error out (#1556)

### LINKED ISSUE #1556: rad_cnst refactor: list_idx gating in carma_aerosol_state_mod.F90 breaks radiation diagnostics
### What happened?

@tilmes noted that the rad_constituents/abstract aerosol interface refactor introduced some `endrun` calls in `carma_aerosol_state_mod.F90` that broke radiation diagnostics when running CARMA.

@fvitt confirmed that the `endrun` calls were causing the issue.

Tracing the code showed that the calls being hardcoded to `list_idx = 0` to underlying rad_aer_* is correct, so only the `endrun` needs to be removed.

Run latest CAM tag (or any after cam6_4_163) with CARMA and rad_diag_1...

cam6_4_163

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

_No response_

jimmielin:
Fixes #1556 

Fix courtesy of @fvitt 

This just removes `endrun`s that were overzealously added in #1506 since I saw they only called list_idx 0 `rad_aer_get_info_by_bin`. But it seems like that was actually intentional.

### DIFF
--- a/src/chemistry/aerosol/carma_aerosol_state_mod.F90
+++ b/src/chemistry/aerosol/carma_aerosol_state_mod.F90
@@ -407,10 +407,6 @@ contains
     character(len=aero_name_len) :: bin_name, shortname
     integer :: igroup, ibin, rc, nchr, ncol
 
-    if (self%list_idx_ /= 0) then
-       call endrun('carma_aerosol_state::hygroscopicity: only valid for climate list (list_idx=0)')
-    end if
-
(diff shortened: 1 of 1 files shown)


jimmielin:
will be included as part of https://github.com/ESCOMP/CAM/pull/1554

PR #1559, opened 2026-05-15
cam6_4_177: Generalize cloud chemistry

### LINKED ISSUE #1428: Generalize cloud chemistry aerosol processes

Refactor the interactions of aqueous chemistry in clouds with aerosols.  

Generalize code in `mo_cldaero_mod.F90` so that the implementation is aerosol model representation independent. Currently there are 3 different implementations for bulk, carma, and modal aerosols:
```
   src/chemistry/bulk_aero/sox_cldaero_mod.F90
   src/chemistry/carma_aero/sox_cldaero_mod.F90
   src/chemistry/modal_aero/sox_cldaero_mod.F90
```

The aerosol model specific code in these modules should be abstracted away and be implemented in aerosol model specific `aerosol_state` objects.

In the `sox_cldaero_create_obj` routine, the code which sets `conc_obj%no3c`, `conc_obj%nh4c`, and `conc_obj%so4c` should be replaced with generalized code that via abstract aerosol interfaces.  For carma aerosols, the code is high-lighted here:
https://github.com/ESCOMP/CAM/blob/4c480c8e8c7e3138d8d5cb3118f635c71886ecf7/src/chemistry/carma_aero/sox_cldaero_mod.F90#L138C1-L158C1
For modal aerosols, see:
https://github.com/ESCOMP/CAM/blob/4c480c8e8c7e3138d8d5cb3118f635c71886ecf7/src/chemistry/modal_aero/sox_cldaero_mod.F90#L95C1-L148C1

In `subroutine sox_cldaero_update`, the aerosol model specific code that partitions the change in cloud-borne  species across modes or bins should be abstracted away.  For carma aerosols, the code is here:
https://github.com/ESCOMP/CAM/blob/4c480c8e8c7e3138d8d5cb3118f635c71886ecf7/src/chemistry/carma_aero/sox_cldaero_mod.F90#L307C1-L322C1
For modal aerosols see:
https://github.com/ESCOMP/CAM/blob/4c480c8e8c7e3138d8d5cb3118f635c71886ecf7/src/chemistry/modal_aero/sox_cldaero_mod.F90#L266C1-L314C1

@K20shores, @tilmes 

No

Yes

fvitt:
This replaces aerosol model specific `sox_cldaero_mod` modules for aqueous cloud chemistry with a generic aerosol model independent module.  This expands on and makes use of the aerosol abstract interfaces.

Closes #1428


src/chemistry/geoschem/chemistry.F90
```diff
@@ -1573,8 +1578,11 @@ subroutine chem_init(phys_state, pbuf2d)
     ENDIF
 
 #if defined( MODAL_AERO )
+    ! instantiate MAM aerosol properties object
+    aero_props => modal_aerosol_properties()
+
     ! Initialize aqueous chem
-    CALL SOx_inti()
+    CALL SOx_inti(aero_props)
```
jimmielin, line 1590: @jimmielin will check whether this call is needed if GEOS-Chem doesn't call setsox

jimmielin: It seems like this is needed to keep gas-surface exchange happy but the stub here is fine. I might add a suggested change in the full review to actually get the real MAM properties object here instead of the stub, in case GEOS-Chem needs it in the future.

src/chemistry/aerosol/sox_cldaero_mod.F90
```diff
@@ -24,110 +23,79 @@ module sox_cldaero_mod
...
+    integer :: i,k,mm, ntot_amode
+    logical :: mode7
 
-    ! local indexing for bins
-    !integer, allocatable :: bin_idx(:,:) ! table for local indexing of modal aero number and mmr
+    conc_obj => cldaero_allocate()
 
+    if (aero_props%model_is('BAM')) then
+       ! no cloud-borne aerosols
+       conc_obj%xlwc(:ncol,:) = lwc(:ncol,:)*cfact(:ncol,:) ! cloud water L(water)/
```
jimmielin: Was this comment truncated? Looking at the lines below maybe it should be
```suggestion
       conc_obj%xlwc(:ncol,:) = lwc(:ncol,:)*cfact(:ncol,:) ! cloud water L(water)/L(air)
```

src/chemistry/aerosol/sox_cldaero_mod.F90
```diff
@@ -140,44 +108,48 @@ function sox_cldaero_create_obj(cldfrc, qcw, lwc, cfact, ncol, loffset) result(
...
-    conc_obj%so4c = so4mmr
 
-  end function sox_cldaero_create_obj
+    ! *** NOTE ***
+    ! should refactor this bit after merging to later cam tag -- where aero_props%model_is is avail
+    if (ntot_amode>0) then
+       if (.not.mode7) then
+          conc_obj%so4_fact = 1._r8
+       end if
+    end if
```
jimmielin: Does this need to be refactored now that `model_is` is available and this has been merged up?

src/chemistry/bulk_aero/aero_model.F90
```diff
@@ -1069,15 +1076,17 @@ subroutine aero_model_gasaerexch( state, loffset, ncol, lchnk, troplev, delt, re
     real(r8) ::  aqso4_o3(ncol)              ! SO4 aqueous phase chemistry due to O3
     real(r8) ::  xphlwc(ncol,pver)           ! pH value multiplied by lwc
 
+    class(aerosol_state), pointer :: aero_state
+
+!----------------------------------------------------------------------
+    aero_state => bulk_aerosol_state(state, pbuf)
```
jimmielin: I think the `aero_state` being constructed here are not being deallocated (here in bulk and in modal) so each call would leak memory. After the refactoring, it might be cleaner to use `aerosol_instances_mod` to get an existing copy. I will make a PR to your branch for your review:

https://github.com/fvitt/CAM/pull/9

jimmielin (changes requested):
Thanks @fvitt! I had a few minor comments and one set of change requests I made a PR (https://github.com/fvitt/CAM/pull/9) to use `aero_state` and `aero_props` from the instances module instead of constructing in the individual `aero_model`.

src/chemistry/aerosol/sox_cldaero_mod.F90
fvitt: fixed the comment

src/chemistry/aerosol/sox_cldaero_mod.F90
fvitt: I reworked the logic here

src/chemistry/bulk_aero/aero_model.F90
fvitt: Thanks for the PR. I merged it into my branch.

jimmielin (approved):
Thanks @fvitt, I looked at the changes and everything looks good to me, approving now.

PR #1567, opened 2026-06-02
part of  cam6_4_179: moving mountain gw bug fix when ZM is off

### LINKED ISSUE #1565: moving mountain gravity wave scheme fails when ZM scheme is off
### What happened?

The moving mountain gravity wave scheme requires the heating rate from the ZM convection scheme (it is not actually used right now, but planned to be used in the scheme in the future). When the ZM scheme is turned off (via ```deep_scheme='off'```), the moving mountain scheme tries to grab the ZM heating pbuf, which isn't registered, and fails.

Run a CAM7 compset with ```deep_scheme='off'```.

cam6_4_173

CISL machine (e.g. cheyenne)

Intel

_No response_

Yes

### Extra info

@JulioTBacmeister @AnonNick

adamrher:
Fixes #1565

### DIFF
--- a/src/physics/cam/gw_drag_cam.F90
+++ b/src/physics/cam/gw_drag_cam.F90
@@ -552,15 +552,16 @@ subroutine gw_drag_cam_init()
      vpwp_clubb_gw_idx   = pbuf_get_index('VPWP_CLUBB_GW')
      wpthlp_clubb_gw_idx = pbuf_get_index('WPTHLP_CLUBB_GW')
      vort4gw_idx         = pbuf_get_index('VORT4GW')
-
-     ttend_dp_idx        = pbuf_get_index('TTEND_DP')
   endif
 
   if (use_gw_oro .or. use_gw_rdg_beta .or. use_gw_rdg_gamma) then
      sgh_idx = pbuf_get_index('SGH')
   endif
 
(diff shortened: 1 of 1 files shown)


jimmielin (commented):
Thanks @adamrher, this looks good. I am planning to bring it in together with https://github.com/ESCOMP/atmospheric_physics/pull/407 which fixes the ZM parentheses bug.

I will also edit this so if `use_gw_convect_dp` is on but `ttend_dp` is unavailable it aborts since I think that is the intended behavior, but please let me know if you want to allow the deep convection scheme to run with zero tendencies, too.

adamrher:
Thanks! There is build-namelist logic that sets ```use_gw_convect_dp=.false.``` if ```deep_scheme='off'```, so I'll leave up to you if you want to an add'l check in gw_drag.F90.

https://github.com/ESCOMP/CAM/blob/8065f6a93790446d1521fff82bff49b8126964e8/bld/build-namelist#L3976-L3979

I don't think we want to put similar logic in build-namelist for ```use_gw_movmtn_pbl``` as convective heating is an optional future enhancement; the default version doesn't need convective heating.

jimmielin (approved):
Thanks for the correction @adamrher, I didn't know about that bit in `build-namelist`. I think this is ready to go in as-is. I will merge it in together with the ZM fix since both are now expected to be bit-for-bit.

PR #1574, opened 2026-06-05
Add COSP's EarthCARE lidar simulator to CAM output

jshaw35:
COSP2 includes a lidar simulator for the new ESA EarthCARE satellite (ATLID) (https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2015JD023919). This simulator package makes use of existing COSP features for the CALIPSO/CALIOP simulator with some additional changes. With EarthCARE recently launched and additional work to develop the ATLID simulator ongoing, I believe that adding this capability to CAM will be of significant value to the broader community.

CAM's coupler currently does not support ATLID outputs for a few reasons:
1. ATLID inputs, outputs, dimensions, etc are not read into the standard COSP DDTs in cospsimulator_intr.F90
2. There is not a namelist field to enable ATLID outputs analogously to other COSP simulators.
3. The lidar_optics routine in CAM is hardcoded for the CALIOP wavelengths, whereas the updated version in COSP can accomodate both CALIOP and ATLID by supplying the wavelength as an additional input. Additionally, the CAM code now includes radiatively-active snow in lidar_optics whereas I believe COSP does not. These diverging changes should be reconciled.

This was initially a *draft* PR to resolve these issues. It was tested it in more detail (specifically make sure that lidar_optics issues with snow and wavelength are reconciled correctly), but wanted to open this PR now to gauge feedback. @brianpm

### DIFF
--- a/src/physics/cam/cospsimulator_intr.F90
+++ b/src/physics/cam/cospsimulator_intr.F90
@@ -921,6 +921,7 @@ CONTAINS
        call add_default('CLDHGH_ATLID',cosp_histfile_num,' ')
        call add_default('CLDTOT_ATLID',cosp_histfile_num,' ')
        call add_default('CLD_ATLID',cosp_histfile_num,' ')
+       call add_default('BETAMOL_ATLID',cosp_histfile_num,' ')
        call add_default('CFAD_SR355_ATLID',cosp_histfile_num,' ')
     end if
 
@@ -1119,7 +1120,7 @@ CONTAINS
        end if
 
        if (latlid_sim) then
-          call add_default('BETAMOL_ATLID',cosp_histfile_num,' ')
+          call add_default('BETATOT_ATLID',cosp_histfile_num,' ')
        end if
     end if
 
@@ -2451,7 +2452,7 @@ CONTAINS
     cospIN%cospswathsIN = cospswathsIN
     call t_stopf('construct_cospIN')
 
-    if (lradar_sim .or. (llidar_sim .or. (lisccp_sim .or. (lmisr_sim .or. lmodis_sim)))) then
+    if (lradar_sim .or. (llidar_sim .or. (lisccp_sim .or. (lmisr_sim .or. (lmodis_sim .or. latlid_sim))))) then
        call t_startf("subsample_and_optics")
        ! The arrays passed here contain only active columns and the limited vertical
        ! domain operated on by COSP.  Unsubscripted array arguments have already been
--- a/src/physics/cam/cospsimulator_intr.F90
+++ b/src/physics/cam/cospsimulator_intr.F90
@@ -642,7 +642,7 @@ CONTAINS
             bounds_name='cosp_tau_bnds', bounds=taulim_cosp)
     end if
 
-    if (lisccp_sim .or. llidar_sim .or. lradar_sim .or. lmisr_sim) then
+    if (lisccp_sim .or. llidar_sim .or. lradar_sim .or. lmisr_sim .or. latlid_sim) then
        call add_hist_coord('cosp_scol', nscol_cosp, 'COSP subcolumn',          &
             values=scol_cosp)
     end if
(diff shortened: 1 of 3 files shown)


cacraigucar:
@jshaw35:  I moved this to draft as the text indicated it should be.  Please move it out of draft and let us know when it is ready for review.

jshaw35:
@cacraigucar this PR is now functioning as expected and is ready for review!

@brianpm feel free to give this a go. One thought on high-resolution simulations is to turn down the number of subcolumns (cosp_ncolumns, currently 50 and adjustable in the namelist), since CAM should be resolving much more features at high resolution anyway.

src/physics/cam/cospsimulator_intr.F90
```diff
@@ -2779,6 +2864,15 @@ subroutine cospsimulator_intr_run(state, pbuf, cam_in, emis, coszrs,     &
           end do
        endif
 
+       if (latlid_sim) then
+          do ih=1,nht_cosp
+             do is=1,nsr_cosp
+                ihs=(ih-1)*nsr_cosp+is
+                cfad_sr355_atlid(i,ihs) = cfad_sr355_atlid(i,ihs)  ! Already reshaped from cospOUT? Check
```
brianpm: Confirmed shape? Remove comment if so.

brianpm (approved):
I went through the code; looks good to me. I didn't run any tests, but assuming things are working then this looks great to me.

src/physics/cosp2/optics/cosp_optics.F90
```diff
@@ -319,6 +324,25 @@ subroutine lidar_optics(npoints,ncolumns,nlev,npart,ice_type,q_lsliq,q_lsice,q_c
...
     
+    ! Which LIDAR frequency are we using?
+    if (lidar_freq .eq. 355) then
+       Cmol   = Cmol_355nm
+       rdiffm = rdiffm_355nm
+    endif
+    if (lidar_freq .eq. 532) then
+       Cmol   = Cmol_532nm
+       rdiffm = rdiffm_532nm
+    endif
```
cacraigucar: We favor `==` instead of `.eq.`.  Also, suggest use an `elseif` for the seond one.  Should there be an `else` which sets the two variables to zero or some other initialized value?   Tthese were always set in the previous version.

cacraigucar (commented):
cosp has a number of settings which CAM does not test in its regression test suite.   We do not appear to test any of the lidar settings. 

1) Have you run with all the different settings and made sure that your changes didn't accidentally break one or more?

2) Should we add a category "cosp" to our testlist_cam.xml file so that when cosp changes are made we can easily run cosp tests to make sure they still run?  If you think this is a good idea, I can meet with you and describe how to set this up (it is fairly easy, but requires someone with knowledge about COSP settings to set up the appropriate tests).

jshaw35:
> cosp has a number of settings which CAM does not test in its regression test suite. We do not appear to test any of the lidar settings.
> 
>     1. Have you run with all the different settings and made sure that your changes didn't accidentally break one or more?
> 
>     2. Should we add a category "cosp" to our testlist_cam.xml file so that when cosp changes are made we can easily run cosp tests to make sure they still run?  If you think this is a good idea, I can meet with you and describe how to set this up (it is fairly easy, but requires someone with knowledge about COSP settings to set up the appropriate tests).

1. I tested with multiple configurations of other COSP simulators to ensure there were no hidden simulator interdependencies I was missing, but I did not rigorously check all combinations (I think this would be pretty extensive and would likely require creating a manual given the many different COSP settings).
2. I think this sounds like a good idea. My bandwidth is very limited but if you think it would take <2 hours I could allocate time on June 30th to meet at ML.

re: Lidar tests in COSP. A subset of the CALIPSO lidar outputs are likely tested if current tests use cosp_amwg. However, this will not include subcolumn outputs, which are triggered by another namelist option.

src/physics/cam/cospsimulator_intr.F90
jshaw35: Line 2871 applies an identity function so this code block can actually be removed entirely.

src/physics/cosp2/optics/cosp_optics.F90
jshaw35: I have taken this code block directly from the COSP repository and will opt to only change the `.eq.` statements in order to stay as consistent as possible with COSP source code. The encompassing function is only called in settings where `lidar_freq` is hardcoded to 355 or 532 so I don't think an additional guard is needed here.

PR #1577, opened 2026-06-10
cam6_4_182: Updates to CAM build of TUV-X lib

### LINKED ISSUE #1537: TUV-x build issue when users have Conda environments active

### LINKED ISSUE #1570: TUV-x copies input files to run directory regardless of whether TUV-x is active
### Issue Type

Other (please describe below)

### Issue Description

It seems like in any case TUV-x will copy its input files to the run directory even when it is not active:
```
run: ls data
total 80
drwxrwxr-x 2 hplin ncar 12288 Jun  2 13:45 cross_sections
-rw-rw-r-- 1 hplin ncar  7094 Jun  2 13:45 dose_rates.json
drwxrwxr-x 3 hplin ncar  4096 Jun  2 13:45 grids
-rw-rw-r-- 1 hplin ncar 42515 Jun  2 13:45 photolysis_rate_constants.json
drwxrwxr-x 4 hplin ncar  4096 Jun  2 13:45 profiles
drwxrwxr-x 2 hplin ncar  4096 Jun  2 13:45 quantum_yields
drwxrwxr-x 2 hplin ncar  4096 Jun  2 13:45 spectral_weights
```

Maybe it doesn't have to do this. Just opening an issue to track it in case we want to modify this behavior, together with the building TUV-x always even when it is not used.

### Will this change answers?

No

Any CAM SE can do this

fvitt:
This builds the TUV-x library only when CAM is configured for TUV-X.  A configure option (-tuvx) has been added which is required for cases that use TUV-X photolysis.  

Closes #1537 
Closes #1570

### DIFF
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -752,7 +752,7 @@ if (not $waccm_phys) {
 
 # Check TUV-X configuration
 my $tuvx_active = $nl->get_value('tuvx_active');
-if ($tuvx_active =~ /$TRUE/) {
+if ($tuvx_active =~ /$TRUE/io) {
     my $tuvx_config = $cfg->get('tuvx');
     if (!$tuvx_config) {
 	die "$ProgName - ERROR: must use -tuvx CAM configure flag \n";
--- a/bld/build-namelist
+++ b/bld/build-namelist
@@ -750,6 +750,15 @@ if (not $waccm_phys) {
     add_default($nl, 'p_bot_for_radmrg');
 }
 
+# Check TUV-X configuration
+my $tuvx_active = $nl->get_value('tuvx_active');
+if ($tuvx_active =~ /$TRUE/) {
+    my $tuvx_config = $cfg->get('tuvx');
+    if (!$tuvx_config) {
+	die "$ProgName - ERROR: must use -tuvx CAM configure flag \n";
+    }
+}
+
 # Solar irradiance
 
 # First check that solar_const and solar_irrad_data_file are not both defined
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -6,7 +6,7 @@ create the cam library
 # pylint: disable=multiple-imports, wrong-import-position, wildcard-import
 # pylint: disable=unused-wildcard-import, bad-whitespace, too-many-locals
 # pylint: disable=invalid-name
-import sys, os, filecmp, shutil
+import sys, os, filecmp, shutil, re
 from glob import glob
 
 _CIMEROOT = os.environ.get("CIMEROOT")
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -253,9 +253,9 @@ def _build_tuvx(caseroot, libroot, bldroot):
             cmake_args += "-DCMAKE_Fortran_COMPILER_WORKS=1 "
             cmake_args += "-DCMAKE_C_COMPILER_WORKS=1 "
             cmake_args += "-DCMAKE_CXX_COMPILER_WORKS=1 "
+            cmake_args += "-DCMAKE_DISABLE_FIND_PACKAGE_yaml-cpp=1 "
             if (case.get_value("MACH") == "izumi") :
                 cmake_args += f"-DCMAKE_PREFIX_PATH={arg_dict['NETCDF_PATH']} "
-                cmake_args += f"-DCMAKE_IGNORE_PATH={os.environ.get('PYTHONHOME')} "
             if (case.get_value("MACH") == "izumi") and (case.get_value('COMPILER') == "nag") :
                 cmake_args += "-DCMAKE_Fortran_FLAGS='-C=all -g ' "
             else :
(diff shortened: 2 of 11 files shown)


cacraigucar (approved):
Recommend possibly looking at the #ifdef's one more time to see if all the code they encapsulate really needs to be encapsulated.  I realize this is the easiest way to proceed and since it is deep in mozart code, I'm not going to push back on it.  I'm guessing some of the encapsulation could be external and controlled by a passed in use_tuvx logical.  Again, nothing I am going to require though

cime_config/buildlib
```diff
@@ -232,60 +232,68 @@ def _build_tuvx(caseroot, libroot, bldroot):
...
+            dst = os.path.join(libroot, "libyaml-cpp.a")
+            if os.path.isfile(dst):
+                os.remove(dst)
+            os.symlink(_yaml_cpp_lib_path(libroot), dst)
+
+            # Add TUVX to CAM linked library list:
+            linked_libs = case.get_value("CAM_LINKED_LIBS",
+                                         subgroup="build_component_cam")
+            tuvx_link_opts = " -ltuvx -lyaml-cpp"
+            case.set_value("CAM_LINKED_LIBS", linked_libs + tuvx_link_opts)
```
jimmielin: I assume this was tested on Izumi GNU, so if it works to build TUV-x there then please ignore.

I noticed that this will _append_ to `CAM_LINKED_LIBS`. The previous value in `config_component.xml` was `-ltuvx -lyaml-cpp -lstdc++`

with this modified code it will be `-lstdc++ -ltuvx -lyaml-cpp`; I suspect `yaml-cpp` might need `stdc++` to be linked after it, so this order might break on GNU which is single-pass. If that does happen, we might have to do a prepend instead.

cime_config/buildlib
```diff
@@ -232,60 +232,68 @@ def _build_tuvx(caseroot, libroot, bldroot):
...
-        dst = os.path.join(libroot, "libtuvx.a")
-        if os.path.isfile(dst):
-            os.remove(dst)
-        os.symlink(_tuvx_lib_path(libroot), dst)
-        dst = os.path.join(libroot, "libyaml-cpp.a")
-        if os.path.isfile(dst):
-            os.remove(dst)
-        os.symlink(_yaml_cpp_lib_path(libroot), dst)
+    with Case(caseroot, read_only=False) as case:
+        config_opts = case.get_value("CAM_CONFIG_OPTS")
```
jimmielin, line 236: I tested (by repeatedly running `./case.build` in one case directory) and this approach appears to keep appending the TUV-x linker options into `CAM_LINKED_LIBS` (even if they already exist) in subsequent compiles:

```
./xmlquery CAM_LINKED_LIBS
	CAM_LINKED_LIBS: -lstdc++ -ltuvx -lyaml-cpp -ltuvx -lyaml-cpp
```

which may be problematic. I suggest removing the existing linking flags here (unconditionally), then appending again if `-tuvx` is in config options, in case the same case directory changes the config opts. A rare case, but still possible:

Also need to add `import re` on top of this file.

```suggestion
        config_opts = case.get_value("CAM_CONFIG_OPTS")

        # Remove any TUV-x link options set by a previous build, so that
        # re-building the case does not accumulate duplicate or stale
        # options in CAM_LINKED_LIBS:
        linked_libs = case.get_value("CAM_LINKED_LIBS",
                                     subgroup="build_component_cam")
        tuvx_link_opts = " -ltuvx -lyaml-cpp"
        new_linked_libs = re.sub(r"\s*-ltuvx -lyaml-cpp", "", linked_libs)
```

cime_config/buildlib
jimmielin: Same comment regarding duplicate / repeated appends. I would suggest
```suggestion
            # Add TUVX to CAM linked library list:
            new_linked_libs = tuvx_link_opts.strip() + " " + new_linked_libs

        if new_linked_libs != linked_libs:
            case.set_value("CAM_LINKED_LIBS", new_linked_libs)
```

bld/build-namelist
```diff
@@ -750,6 +750,15 @@ if (not $waccm_phys) {
     add_default($nl, 'p_bot_for_radmrg');
 }
 
+# Check TUV-X configuration
+my $tuvx_active = $nl->get_value('tuvx_active');
+if ($tuvx_active =~ /$TRUE/) {
```
jimmielin: I would suggest matching case-insensitive (consistent with other matches in `build-namelist`):

```suggestion
if ($tuvx_active =~ /$TRUE/io) {
```

jimmielin (changes requested):
Thanks @fvitt. I have a few suggested changes for improvements in the way linker flags are added.

jimmielin (approved):
Thanks @fvitt, everything looks good to me now!

PR #1583, opened 2026-06-13
Change specification for CO2 concentration-driven compsets - option 1

billsacks:
We have decided to use the CESM2 mechanism for CO2 – `_BGC%BDRD`, `_BGC%BPRP` or no `_BGC` specifier rather than a `C` or `E` at the end of the time period. This PR makes changes towards that.

I'm not sure what the intention was for the compsets that currently had a `C` at the end of the time period. This PR assumes that the intention was that these operate the same as `_BGC%BDRD`.

Should be coordinated with https://github.com/ESCOMP/CESM/pull/407.

Note that @gold2718 's changes in https://github.com/ESCOMP/CAM/pull/1581 and some additional changes (which may be folded into that PR or may come in separately) are needed for co2_cycle to work correctly.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -251,14 +251,14 @@
       <value compset="1850_CAM60%WCTS"   >waccm_tsmlt_1850_cam6</value>
       <value compset="1850_CAM60%WCCM"   >waccm_ma_1850_cam6</value>
       <value compset="1850_CAM60%WCSC"   >waccm_sc_1850_cam6</value>
-      <value compset="1850[CE]?_CAM70%LT"     >1850_cam_lt</value>
-      <value compset="1850[CE]?_CAM70%MT"     >1850_cam_mt</value>
+      <value compset="1850_CAM70%LT"     >1850_cam_lt</value>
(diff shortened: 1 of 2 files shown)


billsacks:
Either this OR https://github.com/ESCOMP/CAM/pull/1584 should come in; not both! @klindsay28 or @fvitt do you know which is correct?

billsacks:
Also, we may want to change the aliases for these changed compsets; I'll leave that step up to the CAM group.

billsacks:
I'm coming to think that this is the wrong option, and that the right thing to do is #1584 (see also discussion in https://github.com/ESCOMP/CESM/pull/407).

I'm realizing that this PR will probably change answers for the compsets that now have `_BGC%BDRD`, which is unfortunate because it would be ideal to verify that the removals of `[CE]?` in the compset matches don't change answers. (#1584, in contrast, I think will be bit-for-bit.)

billsacks:
Based on discussion with @klindsay28, F compsets should probably have no `_BGC` specifier. So I'm closing this PR in favor of #1584 .

PR #1584, opened 2026-06-13
cam6_4_187: Change specification for CO2 concentration-driven compsets

billsacks:
**THIS PR WILL BE BROUGHT IN IMMEDIATELY AFTER THE CESM ALPHA TAG WITH THE ACCUMULATED CAM TAGS.**

We have decided to use the CESM2 mechanism for CO2 – `_BGC%BDRD`, `_BGC%BPRP` or no `_BGC` specifier rather than a `C` or `E` at the end of the time period. This PR makes changes towards that.

Based on discussions with @klindsay28, it probably makes sense for F compsets to *not* have a _BGC specifier, ~~in contrast to B compsets~~ (update 2026-06-18: actually, we're skipping the _BGC specifier for concentration-driven B compsets, too, at least for now). So this PR simply removes the `C` at the end of the time period without adding `_BGC%BDRD`.

Should be coordinated with https://github.com/ESCOMP/CESM/pull/407.

Note that @gold2718 's changes in https://github.com/ESCOMP/CAM/pull/1581 and some additional changes (which may be folded into that PR or may come in separately) are needed for co2_cycle to work correctly. However, the changes in this PR don't add any triggering of co2_cycle, so this PR could be brought in before #1581. #1581 *will* be needed for the emissions-driven compsets in https://github.com/ESCOMP/CESM/pull/407 to work.


billsacks:
Either this OR https://github.com/ESCOMP/CAM/pull/1583 should come in; not both! @klindsay28 or @fvitt do you know which is correct?

billsacks:
Also, we may want to change the aliases for these changed compsets; I'll leave that step up to the CAM group.

billsacks:
I'm coming to think that this is the right option (**not** https://github.com/ESCOMP/CAM/pull/1583), but I want confirmation from others (like @klindsay28 or @fvitt ). See also discussion in https://github.com/ESCOMP/CESM/pull/407.

I *think* this PR should be bit-for-bit, which is good because it would be good to verify that the removals of `[CE]?` don't change answers for anything.

billsacks:
Based on discussion with @klindsay28, F compsets should probably have no _BGC specifier, so this PR seems like the correct option. I'm therefore taking it out of draft.

billsacks:
> Based on discussion with @klindsay28, F compsets should probably have no _BGC specifier, so this PR seems like the correct option. I'm therefore taking it out of draft.

However, we'd welcome confirmation of this from @fvitt or others.

billsacks:
From talking with @fvitt - we want confirmation from others on this, especially with what's wanted for compset aliases... so @nusbaume and @cacraigucar - tentatively assigning you as reviewers.

fvitt (approved):

cacraigucar (approved):

Issue #1592, opened 2026-06-16
check_energy tracer mass-conservation check never runs: 'exit' instead of 'cycle'


duncanwp:
## Summary

The tracer mass-conservation check in `check_energy.F90` has never actually run: the loop over constituents uses `exit` where it should use `cycle`, so it terminates on the very first iteration.

## Location

`src/physics/cam/check_energy.F90:237-241` (and again at 325-328)

```fortran
do m = 1,pcnst
   if ( any(m == (/ 1, ixcldliq, ixcldice, &
                       ixrain,   ixsnow, ixgrau /)) ) exit   ! dont process water substances
```

## Problem

Water vapor is always constituent `m = 1`, so the condition is true on the first iteration and `exit` terminates the **entire** loop. The comment ("dont process water substances") shows the intent was `cycle` (skip water species, check the rest).

Consequence: `check_tracers_init` computes nothing, and `check_tracers_chng` — called throughout `physpkg.F90` after chemistry, diffusion, and convection — never checks anything, including its `endrun` on conservation violations. This has been present since the initial CAM6 release: the model's tracer-conservation guard has been a silent no-op for years.

## Related

The error test that becomes live once this is fixed, `maxval(tracer_rer)` at `check_energy.F90:385`, reduces over a mostly-uninitialized `(pcols,pcnst)` array and will need fixing at the same time.

## Suggested fix

Change `exit` to `cycle` at both sites, and fix the `tracer_rer` reduction bounds.

---
_Found via a systematic line-by-line scientific code review of `src/physics/cam/` at commit `21a78294` (≈cam6_4_180). This is a candidate bug identified by code inspection; it has not yet been confirmed by a model run. An issue-tracker search on 2026-06-11 found no existing report._

cacraigucar:
@PeterHjortLauritzen - I am working on a misc tag, and would like to include this fix if you verify that it is a correct thing to do

PeterHjortLauritzen:
Some notes from looking into this:

1. Yes, this is a bug — exit should have been cycle, and as a result check_tracers_init/check_tracers_chng have been no-ops since they only ever reached m = 1 (water vapor) before terminating the loop. The uninitialized tracer_rer in the maxval reduction is confirmed as well.

2. However, simply activating the check (as we tried on a branch) reveals deeper design problems, so the fix is not just exit → cycle:   

[...]

PR #1607, opened 2026-07-08
part of cam6_4_192: Remove defunct tracer mass-conservation check

PeterHjortLauritzen:
The check has been a no-op since CAM6 ('exit' instead of 'cycle' terminated the constituent loop on water vapor; see issue #1592). Activating it reveals the design is not valid: individual chemical species are not conserved under reactions (only the column sum is). Closes https://github.com/ESCOMP/CAM/issues/1592

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,5 +1,29 @@
 ===============================================================
 
+Tag name: cam6_4_186
+Originator(s): pel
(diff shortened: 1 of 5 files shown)


nusbaume (approved):
Thanks @PeterHjortLauritzen!

PR #1623, opened 2026-07-17
Set nx and ny to 1 for single column

### LINKED ISSUE #1622: atm nx and ny attributes are incorrect for single-column runs
### What happened?

When running a single-column case, the atm nx and ny attributes passed to CMEPS are incorrectly set to global sizes rather than 1.

This is observable in case2 of SCT - e.g., `SCT_D_Ln7.ne3_ne3_mg37.QPC6.derecho_intel.cam-scm_prep_c6` or `SCT_D_Ln7.ne3_ne3_mg37.QPC5.derecho_intel.cam-scm_prep`. Observe the dimension sizes of the cpl hist file: the nx is 488 for atm dimensions, when this should be 1. (The first grid cell is filled with the actual data.)

cam6_4_173

CISL machine (e.g. derecho)

Intel

/glade/derecho/scratch/sacks/cesm3_0_alpha09d_fixscol_0715a/SCT_D_Ln7.ne3_ne3_mg37.QPC6.derecho_intel.cam-scm_prep_c6.C.20260716_092504_k4ahu7/run/case2run

Yes

### Extra info

As far as I know, this doesn't cause any significant problems. But it leads to larger-than-necessary cpl hist and cpl restart files, and could potentially lead to future issues.

I have a suggested fix incoming.

billsacks:
Resolves #1622 - see that issue for details.

### DIFF
--- a/src/cpl/nuopc/atm_comp_nuopc.F90
+++ b/src/cpl/nuopc/atm_comp_nuopc.F90
@@ -762,7 +762,12 @@ contains
        call export_fields( gcomp, model_mesh, model_clock, cam_out, rc=rc )
        if (ChkErr(rc,__LINE__,u_FILE_u)) return
 
-       call get_horiz_grid_dim_d(hdim1_d, hdim2_d)
+       if (single_column) then
+          hdim1_d = 1
+          hdim2_d = 1
+       else
+          call get_horiz_grid_dim_d(hdim1_d, hdim2_d)
+       end if
        call State_SetScalar(dble(hdim1_d), flds_scalar_index_nx, exportState, &
             flds_scalar_name, flds_scalar_num, rc)
        if (ChkErr(rc,__LINE__,u_FILE_u)) return


billsacks:
I am not totally confident in this fix, so would be happy to have someone else look at this and think about whether this seems like the right thing to do.

@mvertens - FYI, and I'd welcome your eyes on this, too.

billsacks:
I ran two tests with comparisons against baselines:
```
SCT_D_Ln7.ne3_ne3_mg37.QPC5.derecho_intel.cam-scm_prep
SCT_D_Ln7.ne3_ne3_mg37.QPC6.derecho_intel.cam-scm_prep_c6
```

Both tests also included the changes in https://github.com/ESCOMP/CDEPS/pull/422 and https://github.com/ESCOMP/CMEPS/pull/679, but I don't think those changes are needed.

Tests passed and baseline comparisons passed.

I also manually compared the CAM and CPL hist files from case2 (the single-column case) in these two tests against baselines. CAM hist files were bit-for-bit. CPL hist files had different dimension sizes, as expected / intended: the new test has dimension sizes of 1 for all variables in the cpl hist files. I created versions of the baseline cpl hist files where I extracted just the first point for each dimension and then redid the comparison; this comparison was bit-for-bit, as hoped for.

billsacks:
@jtruesdal - FYI

PR #1631, opened 2026-07-23
part of cam6_4_192: Fix mismatched species MW in GEOS-Chem 14.5 in CAM

### LINKED ISSUE #1630: GEOS-Chem coupling has incorrect molecular weights in mo_sim_dat.F90 (mismatched) breaking MAM coupling
### What happened?

It appears that the advected mw in GEOS-Chem mo_sim_dat.F90 is wrong.

Since the MAM <-> bulk coupling uses VMR/MMR conversions this means the MAM coupling is broken.

Observe bc_a4 concentrations (~1/10 of what it should be) in a modeled run and AerMassSO4 (~1/3 of what it should be)

cam6_4_090

CISL machine (e.g. derecho)

Intel

_No response_

Yes

### Extra info

This is answer changing for FCHIST_GC compsets

Copying @lizziel

jimmielin:
Fixes #1630 

When GEOS-Chem v14.5 was brought in EBZ was added to solsym at index 89 without its 106.167 value being added to adv_mass; and TMB was moved from the non-advected tail into the advected block, but its old 106.167 was left behind between TLFUO2 and TRO2

This has immediate consequences for sulfate levels and seasalt and bc_a4 due to broken MAM coupling.

All users of GEOS-Chem within CESM running a tag more recent than cam6_4_090 should be advised to apply this fix immediately.


lizziel (approved):
Apologies if I brought in that bug!!!

PR #1643, opened 2026-08-11
Reduce CAM initialization memory footprint that is important for ne1024 simulation- #4102

### LINKED ISSUE #1640: Out of memory issue in CAM at ne1024 resolution

CAM initialization consumes excessive memory (and time) at very high resolution (e.g. ne1024), to the point of risking out-of-memory during startup. Currently we have to undersubscribe a compute node on Derecho, which is probably unavoidable, but we are seeking ways to reduce the memory usage in CAM initialization.

A large fraction of the per-rank init footprint is the read-only lookup tables that are bit-identical on every rank — photolysis cross-sections and radiative source functions, aerosol optics tables, soil erodibility, and the global lat-lon source fields staged by `tracer_data`. Every rank allocates and holds its own copy, so the node-level cost scales with ranks-per-node while the useful information does not.

@jtruesdal 

No

Yes

sjsprecious:
This PR reduces the memory usage during CAM initialization by adding a new module named `cam_shmem_mod.F90`, which allocates the read-only looku tables once per shared-memory node and maps it into every rank on that node.

Then the following modules are updated to reduce duplicated fields/tables:

- **mo_jlong** — **xsqy**, **rsf_tab**, **xs_o2b/o3a/o3b**. Also changes the bulk broadcast from all-ranks to leaders-only, cutting the inter-node broadcast volume.
- **phys_prop** — the modal, hygroscopic and volcanic-radius aerosol-optics tables.
- **soil_erod_mod** — the source-grid erodibility field, now read on **masterproc** only.
- **tracer_data** — one per-file buffer holding the global lat-lon source field in flight, so the node leader reads each field once instead of every rank reading it redundantly. Files already on the unstructured physics grid and zonal-average files are unaffected.

A few additional updates:

- New finalization path (**jlong_final**, **photo_final**, **physprop_final**) so the windows are freed collectively from **chem_final/phys_final**.
- **cam_shmem_init()** is called early in **cam_init** so the first **MPI_Comm_split_type(SHARED)** is paid at a controlled collective point rather than mid-**physprop**.
- Initial-condition and topo files are closed at the end of **cam_init**, after their last use in **phys_init**, instead of being held open until **cam_final**.
- **tracer_data** carries the overlapping time record forward in memory when the bracketing window advances by one record, instead of re-reading it.

This PR is done by Claude code under my supervisory.

Fixes #1640

### DIFF
--- a/src/utils/cam_shmem_mod.F90
+++ b/src/utils/cam_shmem_mod.F90
@@ -16,9 +16,10 @@ module cam_shmem_mod
 ! further synchronization.  Free it with cam_shmem_free(ptr, win) at finalize
 ! (collective over the node communicator); otherwise MPI_Finalize reclaims it.
 !
-! This module uses the F90 'mpi' module (not mpif.h) because the TYPE(C_PTR)
-! overloads of MPI_WIN_ALLOCATE_SHARED / MPI_WIN_SHARED_QUERY are only guaranteed
-! to be available through the 'mpi'/'mpi_f08' modules (MPI-3.0).
+! This module uses the F90 'mpi' module (not mpif.h).  MPI-3.0 makes the TYPE(C_PTR)
+! overloads of MPI_WIN_ALLOCATE_SHARED / MPI_WIN_SHARED_QUERY optional in the 'mpi'
+! module, and some implementations (cray-mpich) do not expose those two routines
+! there at all, so explicit interfaces for them are declared below instead.
 !-------------------------------------------------------------------------------
 
    use shr_kind_mod,   only: r4 => shr_kind_r4, r8 => shr_kind_r8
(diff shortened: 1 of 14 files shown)


src/utils/cam_shmem_mod.F90
sjsprecious, line 1: @cacraigucar @jtruesdal this file is generic and not specific to CAM. My suggestion is to merge this PR with it for now (after addressing all the review comments). After that I will merge this code with a similar one in CTSM (https://github.com/ESCOMP/CTSM/pull/4102), and bring it into `CESM_share` instead. Once we have a new `CESM_share` tag and bring it into CAM, we can then safely remove this code later.

PR #1661, opened 2026-08-25
cam6_4_198: Fix constants that were missing . in _r8 declarations 

### LINKED ISSUE #1659: Some _r8 constants are missing '.' and are being used as integers - need to fix these
### What happened?

It was recently pointed out that 1_r8 evaluates to an integer and not a double precision.  The TR8.sh script only checks for _r8, and not also having a `.`  or `e` to make the number be floating point.  

Need to enhance the script to also check that the constants are floating point and not inadvertent integers.

Will use Claude to enhance the script and will manually change the places it flags.

closes #1659

Visual inspection of code

cam6_4_196

Other (please explain below), CGD machine (e.g. izumi)

NAG

_No response_

Yes

### Extra info

_No response_

cacraigucar:
Any constant that is declared without a `.` or `e` is defined as an integer even when it it has `_r8` as its kind type.  This can lead to integer arithmetic if other variables are integers as well.

Closes #1659


test/system/TR8_enhanced.sh
```diff
@@ -0,0 +1,275 @@
...
+# The optional second argument is a comma-separated list of
+# sub-directories/files to skip, using the same skip semantics as
+# ADDREALKIND_EXE's "-s" option.
+check_r8_has_dot() {
+  dir="$1"
+  skiplist="$2"
+
+  [ -d "$dir" ] || return 0
+
+  filelist=$(find "$dir" -type f \( -name '*.F90' -o -name '*.f90' -o -name '*.F' -o -name '*.f' \) 2>/dev/null)
```
jimmielin: I would suggest
```suggestion
  filelist=$(cd "$dir" && find . -type f \( -name '*.F90' -o -name '*.f90' -o -name '*.F' -o -name '*.f' \) 2>/dev/null)
```

Otherwise the skip list is matched against the full path's name, so a CAM checkout at `/home/user/ext/` would be skipped completely, too.

test/system/TR8_enhanced.sh
```diff
@@ -0,0 +1,275 @@
...
+  # neither is present (e.g. "3_r8", "100_r8", "-5_r8") it is really an
+  # INTEGER literal of kind r8. A single-character grep lookbehind can't
+  # tell an exponent sign ("E+0_r8", valid) apart from an operator sign
+  # ("a + 3_r8", not valid), so this is parsed with perl instead.
+  # The lookbehind also must not treat every preceding "." as "mid
+  # number" -- old-style relational operators butt a "." directly
+  # against the literal (e.g. "res.gt.1.e-14_r8"), so only a dot that is
+  # itself preceded by a digit (i.e. an actual decimal point of the same
+  # number, as in "...3.5_r8") disqualifies a fresh match start.
+  bad=$(printf '%s\n' "$filelist" | xargs -d '\n' perl -ne '
```
jimmielin: 1/2 of changes to get rid of the redirect in case `perl` is potentially broken itself:

```suggestion
  bad=$(cd "$dir" && printf '%s\n' "$filelist" | tr '\n' '\0' | xargs -0 perl -ne '
```

test/system/TR8_enhanced.sh
```diff
@@ -0,0 +1,275 @@
...
+    my $bad = 0;
+    while (/(?<![0-9A-Za-z_])(?<!\d\.)(\d+\.?\d*|\.\d+)([eEdD][+-]?\d+)?_[rR]8\b/g) {
+      if (!defined($2) && $1 !~ /\./) { $bad = 1; last }
+    }
+    print "$ARGV:$.:$_" if $bad;
+    close ARGV if eof;
+  ' 2>/dev/null)
```
jimmielin: changes 2/2
```suggestion
  ')
  scan_rc=$?

  # check exit status for the scan actually completing
  # e.g., missing perl, an xargs without "-0", and any future typo in the perl above.
  if [ $scan_rc -ne 0 ]; then
    echo "TR8_enhanced:  ERROR: the _r8 scan of \"${dir}\" failed to run (rc = ${scan_rc})"
    return 1
  fi
```

test/system/TR8_enhanced.sh
```diff
@@ -0,0 +1,275 @@
...
+ruby $ADDREALKIND_EXE -r r8 -l 1 -d $CAM_ROOT/src/cpl
+rc=`expr $? + $rc`
+check_r8_has_dot $CAM_ROOT/src/cpl
+rc=`expr $? + $rc`
+
+fi
+
+echo $rc
+
+if [ $rc = 255 ]; then
```
jimmielin:
```suggestion
if [ $rc -ne 0 ]; then
```

I think `$rc` is a sum of "problematic" files (see `rc=expr $? + $rc` above) so if there are more than 255 errors `$rc` will wrap over to `0` and silently PASS

test/system/test_driver.sh
```diff
@@ -507,7 +507,7 @@ if [ "${cesm_test_suite}" != "none" -a -n "${cesm_test_mach}" ]; then
       sepstr="################################################################"
       echo "${sepstr}" | tee -a ${logfile}
       ark_file="/fs/cgd/csm/tools/addrealkind/addrealkind"
-      tr8_script="${CAM_ROOT}/test/system/TR8.sh"
+      tr8_script="${CAM_ROOT}/test/system/TR8_enhanced.sh"
```
jimmielin, line 510: I wonder if we should remove `TR8.sh` now that it is unused or replace it entirely with the "enhanced" version.

jimmielin (commented):
Thanks @cacraigucar! The CAM .F90 changes look good to me, but I had some suggestions on the script's robustness.

peverwhee (approved):
thanks @cacraigucar !

Reminder to remove the TR8.sh script.

PR #1664, opened 2026-08-31
 cam6_4_200: Remove PS noise in SE-CSLAM, WACCM stability and other stuff

### LINKED ISSUE #574: Setting time-steps based on model top location #574
Question to the group on how to set time-steps based on model top location:

Issue: the current namelist defaults queues off which physics package is used (for example, waccm_phys) to set time-steps such as setting se_nsplit. This is problematic since the new ~80km top CAM out-of-the-box uses ~140km top (WACCM) time-stepping which is much more restrictive than what is needed in the ~80km top leading to very low throughput. I don't think the physics package is a robust indicator of model top location!

Option 1: sponge layer settings queue off model top location inline in the code in SE. We could set time-steps there based on model top location (see dynamics/se/dycore/global_norms_mod.F90)
Option 2: if buildnamelist would know model top location then that could be used to set time-steps during the "build namelist" step

thanks

### LINKED ISSUE #1487: 2 degree (ne16pg3) out-of-the-box LT and MT functionality

2 degree (ne16pg3) out-of-the-box LT and MT functionality:

- create spun-up initial conditions and add to namelist defaults
- optimized dycore settings and add to namelist defaults

This issue does not concern forcing files etc. (which currently are CMIP6-based as far as I can tell)

_No response_

I Don't Know

Yes

### LINKED ISSUE #1523: optimize SE dycore for WACCM ne30pg3 and ne16pg3

The current namelist defaults and other settings are not optimal (in terms of computational performance) for 1 degree and 2 degree WACCM when using the spectral-element (SE) dycore. This issue is to discuss and improve on that ...

@JulioTBacmeister @mijeong135 @mbramberger

Yes

Yes

### LINKED ISSUE #1595: Remove subtle noise in PS in SE-CSLAM
### Issue Type

Other (please describe below)

### Issue Description

In the SE-CSLAM configuration, CSLAM overwrites water vapor (and other water species) on the GLL grid at every vertical remapping time step. In highly balanced flow with moisture (e.g., the FKESSLER idealized moist baroclinic wave), noise grows and can be seen at day 10 by choosing contour levels around 999–1001 hPa (see plot (b)):

Image

By applying a biharmonic viscosity operation after the mapping, the noise in the Tropics goes away, but the price to pay is a spurious wave in the Southern Hemisphere, similar to what is seen for MPAS. It is widely accepted that these waves are not problematic for real-world applications, as they grow much more slowly than the physical waves in the system. To compare different dycores with respect to spurious wave growth, see https://agupubs.onlinelibrary.wiley.com/doi/full/10.3894/JAMES.2010.2.15

### Will this change answers?

Yes

Yes

### LINKED ISSUE #1648: Wrong dycore settings when running WACCM physics with LT
### What happened?

With 

``` CAM_CONFIG_OPTS: -waccm_phys -model_top lt ...```

the SE dycore settings are set incorrectly in /bld/namelist_files/namelist_defaults_cam.xml::

Image

In particular, one gets a 30 level deep sponge which is very detrimental in the low top run. 

...

latest cam tag

CISL machine (e.g. derecho)

Intel

_No response_

Yes

### Extra info

@tilmes

### PR DESCRIPTION
**Pull request by PeterHjortLauritzen**
closes #1595  Remove subtle noise in PS in SE-CSLAM
  closes #1648  Wrong dycore settings when running WACCM physics with LT
  closes #1523  optimize SE dycore for WACCM ne30pg3 and ne16pg3
  closes #1487  2 degree (ne16pg3) out-of-the-box LT and MT functionality
  closes #574   Setting time-steps based on model top location
 
Also other changes like default topo file for ne120pg3 etc. See ChangeLog for details

### CODE REVIEW

File: bld/namelist_files/use_cases/waccm_ma_hist_cam7.xml [hunk 1]
```diff
@@ -16,7 +16,7 @@
 
 <!--Species IC -->
 <ncdata dyn="se" hgrid="ne30np4" nlev="135">atm/waccm/ic/FHIST_WAt1ma_ne30pg3_spinup01.cam.i.1980-01-01-00000_c250124.nc</ncdata>
-<ncdata dyn="se" hgrid="ne16np4" nlev="135">atm/waccm/ic/FHIST_WAt1ma_ne16pg3_spinup01.cam.i.1980-01-01-00000_c250124.nc</ncdata>
+<ncdata dyn="se" hgrid="ne16np4" nlev="135">/glade/work/fvitt/waccm_ic/FHISTC_WAt1ma_ne16pg3_watervapor_initial_spinup01.cam.i.1980-01-01-00000.nc</ncdata>
```
nusbaume, line 19: @fvitt is this IC file one that should go into the inputdata (GDEX) repo?

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3238,33 +3243,31 @@
...
 <se_hypervis_subcycle  hgrid="ne0np4.ARCTIC.ne30x4"                             > 2  </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4.POLARCAP.ne30x4"                           > 2  </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4CONUS.ne30x8"                               > 1  </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4.ARCTICGRIS.ne30x8"                         > 1  </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4.NATL.ne30x8"                               > 1  </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4CONUS.ne30x8" waccm_phys="1"                > 1 </se_hypervis_subcycle>
-<se_hypervis_subcycle  hgrid="ne16np4" model_top="ht"                           > 18</se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne16np4" model_top="xt"                           > 18</se_hypervis_subcycle>
-<se_hypervis_subcycle  hgrid="ne30np4" model_top="ht"                           > 18</se_hypervis_subcycle>
+<se_hypervis_subcycle  hgrid="ne30np4" model_top="ht"                           >-1</se_hypervis_subcycle>
```
nusbaume: I don't think you need this entry anymore, as it is the same as the new default value:


```suggestion
```

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3238,33 +3243,31 @@
...
 <se_hypervis_subcycle  waccmx="1" hgrid="ne120np4" nlev="273"                   > 30</se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne3np4"                                           > 3 </se_hypervis_subcycle>
 <se_hypervis_subcycle  hgrid="ne0np4TESTONLY.ne5x4"                             > 3 </se_hypervis_subcycle>
 
 <se_hypervis_subcycle_sponge                                                 >-1  </se_hypervis_subcycle_sponge>
 <se_hypervis_subcycle_sponge hgrid="ne30np4" model_top="mt" waccm_phys="1"   > 3  </se_hypervis_subcycle_sponge>
-<se_hypervis_subcycle_sponge hgrid="ne16np4"  waccm_phys="1" waccmx="0"      > 2  </se_hypervis_subcycle_sponge>
-<se_hypervis_subcycle_sponge hgrid="ne30np4"  waccm_phys="1" waccmx="0"      > 4  </se_hypervis_subcycle_sponge>
+<se_hypervis_subcycle_sponge hgrid="ne16np4"  waccm_phys="1" waccmx="0"      >-1</se_hypervis_subcycle_sponge>
+<se_hypervis_subcycle_sponge hgrid="ne30np4"  waccm_phys="1" waccmx="0"      >-1</se_hypervis_subcycle_sponge>
```
nusbaume: Do you still need these entries?  It looks like they are the same as the new default value (`-1`), and so may not be doing anything.

This same question holds for all of the other settings below which are using `-1` when it is also the default value for that namelist variable.

src/dynamics/se/dycore/dimensions_mod.F90
```diff
@@ -38,6 +38,12 @@ module dimensions_mod
   integer         :: ntrac = 0           !ntrac is set in dyn_comp
   logical, public :: use_cslam = .false. !logical for CSLAM
   integer         :: qsize = 0           !qsize is set in dyn_comp
+  integer, public :: wv_idx_dycore = -1  !water vapor index in dycore Qdp; set in dyn_comp
+  logical, public :: del4_cslam_qgll = .true. !apply del4 hyperviscosity to water vapor on GLL after cslam2gll
+  logical, public :: gll_advect_q = .false.   !GLL double advection of thermodynamic tracers under CSLAM
+                                              !(cslam2gll overwrite not used)
+  logical, public :: cslam_q_filter = .false. !mass-conservative del4 damping of water vapor on the CSLAM grid
+  real(r8), public :: cslam_q_filter_nu_fac = 0.5_r8 !factor on nu_p for the CSLAM-grid del4
```
nusbaume: I'm happy to be over-ruled here, but is `dimensions_mod` the best location for these variables?  I would have naively assumed that somewhere like `control_mod` would be a better fit, as that is where many of the other "del4" variables are, and these quantities don't appear to be directly related to the number of elements or vertical levels.

src/dynamics/se/dycore/global_norms_mod.F90
```diff
@@ -18,13 +18,17 @@ module global_norms_mod
...
   public :: set_global_max_normDinv
   public :: global_integral
   public :: global_integrals_general
   public :: wrap_repro_sum
 
+  public :: nu_q_cslam
+
   private :: global_maximum
   type (EdgeBuffer_t), private :: edgebuf
+  real(r8), save :: nu_q_cslam = -1.0_r8
```
nusbaume: You don't need `save` here, as it is automatically saved by being at the module-level:


```suggestion
  real(r8) :: nu_q_cslam = -1.0_r8
```

src/dynamics/se/dycore/global_norms_mod.F90
```diff
@@ -742,6 +770,39 @@ subroutine print_cfl(elem,hybrid,nets,nete,ptop,pmid)
     dt_dyn_del2_actual     = tstep/hypervis_subcycle_sponge
     dt_tracer_visco_actual = tstep*qsplit/hypervis_subcycle_q
     dt_phys                = dtime
+    !
+    ! auto-set cslam_q_filter_nsub for the CSLAM-grid del4 Q filter
+    !
+    if (use_cslam .and. cslam_q_filter) then
+      ! Smallest CSLAM cell area = 0.833 * mean cell area on the equiangular
+      ! gnomonic cubed sphere.
+      min_area_fvm_m2 = 0.83_r8*4.0_r8*pi/dble(6*ne*ne*nc*nc)*rearth*rearth
```
nusbaume: Should this be `0.833_r8` here instead of `0.83_r8`?  I am just basing this off the comment right above it.  Also I would use `real` instead of `dble` here so that the precision can be specified exactly:


```suggestion
      min_area_fvm_m2 = 0.83_r8*4.0_r8*pi/real(6*ne*ne*nc*nc, r8)*rearth*rearth
```

src/dynamics/se/dycore/viscosity_mod.F90
```diff
@@ -156,8 +157,17 @@ subroutine biharmonic_wk_dp3d(elem,dptens,dpflux,ttens,vtens,deriv,edge3,hybrid,
...
+    !$omp parallel do num_threads(vert_num_threads) private(k,v,tmp,tmp2,nu_ratio2)
     do k=kbeg,kend
+      ! recompute nu_ratio2 per level (see first loop); it is level-dependent
+      nu_ratio2=1
+      if (nu_div_lev(k)/=nu_lev(k)) then
+        if(hypervis_scaling /= 0) then
+          nu_ratio2=1
+        else
+          nu_ratio2=sqrt(nu_div_lev(k)/nu_lev(k))
+        endif
```
nusbaume: I don't think you need one of the conditional paths here, as it looks like a no-op:


```suggestion
        if(hypervis_scaling == 0) then
          nu_ratio2=sqrt(nu_div_lev(k)/nu_lev(k))
        endif
```

nusbaume (changes requested):
Thanks For the dycore updates and fixes @PeterHjortLauritzen!  I have some change requests, but hopefully nothing too difficult (and if there is a problem then just let me know).  I should also note that I did not review `fvm_filter_mod.F90`, as I believe @jtruesdal was going to look over that file as part of his review.

src/dynamics/se/dycore/fvm_filter_mod.F90
```diff
@@ -0,0 +1,641 @@
...
+       kptr = kmin - 1 + nlev
+       call ghostpack(ghostBufQfilter, fvm(ie)%c(0:nc+1, 0:nc+1, kmin:kmax, q), kblk, kptr, ie)
+    end do
+    call ghost_exchange(hybrid, ghostBufQfilter, location='cslam_q_filter_del4_passQ')
+    do ie = nets, nete
+       kptr = kmin - 1
+       call ghostunpack(ghostBufQfilter, fvm(ie)%dp_fvm(0:nc+1, 0:nc+1, kmin:kmax), kblk, kptr, ie)
+       kptr = kmin - 1 + nlev
+       call ghostunpack(ghostBufQfilter, fvm(ie)%c(0:nc+1, 0:nc+1, kmin:kmax, q), kblk, kptr, ie)
+    end do
```
jtruesdal, line 141: The layer mass array dp_fvm isn't modified within the isub loop and can be exchanged outside.  This halves the mpi communications for the first exchange and should be more efficient.  I made a quick test of the code and it was bfb.   It required creating a new 1nlev length array to use for the dp_fvm array.

src/dynamics/se/dycore/fvm_filter_mod.F90
```diff
@@ -0,0 +1,641 @@
...
+    !----- 3. Halo exchange of L -----
+    do ie = nets, nete
+       kptr = kmin - 1
+       call ghostpack(ghostBufQfilter, L(0:nc+1, 0:nc+1, kmin:kmax, ie), kblk, kptr, ie)
+    end do
+    call ghost_exchange(hybrid, ghostBufQfilter, location='cslam_q_filter_del4_passL')
+    do ie = nets, nete
+       kptr = kmin - 1
+       call ghostunpack(ghostBufQfilter, L(0:nc+1, 0:nc+1, kmin:kmax, ie), kblk, kptr, ie)
+    end do
```
jtruesdal, line 198: This halo exchange is also using a 2*nlev buffer for a 1*nlev field so doubling the MPI communication needed.  Just reusing the new 1*nlev smaller buffer that was used for the mass array above should make this more efficient.

src/dynamics/se/dycore/fvm_filter_mod.F90
```diff
@@ -0,0 +1,641 @@
...
+             do j = 0, nc
+                do i = 1, nc
+                   Gy(i, j) = xw_ns(i, j) * (fvm(ie)%c(i, j+1, k, q) - fvm(ie)%c(i, j, k, q)) + &
+                              xt_ns(i, j) * ( (fvm(ie)%c(i+1, j  , k, q) + fvm(ie)%c(i+1, j+1, k, q))   &
+                                            - (fvm(ie)%c(i-1, j  , k, q) + fvm(ie)%c(i-1, j+1, k, q)) )
+                end do
+             end do
+             do j = 1, nc
+                do i = 1, nc
+                   inv_area = 1.0_r8 / fvm(ie)%area_sphere(i, j)
```
jtruesdal: Small improvement in efficiency maybe.  Your choice.  inv_area has no level dependence and could be precomputed to save kmin:kmax*nc*nc divisions per subcycle.  It seems like it could also be part of the fvm structure and done just once and used everywhere.  inv_dp_area below could also be cached after the first calculation before the exchange.

jtruesdal (changes requested):
Nice job Peter. I did my best to check the stenciling and algorithms they all look good.  I did notice that the mpi exchanges could be made more efficient by pulling a static nlev arrays out of the isub loop and exchanging proper sized buffers.  I ran the mpi improvements and it was BFB on Izumi. If you agree with the suggestions I could update your branch if I have permission or push a PR with the changes if you'd like.

src/dynamics/se/dycore/dimensions_mod.F90
PeterHjortLauritzen: done

PeterHjortLauritzen:
Thanks Jesse and John! I think the issues (thus far) have been addressed (but since Derecho is down I have not been able to check regression test impact of changes)

src/dynamics/se/dycore/fvm_filter_mod.F90
PeterHjortLauritzen: good point! Done

src/dynamics/se/dycore/fvm_filter_mod.F90
PeterHjortLauritzen: Oh dear ... good catch!

src/dynamics/se/dycore/fvm_filter_mod.F90
PeterHjortLauritzen: Oh dear ... good catch!

bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -3322,14 +3316,8 @@
...
 <se_sponge_del4_nu_fac hgrid="ne120np4" nlev="273" waccmx="1"> 3.0 </se_sponge_del4_nu_fac>
 <se_sponge_del4_nu_div_fac> -1 </se_sponge_del4_nu_div_fac>
-<se_sponge_del4_nu_div_fac hgrid="ne16np4" waccm_phys="1" waccmx="0">-1</se_sponge_del4_nu_div_fac>
-<se_sponge_del4_nu_div_fac model_top="ht"                           >-1</se_sponge_del4_nu_div_fac>
 <se_sponge_del4_nu_div_fac hgrid="ne120np4" nlev="273" waccmx="1"   > 5.0 </se_sponge_del4_nu_div_fac>
 <se_sponge_del4_lev       > -1 </se_sponge_del4_lev>
-<se_sponge_del4_lev waccm_phys="1" model_top="mt"> -1 </se_sponge_del4_lev>
-<se_sponge_del4_lev waccm_phys="1">-1</se_sponge_del4_lev>
-<se_sponge_del4_lev model_top="ht">-1</se_sponge_del4_lev>
-<se_sponge_del4_lev model_top="xt"> 30 </se_sponge_del4_lev>
```
nusbaume, line 3331: Do we want to keep the value at 30 when using the WACCM-X top?

nusbaume (approved):
Thanks for addressing my comments @PeterHjortLauritzen!  I just had one final comment (mostly to double-check if the latest changes are correct).

jtruesdal:
@PeterHjortLauritzen I can run the regression tests later tonight if derecho comes up.  @nusbaume I'm helping Peter get the mpi exchange mods correct and I'll test the new version just to make sure its BFB.

bld/namelist_files/namelist_defaults_cam.xml
PeterHjortLauritzen: Good catch (I don't think it is needed anymore but it will need to be tested). Restored for now ...

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

src/dynamics/se/dycore/fvm_filter_mod.F90
jtruesdal: Done.

jtruesdal (approved):
My issues have been addressed and the code is BFB with what came before the fvm_filter_mod.F90 improvements.

PR #1668, opened 2026-09-01
cam6_4_201: Age of air test tracer fixes

### LINKED ISSUE #1654: AOA1 (Age Of Air Tracers) IC
### What happened?

In aoa_tracers.F90, AOA1 tracer value read from the IC file is not compatible with the initial value. 
It should be read from the IC file (lowest layer) then averaged globally then fed back into the code.

Use an IC file that has pre-calculated AOA1 in it.
In user_nl_cam, set "aoa_tracers_flag = .true."
In aoa_tracers.F90, read_from_ic_file is set to be true by default.

cesm3_0_alpha07e

CISL machine (e.g. derecho)

Intel

/glade/work/tilmes/derecho/carma/fixed_sst_paper/FWmaCARMA2000climo_2040ghg.f19_f19_mg17.cesm3_0_alpha07e.solid.control.002

No

### Extra info

 aoa_tracers.F90 should be checked and this should be documented.

fvitt:
Closes #1654 

We are still testing these mods.  This might not be the final version.


fvitt:
This is ready for review now

src/physics/cam/aoa_tracers.F90
```diff
@@ -255,12 +283,15 @@ subroutine aoa_tracers_timestep_init( phys_state )
 
     use ppgrid,         only: begchunk, endchunk
```
peverwhee: this use statement can be removed (it's now up at the module level)

src/physics/cam/aoa_tracers.F90
```diff
@@ -349,8 +397,13 @@ subroutine aoa_tracers_timestep_tend(state, ptend, dt)
 
     ! AOA1
     xmmr = mmr0*(1._r8 + per_yr*years)
+
+    ! Lower boundary
     ptend%q(1:ncol,pver,ixaoa) = (xmmr - state%q(1:ncol,pver,ixaoa)) / dt
 
+    ! Upper boundary AOA1 -> 0.0 mmr
```
peverwhee: Maybe a clearer comment here?

```suggestion
    ! Set upper boundary AOA1 tendency so that upper boundary AOA1 state will end up 0.0 mmr
```

peverwhee (approved):
two small nits. Once fixed, I'm approving this for cam6_4_201

src/physics/cam/aoa_tracers.F90
fvitt: done

Issue #1650, opened 2026-08-17
Broken threading (vertical/tracer) in CAM


briandobbins:
### What happened?

Threading has been broken in CAM for a while.  I've got some fixes for the vertical and tracer threading that I'll submit PRs for this week.

Horizontal threading may or may not come in with these PRs; the primary focus was on threaded cases where CAM is already scaled out to the maximum ranks (1 element/rank), hence the vertical/tracer title.

### What are the steps to reproduce the bug?

Run an FHISTC_MTt4s case multiple times with >1 threads in the atmosphere and you'll get crashes or different results.

### What CAM tag were you using?

cam6_4_195

### What machine were you running CAM on?

CISL machine (e.g. derecho)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Yes

PeterHjortLauritzen:
Lets make sure we introduce a threading regression test with the SE dycore once this issue has been resolved @cacraigucar @nusbaume

briandobbins:
OK, three PRs open to get things working:
#1651 (gravity wave fix)
#1672 (radiation fix)
#1673 (SE dycore fixes)

@PeterHjortLauritzen @nusbaume In addition to adding a test (probably more than one), it's probably worth discussing programmatically calculating the correct 'pcols' value for CAM_CONFIG_OPTS, since it has huge impacts especially at the 'scaled-out' bit.  Eg, with 1 element/core, and a pg3 grid, you have 9 columns -- with 2 threads, you want pcols=5, and with 3 threads you want pcols=3.  

The impact on performance is big, and also relates to the previous observation you both had about pcols=16 when running at 1 element/core.  I incorrectly felt that wouldn't be large because I thought we used ncols in most places, but for automatic array assignments, they still get the full 'pcols' dimension, wasting memory & cache.  That's even more important with, say, 3 threads, where the ratio of actual columns to defined columns goes from 9:16 -> 3:16.  

I'll have more data on the non-scaled out version of that soon, and I have a tentative PR for doing this automatically, but figured I'd get your take first.  I know CAM-SIMA does it already, and that's my guess for a part of why we saw better performance there (better caching).

briandobbins:
I'll note that I've _not_ run the full CAM regression test, but did do bit-for-bit comparisons with unmodified and non-threaded versions of an MTt4s case.

If we have any old compilers, I may need to introduce a helper function since omp_set_nested is deprecated now.

briandobbins:
[...]

PR #1673, opened 2026-09-07
part of cam6_4_202: SE dycore vertical/tracer threading fixes

briandobbins:
This PR has multiple small changes to enable threading in the SE dycore again -- it fixes some race conditions, updates the OpenMP standard to 5.0 (circa 2018), and has been tested as bit-for-bit with 1-4 threads on the same MTt4s case.

Partial fix for #1650

### DIFF
--- a/src/dynamics/se/dycore/thread_mod.F90
+++ b/src/dynamics/se/dycore/thread_mod.F90
@@ -7,7 +7,8 @@ module thread_mod
        omp_get_max_threads, &
        omp_get_num_threads, &
        omp_get_nested,      &
-       omp_set_nested
+       omp_set_nested,      &
+       omp_set_max_active_levels
 #endif
   use cam_logfile,            only: iulog
   use spmd_utils,             only: masterproc
@@ -25,6 +26,7 @@ module thread_mod
   public :: omp_get_num_threads
   public :: omp_get_nested
   public :: omp_set_nested
+  public :: omp_set_max_active_levels
   public :: initomp
 contains
 
@@ -60,6 +62,10 @@ contains
   subroutine omp_set_nested(flag)
     logical :: flag
   end subroutine omp_set_nested
+
+  subroutine omp_set_max_active_levels(level)
+    integer :: level
+  end subroutine omp_set_max_active_levels
 
   subroutine initomp
     max_num_threads = 1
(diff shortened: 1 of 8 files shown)


src/dynamics/se/dycore/thread_mod.F90
```diff
@@ -7,7 +7,8 @@ module thread_mod
        omp_get_max_threads, &
        omp_get_num_threads, &
        omp_get_nested,      &
-       omp_set_nested
+       omp_set_nested,      &
```
jimmielin, line 10: I believe these are fully unused now except for mentions in comments in `prim_driver_mod.F90`, so they could be removed as deprecated.

jimmielin (approved):
Thanks @briandobbins! The PR looks good to me. I have an optional quick cleanup comment but it is OK if we leave it to a later time due to the priority of this.

src/dynamics/se/dycore/thread_mod.F90
cacraigucar: > I believe these are fully unused now except for mentions in comments in `prim_driver_mod.F90`, so they could be removed as deprecated.

I have started up regression tests, so I would need to restart them if we make this change.  Let's open an issue and assign a "misc tag" to it, and I'll remove it in a future PR

briandobbins:
Let's leave them in for now -- that was 'semi-intentional', since I wasn't
sure if we test with any older compilers.  Ones prior to 2018 without
OpenMP 5.0 support would need an #ifdef to swap between the (old)
omp_*_nested and (new) omp_*_max_active_levels definitions.  I figured a
test would highlight that if it is the case, and if not, it's an easy
removal later.

Thanks!
  - Brian

cacraigucar:
> Let's leave them in for now -- that was 'semi-intentional', since I wasn't sure if we test with any older compilers. Ones prior to 2018 without OpenMP 5.0 support would need an #ifdef to swap between the (old) omp_*_nested and (new) omp_*_max_active_levels definitions. I figured a test would highlight that if it is the case, and if not, it's an easy removal later. Thanks! - Brian

Haipeng actually removed that in my PR #1674 which combines all of your PRs and I restarted the regression tests (which needed an updated atmos_phys tag as well).  Since the tests have been submitted, I will keep the line removed and we can put it back in if we need it in the future.

PR #1676, opened 2026-09-15
Add clubb_a_const into the namelist.

HingOng:
This PR adds clubb_a_const into the namelist.

Before this change, a_const = 1.8 was set in src/physics/clubb/src/CLUBB_core/parameters_tunable.F90, and it could be reset with SourceMods.

After this change, a_const = 1.8 is set in bld/namelist_files/namelist_defaults_cam.xml, and it can be reset with user_nl_cam.

My 5-day run results appear BFB.

Setting a_const may be useful for tuning the lowest-level wind speed, as shown in Section 4.1 of https://doi.org/10.22541/essoar.177135891.11467584/v3


adamrher:
@HingOng just so I'm clear. If there is a CLUBB_core namelist in parameters_tuneable.F90 that is not a CAM namelist, we can make it a CAM namelist without making any modifications to CLUBB_core?

HingOng:
> @HingOng just so I'm clear. If there is a CLUBB_core namelist in parameters_tuneable.F90 that is not a CAM namelist, we can make it a CAM namelist without making any modifications to CLUBB_core?

Ideally, any parameter set in parameters_tunable.F90 (CLUBB) can be overwritten by passing a parameter value from CAM to CLUBB via using clubb_api_module.F90 (CLUBB) in clubb_intr.F90 (CAM). However, it does need CLUBB developers' attention to match the parameters between parameters_tunable.F90 and clubb_api_module.F90.

In the present case, a CLUBB developer has matched a_const between those in the following commit: https://github.com/larson-group/clubb_release/commit/903169aafe5cfd2f7cae35adef3d31e6f51641fe. Thus, no more modification to CLUBB is needed.

